PR #29, opened 2025-12-29
Add auto-update check and CLI update command

willwashburn:
Introduces an update-checker utility that checks for new versions of agent-relay via the npm registry, with caching to limit API calls. The CLI now checks for updates in the background for interactive commands and adds a new 'update' command to check for and install updates. This improves user awareness of new releases and streamlines the update process.


khaliqgant (approved):
awesome, thanks!

PR #530, opened 2026-03-10
fix: pass failure context to workflow step retry attempts

khaliqgant:
## Summary

Closes #497

- When a workflow step fails and retries, the retried agent now receives failure context prepended to its task string
- Context includes: retry attempt number, the previous error message, and the last 2000 characters of PTY/stdout output from the failed attempt
- Output is captured from both interactive (PTY-based) and non-interactive (subprocess) execution paths

## Test plan

- [ ] Run a workflow with a step configured with `retries: 1+` that is expected to fail on the first attempt
- [ ] Verify the retried agent receives the `[RETRY]` header with the failure message and prior output
- [ ] Verify that successful first attempts are unaffected (no retry header prepended)
- [ ] Verify non-interactive agent retries also include stdout/stderr from the failed attempt

---

<a href="https://app.devin.ai/review/agentworkforce/relay/pull/530" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/packages/sdk/src/workflows/runner.ts
+++ b/packages/sdk/src/workflows/runner.ts
@@ -223,6 +223,8 @@ export class WorkflowRunner {
 
   // PTY-based output capture: accumulate terminal output per-agent
   private readonly ptyOutputBuffers = new Map<string, string[]>();
+  /** Snapshot of PTY output from the most recent failed attempt, keyed by step name. */
+  private readonly lastFailedStepOutput = new Map<string, string>();
   private readonly ptyListeners = new Map<string, (chunk: string) => void>();
   private readonly ptyLogStreams = new Map<string, WriteStream>();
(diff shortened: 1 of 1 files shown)


khaliqgant (commented):
### Findings

1. **[medium] `lastFailedStepOutput` can retain stale output when a failed attempt produces no output**
`lastFailedStepOutput` is only updated in `executeAgentStep` and `spawnAndWait` when output is non-empty (`if (combinedOutput) ...` / `if (ptyChunks.length > 0) ...`).

This means if attempt N fails with output and attempt N+1 also fails with no output, the retry context for attempt N+2 still includes the older output from attempt N, not the most recent failed attempt. That violates the “error + last output from previous attempt” intent and the no-previous-output edge case.

Suggested fix: clear/update the map for each attempt (e.g., set `""` when output is empty) before the next retry context is built.

PR #730, opened 2026-04-13
fix(sdk): align WorkflowTrajectory writer with canonical layout + fields

khaliqgant:
## Summary

Aligns `WorkflowTrajectory` (the standalone SDK writer used by the workflow runner) with the canonical `agent-trajectories` layout and top-level field contract. Defense-in-depth follow-up to [AgentWorkforce/trajectories#22](https://github.com/AgentWorkforce/trajectories/pull/22), which makes the reader tolerant of the current SDK output.

## Why

`WorkflowTrajectory` writes trajectory JSON directly — it never touches the `agent-trajectories` library. Over time its output diverged from what `FileStorage.save()` produces:

1. **Completed files landed in a flat `completed/` root** instead of `completed/YYYY-MM/{id}.json`. This is why workforce's `.trajectories/completed/` is a flat pile of files — every one of them came through this writer.
2. **Top-level `commits`, `filesChanged`, and `tags` arrays were missing**, even though the canonical schema declares them. The reader had to default them on load.

Neither is a correctness bug today (trajectories#22 made the reader tolerant), but the divergence forces the reader to carry legacy-layout fallbacks forever. Aligning the writer lets those fallbacks be removed in a future release.

## Changes

- **`moveToCompleted()`** — computes a `YYYY-MM` bucket from `completedAt` (falling back to `startedAt`) and writes the file to `completed/<bucket>/{id}.json`. Bucket directory is created with `mkdir -p`.
- **`TrajectoryFile` interface** — declares `commits: string[]`, `filesChanged: string[]`, `tags: string[]`.
- **`start()`** — initializes all three to `[]` so the written trajectory is canonical-complete from the first `flush()`.

## Test plan

- [x] `npm run build` (SDK package)
- [x] `npx vitest run src/__tests__/workflow-trajectory.test.ts` — 29 passed (27 pre-existing + 2 new)
- [x] New test: completed file path has exactly one `YYYY-MM` intermediate directory
- [x] New test: `commits` / `filesChanged` / `tags` populated as `[]` after `start()`
- [x] Existing `readCompletedTrajectoryFile` helper now walks `completed/` recursively so tests don't hardcode the bucket name

Unrelated full-suite failures in `verification.test.ts`, `step-executor.test.ts`, etc. are pre-existing on `origin/main` and out of scope for this change.

## Reader companion

AgentWorkforce/trajectories#22 — relaxes the zod schema on the reader side so ids like `traj_<ts>_<hex>`, roles like `workflow-runner`/`specialist`, and missing `commits`/`filesChanged`/`projectId`/`tags` no longer fail validation. This SDK PR is the upstream "fix it at source" companion — both can ship independently; the reader PR unblocks the immediate bug (`trail compact --all` → "No trajectories found"), and this SDK PR makes new trajectories canonical so the reader can stop tolerating legacy shapes in a future release.

---

<a href="https://app.devin.ai/review/agentworkforce/relay/pull/730" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/packages/sdk/src/__tests__/workflow-trajectory.test.ts
+++ b/packages/sdk/src/__tests__/workflow-trajectory.test.ts
@@ -32,15 +32,32 @@ function readTrajectoryFile(dir: string): any {
   return JSON.parse(readFileSync(path.join(activeDir, jsonFiles[0]), 'utf-8'));
 }
 
-function readCompletedTrajectoryFile(dir: string): any {
+function findCompletedTrajectoryJson(dir: string): string | null {
   const completedDir = path.join(dir, '.trajectories', 'completed');
   if (!existsSync(completedDir)) return null;
 
-  const files = readdirSync(completedDir);
-  const jsonFiles = files.filter((f: string) => f.endsWith('.json'));
-  if (jsonFiles.length === 0) return null;
+  // Completed trajectories now live under completed/YYYY-MM/. Walk the
+  // tree so tests don't have to know the exact bucket name.
+  const stack: string[] = [completedDir];
+  while (stack.length > 0) {
+    const current = stack.pop() as string;
+    const entries = readdirSync(current, { withFileTypes: true });
+    for (const entry of entries) {
+      const entryPath = path.join(current, entry.name);
+      if (entry.isDirectory()) {
+        stack.push(entryPath);
+      } else if (entry.isFile() && entry.name.endsWith('.json')) {
+        return entryPath;
+      }
+    }
+  }
+  return null;
+}
 
-  return JSON.parse(readFileSync(path.join(completedDir, jsonFiles[0]), 'utf-8'));
+function readCompletedTrajectoryFile(dir: string): any {
+  const jsonPath = findCompletedTrajectoryJson(dir);
+  if (!jsonPath) return null;
+  return JSON.parse(readFileSync(jsonPath, 'utf-8'));
 }
 
 // ── Tests ────────────────────────────────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


khaliqgant:
Closing in favor of a comprehensive unification approach.

This PR was a tactical alignment: make `WorkflowTrajectory` emit the canonical layout (`completed/YYYY-MM/`) and populate canonical top-level arrays. It fixes the symptom — writer output drifting from the reader's expected shape — without fixing the root cause: **the SDK re-implements the trajectory data model instead of using `agent-trajectories` directly**.

`@agent-relay/sdk` does not currently depend on `agent-trajectories`, despite the latter shipping a complete public SDK (`TrajectoryClient`, `TrajectorySession`, `TrajectoryBuilder`) at `./sdk`. `WorkflowTrajectory` is a 781-line parallel implementation with its own types, event model, chapter logic, ID format, and file layout. Every change to the canonical schema requires a parallel change here, and drift is inevitable — this PR and [AgentWorkforce/trajectories#22](https://github.com/AgentWorkforce/trajectories/pull/22) are both manifestations of that drift.

The right fix is to replace `WorkflowTrajectory`'s internals with a thin wrapper around `TrajectoryClient` / `TrajectorySession`, mapping workflow-specific events to canonical `addChapter` / `addEvent` / `addDecision` calls. Non-blocking behavior is preserved via `try/catch` around the delegated calls. Net effect: ~+250 / –700 lines, one source of truth for the trajectory model, drift becomes structurally impossible.

A follow-up PR implementing the unification will land shortly.

Companion reader-tolerance fix (ships independently): https://github.com/AgentWorkforce/trajectories/pull/22

PR #733, opened 2026-04-13
fix(cli): make preParseWorkflowFile async to avoid Bun-compiled CLI hang

khaliqgant:
## Problem

PR #727 added preParseWorkflowFile using esbuild.transformSync for actionable template-literal error hints. Unfortunately, esbuild.transformSync uses a blocking read loop to IPC with esbuild internal Go subprocess — and Bun spawnSync is subtly incompatible with that loop in compiled standalone binaries.

User-visible symptom: running agent-relay run workflow.ts silently hangs after printing

    [agent-relay] runScriptFile: preParseWorkflowFile start

with no further output and no way to recover except ctrl-c.

Empirical proof that the same file parses fine in regular Node: running node -e to call esbuild.transformSync on the same file completes in 254ms. The hang is specific to Bun-compiled standalone binaries.

## Fix

Convert preParseWorkflowFile to async and use esbuild.transform (async IPC via pipe streams) instead of transformSync. Wrap the call in Promise.race with a 5 second hard timeout as a safety net.

On timeout: log a warning and skip the pre-parse. tsx will still catch actual syntax errors downstream, so the worst case is losing the actionable hint — same behavior as pre-#727.

runScriptFile is converted to async to await preParseWorkflowFile. The SetupDependencies.runScriptWorkflow type is widened to accept void or Promise of void. Commander action handlers already support async natively; no other changes.

## Test plan

- [x] Unit tests in setup.test.ts updated to await the async API
- [x] Full typecheck clean
- [x] Full vitest run for packages/sdk/src/__tests__/
- [x] Integration test A: build CLI from source, spawn it against a synthetic TS workflow file, measure duration (< 15 seconds), assert expected diag checkpoints fire in order (resolving → preParseWorkflowFile start → preParseWorkflowFile done → trying runner). Proves the happy path.
- [x] Integration test B: build CLI from source, spawn it against the actual workflow file that was hanging in production (workflows/wire-up-slack-integration.ts from the cloud repo) with DRY_RUN=true, measure duration (< 20 seconds), assert preParseWorkflowFile start AND done diag lines fire. This is the real-world 80-to-100 gate that proves the user-visible hang is fixed on the exact input that triggered it.
- [ ] Post-merge: re-run the workflows that were silently hanging (wire-up-slack-integration.ts, and similar) to confirm.

## Why not revert PR #727?

Reverting would lose the actionable template-literal error hints that made debugging common workflow authoring mistakes much easier. The async fix preserves that feature while eliminating the hang.

---

<a href="https://app.devin.ai/review/agentworkforce/relay/pull/733" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/cli/commands/setup.test.ts
+++ b/src/cli/commands/setup.test.ts
@@ -202,7 +202,7 @@ describe('preParseWorkflowFile', () => {
     return full;
   }
 
-  it('returns silently for a valid TypeScript workflow file', () => {
+  it('returns silently for a valid TypeScript workflow file', async () => {
     const file = writeTempWorkflow(
       'valid.ts',
       `
@@ -215,10 +215,10 @@ workflow('w')
   });
 `.trim()
     );
-    expect(() => preParseWorkflowFile(file)).not.toThrow();
+    await expect(preParseWorkflowFile(file)).resolves.toBeUndefined();
   });
 
-  it('wraps a raw backtick inside a template literal with an actionable hint', () => {
+  it('wraps a raw backtick inside a template literal with an actionable hint', async () => {
     // A raw backtick inside a command: template literal terminates
     // the outer JS template literal early and produces an esbuild
     // parse error. We want the error message to tell the user how
@@ -227,9 +227,9 @@ workflow('w')
       'bad-backtick.ts',
       ['const step = {', '  command: `git commit -m "use `npm install` here"`,', '};'].join('\n')
     );
-    expect(() => preParseWorkflowFile(file)).toThrow(/Workflow file failed to parse/);
+    await expect(preParseWorkflowFile(file)).rejects.toThrow(/Workflow file failed to parse/);
     try {
-      preParseWorkflowFile(file);
+      await preParseWorkflowFile(file);
     } catch (err) {
       const msg = (err as Error).message;
       expect(msg).toMatch(/Hint:/);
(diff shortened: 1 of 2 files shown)


xkonjin (commented):
## Code Review: Async preParseWorkflowFile to prevent CLI hangs

Overall, this is a solid fix for a tricky Bun-specific edge case. The timeout mechanism is sensible and the graceful degradation when esbuild hangs is exactly the right approach.

### Security Considerations
- **None identified** — The 5000ms timeout prevents resource exhaustion without introducing new attack vectors.

### Potential Issues

1. **Error handling in timeout path** (line ~206-210): When PREPARSE_TIMEOUT occurs, you console.warn and return silently. This could mask genuine esbuild issues in production. Consider adding a metric/counter for timeouts so operators can spot patterns.

2. **Timer cleanup race condition**: The finally block clears the timer, but there's a subtle race: if the timeout fires after esbuild succeeds but before Promise.race resolves, the rejection could be unhandled. In practice this is unlikely given Node's event loop, but worth keeping in mind.

### Test Coverage

The new test case for timeout behavior is excellent — it properly mocks esbuild, uses fake timers, and cleans up state. One suggestion: add an integration test for the actual Bun-compiled CLI hang scenario if feasible.

### Suggestions

- Line ~207: The warning says "tsx will still catch real syntax errors" — is this fully accurate? If the workflow file has a genuine syntax error, tsx will catch it at runtime, but the pre-parse phase won't. Consider clarifying: "syntax errors will surface at runtime instead of pre-parse."

- Type safety: The runScriptWorkflow return type changed from void to void | Promise<void>. Ensure all callers handle both cases correctly. The await on line 660 is good, but verify there are no other call sites.

### Verdict

LGTM — the fix is correct and the timeout prevents the hang. The suggestions above are polish, not blockers.

PR #737, opened 2026-04-13
fix(messaging): inbox --agent flag, history DM support, history --from DM context

khaliqgant:
## Summary

Three messaging CLI fixes for correct DM visibility:

- **inbox --agent flag**: `agent-relay inbox` was registering as `__cli_inbox__` (a dummy agent that receives nothing). Added `--agent <name>` flag — registers as the named agent so relaycast returns its actual unread DMs. Defaults to `__cli_inbox__` for backwards compat.
- **history --to shows real DM history**: `history --to <agent>` (no `#` prefix) now registers as that agent, fetches DM conversations via `dms.conversations()`, and optionally drills into a specific thread with `--from`. Previously exited with an error.
- **history --from includes DM messages**: `history --from <agent>` was only querying the channel. Now also registers as the sender, walks their DM conversations, and includes messages they sent as DMs (labeled with `(DM → recipient)`).

## Test plan

- [x] `agent-relay inbox --agent <name>` registers as the specified agent
- [x] `agent-relay inbox` (no flag) still defaults to `__cli_inbox__`
- [x] `agent-relay history --to <agent>` shows DM conversation list
- [x] `agent-relay history --to <agent> --from <other>` shows thread between two agents
- [x] `agent-relay history --to #general` still works normally
- [x] `agent-relay history --from <agent>` shows channel + DM messages sent by that agent
- [x] All unit tests pass (`npx vitest run src/cli/commands/messaging.test.ts`)


xkonjin (commented):
## Code Review: inbox --agent flag + DM history

Clean feature addition with good test coverage. A few observations:

### Logic issue: Missing early return
In the `history` command, when `--to` is provided and doesn't start with `#`, the DM history branch has an early `return` for JSON and empty conversation cases, but the channel history code below it is not inside an `else` block. Looking again — actually the DM branch ends with `return`, so control flow is correct. Disregard.

### Suggestion: Validate --to format for DMs
When `--to` is an agent name (no `#` prefix), there's no validation that it looks like a valid agent name. Consider adding a regex check or length limit to fail fast with a helpful error if someone passes a channel ID by mistake (e.g., `--to C123456`).

### Test coverage gap
The tests mock `dms.conversations` and `dms.messages`, but there's no test for:
- The error path when `dmClient.dms.conversations()` throws
- The error path when `dmClient.dms.messages()` throws
- Invalid `--to` values (empty string, only whitespace)

### Security/UX consideration
The `agentName: options.to` is passed directly to `createRelaycastClient`. If `options.to` contains whitespace or special characters, this could cause issues downstream. Consider sanitizing: `agentName: options.to.trim()` (you do this for `inbox --agent`, but not for `history --to`).

### Consistency nit
`inbox --agent` uses `options.agent?.trim()` but `history --to` doesn't trim. Consider aligning them.

None of these are blockers — the PR is well-structured and the tests cover the core functionality.

PR #747, opened 2026-04-16
feat(sdk): add ProcessBackend workflow for cloud sandbox execution

khaliqgant:
## Summary

Adds `ProcessBackend` and `ProcessEnvironment` interfaces to the SDK, accepts `processBackend` in `WorkflowRunnerOptions`, and creates a ProcessBackend-backed `RunnerStepExecutor` when no explicit executor is provided.

## What changed

- Exports `ProcessBackend` and `ProcessEnvironment` from `@agent-relay/sdk/workflows`
- `WorkflowRunnerOptions` accepts an optional `processBackend`
- Agent and deterministic workflow steps can execute through `ProcessEnvironment.exec`
- `env`, `cwd`, and ceil-rounded `timeoutSeconds` are passed through structured exec options
- `RelayCast.createWorkspace()` API-key handling now guards the optional `apiKey` shape from the latest `@relaycast/sdk` types

## Boundary

| Relay owns | ProcessBackend provides |
|---|---|
| CLI command construction and args | `createEnvironment()` |
| auth env, cwd, timeout metadata | `exec(command, opts)` |
| workflow step lifecycle | `destroy()` |

`uploadFile()` remains part of the interface for future file asset staging and is not called by the current executor.

## PR feedback addressed

- Workflow commit step now checks out `feat/process-backend-runner` before committing and pushes that branch explicitly
- Workflow task prompt no longer contains the old contradictory "stored but unused" guidance
- Public comments/docs now describe the TS executor path accurately instead of implying broker-side remote spawning is already implemented
- ProcessBackend executor keeps the previously resolved `env` and `cwd` behavior

## Test plan

- [x] `npm --prefix packages/sdk run build`
- [x] `npm run build:packages`
- [x] `npm run build`
- [x] `npx vitest run --config vitest.config.ts src/workflows/__tests__/process-backend-executor.test.ts` from `packages/sdk`

### DIFF
--- a/.trajectories/active/traj_1776113772922_bc92f121.json
+++ b/.trajectories/active/traj_1776113772922_bc92f121.json
@@ -175,6 +175,29 @@
             "reasoning": "Codex review noted the new job used build:packages while publish runs npm run build. The CI guard now runs the full root build so root CLI/TypeScript failures are caught before merge."
           },
           "significance": "high"
+        },
+        {
+          "ts": 1776455373340,
+          "type": "decision",
+          "content": "Handled PR 747 feedback in tracked parent worktree: Handled PR 747 feedback in tracked parent worktree",
+          "raw": {
+            "question": "Handled PR 747 feedback in tracked parent worktree",
+            "chosen": "Handled PR 747 feedback in tracked parent worktree",
+            "alternatives": [],
+            "reasoning": "The cwd copy under .relay/workspace is not the git worktree; fixes must be applied to /Users/khaliqgant/Projects/AgentWorkforce/relay so the feature branch and PR contain them."
+          },
+          "significance": "high"
+        },
+        {
+          "ts": 1776455534306,
+          "type": "reflection",
+          "content": "PR 747 feedback fixes verified with SDK build, package build, full root build, and focused ProcessBackend executor tests",
+          "raw": {
+            "focalPoints": ["pr-747", "typescript", "process-backend"],
+            "confidence": 0.9
+          },
+          "significance": "high",
+          "tags": ["focal:pr-747", "focal:typescript", "focal:process-backend", "confidence:0.9"]
         }
       ]
     }
--- a/.trajectories/index.json
+++ b/.trajectories/index.json
@@ -1,6 +1,6 @@
 {
   "version": 1,
-  "lastUpdated": "2026-04-15T21:45:41.032Z",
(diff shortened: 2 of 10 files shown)


xkonjin (commented):
**Code review — ProcessBackend wiring workflow**

This is a workflow script that generates a PR, not a runtime change, so the risk surface is low. A few things worth tightening before it lands:

1. **Shell escaping in the workflow script** — the generated  uses a hand-rolled regex  to decide when to quote args. That's fragile for args containing spaces, unicode, or pipes. Prefer  or a proper shell-escape helper, or pass  and  as an array to  if the backend supports it.

2. **Timeout math** —  can round a 500ms timeout up to 1s, which is fine, but document that the backend's  is ceil-rounded from milliseconds.

3. **Missing  usage** — the task narrative mentions  as part of the interface but the workflow executor never calls it. If agent steps need file assets (e.g. context bundles), consider where  fits in .

4. **Broker still required?** — the workflow itself disables the fork change (). Make sure the final PR description is crystal clear that  is accepted but not yet wired into execution.

Nothing blocking, just polish.

xkonjin (commented):
**Code review — ProcessBackend wiring workflow**

This is a workflow script that generates a PR, not a runtime change, so the risk surface is low. A few things worth tightening before it lands:

1. **Shell escaping in the workflow script** — the generated `command` uses a hand-rolled regex `/^[a-zA-Z0-9._\\-\\/=]+$/` to decide when to quote args. That's fragile for args containing spaces, unicode, or pipes. Prefer `JSON.stringify(arg).slice(1, -1)` or a proper shell-escape helper, or pass `cmd` and `args` as an array to `env.exec` if the backend supports it.

2. **Timeout math** — `Math.max(1, Math.ceil(timeoutMs / 1000))` can round a 500ms timeout up to 1s, which is fine, but document that the backend's `timeoutSeconds` is ceil-rounded from milliseconds.

3. **Missing `uploadFile` usage** — the task narrative mentions `uploadFile` as part of the interface but the workflow executor never calls it. If agent steps need file assets (e.g. context bundles), consider where `uploadFile` fits in `executeAgentStep`.

4. **Broker still required?** — the workflow itself disables the fork change (`DO NOT change the fork at line 4038`). Make sure the final PR description is crystal clear that `processBackend` is accepted but not yet wired into execution.

Nothing blocking, just polish.

PR #756, opened 2026-04-20
fix(claude): harden folder-trust auto-accept + bypass injection

khaliqgant:
## Problem

Interactive PTY claude in a fresh cwd (e.g. Daytona `/project`) silently hangs on the "Do you trust the files in this folder?" dialog after completing an MCP tool call. MCP calls don't need filesystem trust, so the observed cloud behavior is:

- claude posts "Hi from claude!" to the channel (MCP, no trust needed) ✓
- claude tries to Write intro-claude.md, hits trust dialog, blocks silently
- runner gives up on the idle agent, step fails

Captured prompt from a live sandbox (claude-code 2.1.19):

```
Do you trust the files in this folder?
/project
Claude Code may read, write, or execute files contained in this directory.
This can pose security risks, so only use files from trusted sources.

Learn more

❯ 1. Yes, proceed
  2. No, exit

Enter to confirm · Esc to cancel
```

Three failure modes stacked:
1. `detect_claude_trust_prompt` matches the legacy "Yes, I trust this folder" wording but the modern wording is "Yes, proceed". The detector still happens to match via the `lower.contains("yes") && lower.contains("trust")` path (because "do you trust" + "Yes, proceed" appears in the same buffer), but it's incidental — any wording tweak will break it.
2. `claude_trust_buffer` is capped at 2500 bytes with a 2000-byte slide. The trust dialog lands ~1.5 KB into the stream after box-drawing + bracketed-paste escapes; a burst of ANSI output can slide the trust lines past the detector window before `strip_ansi` runs.
3. `--dangerously-skip-permissions` alone doesn't reliably suppress the trust dialog in 2.1.x. Empirically the dialog still renders on first filesystem tool use.

## Fix

Three reinforcing changes — one path failing doesn't strand the agent.

### 1. `helpers.rs::detect_claude_trust_prompt`

Explicitly match the modern numbered-menu layout in addition to the legacy keyword pattern. Keeps existing tests passing; adds regression tests with the exact captured prompt.

### 2. `wrap.rs::handle_claude_trust`

- Widen the trust buffer: `append_buf(buf, text, 6000, 5000)` (was 2500/2000).
- After sending Enter, sleep 120ms and send Enter again. Claude's selection TUI occasionally drops the first keystroke during render.

### 3. `worker.rs::spawn` (PTY branch)

Auto-inject `--permission-mode bypassPermissions` alongside `--dangerously-skip-permissions` when neither is already in the agent's args. `--permission-mode` is documented per-mode in claude's `--help` and has proven more reliable at suppressing the dialog than the dangerously-skip flag.

## Test plan

- [x] `cargo test --bin agent-relay-broker -- trust` — 14/14 pass, including two new regression cases:
  - `claude_trust_prompt_modern_wording` — full prompt captured from live sandbox
  - `claude_trust_prompt_modern_menu_without_trust_context` — ensures the menu shape alone doesn't false-positive without trust keywords
- [x] `cargo check --lib` clean
- [ ] After merge + SDK release: rerun `hi-interactive.ts` in cloud; claude-hi writes `intro-claude.md`.

## Companion cloud change

Cloud PR AgentWorkforce/cloud#221 pre-seeds `projects."/project".hasTrustDialogAccepted = true` in the snapshot's `.claude.json`. That's defense-in-depth — this relay PR is the actual fix, the cloud PR removes the prompt at the source for Daytona `/project` spawns.

---

<a href="https://app.devin.ai/review/agentworkforce/relay/pull/756" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/helpers.rs
+++ b/src/helpers.rs
@@ -2,7 +2,10 @@ use std::{
     collections::HashMap,
     ffi::OsStr,
     path::Path,
-    sync::atomic::{AtomicU64, Ordering},
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        Mutex, OnceLock,
+    },
     time::{Duration, Instant, SystemTime, UNIX_EPOCH},
 };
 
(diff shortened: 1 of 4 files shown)


khaliqgant:
Addressed: **--permission-mode is now feature-detected** before injection.

New helper `helpers.rs::cli_supports_flag(cli, flag)` probes `<cli> --help` once per (cli, flag) pair, caches the result, and does a word-boundary match on the flag name. `worker.rs::spawn` only pushes `--permission-mode bypassPermissions` if the installed claude advertises the flag.

Older claude installs (pre-2.0) silently fall back to `--dangerously-skip-permissions` only — no spawn failure, pre-PR behavior preserved. If the probe itself fails (missing binary, exec error), the flag is treated as unsupported so we skip injection rather than ship an arg we can't verify.

8 new tests cover: exact match, `=` form, word-boundary rejection of superstrings (`--permission-mode` vs `--permission-mode-legacy`), absence, empty help text, EOF boundary, bracketed synopsis (`[--flag]`), and cache-on-missing-binary returning false.

PR #757, opened 2026-04-20
fix(sdk): FileDb in-memory cache authoritative — fixes stale status after disk write failures

khaliqgant:
## Problem

\`JsonFileWorkflowDb\` re-reads the backing jsonl on every \`getRun\` / \`getStepsByRunId\`. When the filesystem denies writes — as it does in a cloud Daytona sandbox where the relayfile workspace ACL marks \`/project/.agent-relay/\` read-only — every state mutation is silently dropped and subsequent reads return the pre-failure row.

Observed failure: workflows pass every step, post to channels correctly, write every deliverable file, and still report \`status: 'running'\` to the caller. Concrete trace from a live cloud run:

\`\`\`
[workflow] completed
Workflow \"hi-interactive-workflow\" — COMPLETED
  5 passed, 0 failed, 0 skipped
[workflow 01:48] Shutting down broker...
Run status: running                      ← execute() returns stale row
FAILED                                   ← caller's status check fails
Bootstrap fatal error: ... bun exit 1    ← non-zero exit kills the run
```

The warn that fires in this scenario also confirms it:

```
[workflow] warning: failed to write run state to /project/.agent-relay/workflow-runs.jsonl
  — --resume will not be available for this run. Use --start-from instead.
  Error: EACCES: permission denied, open
```

## Fix

**The in-memory cache is now the single source of truth.** Every mutation updates the cache synchronously. The jsonl file becomes a best-effort persistence layer — used once at construction to replay state for `--resume`, and otherwise write-only. Runtime correctness no longer depends on disk writes succeeding.

Three pieces:

1. **Cache-first reads**: `getRun` / `getStepsByRunId` return from the in-memory map. `snapshot()` is only called once, at construction, to replay whatever exists on disk.

2. **Structured failure signal**: new `JsonFileWorkflowDbOptions.onWriteFailure` listener fires on every failed append. Callers (CLI, dashboard, cloud bootstrap) can aggregate / surface write failures beyond the once-per-process `console.warn`.

3. **Opt-in `$HOME` fallback**: `JsonFileWorkflowDbOptions.homeFallback = true` routes to `$HOME/.agent-relay/workflow-runs-<base>.jsonl` when the preferred path is unwritable. Default is `false` — strict path semantics, matching the pre-cache behavior (the existing `'should warn once when append fails'` test passes unchanged). \`--resume\` in cloud is tracked as follow-up work.

Plain-string construction is preserved: \`new JsonFileWorkflowDb(path)\` still works identically. Callers opt in to the new options via \`new JsonFileWorkflowDb({ filePath, onWriteFailure, homeFallback })\`.

## Test plan

- [x] New `packages/sdk/src/__tests__/file-db.test.ts` — 8 tests covering round-trip, **stale-disk cache consistency (the regression guard)**, strict-default memory-only, opt-in `$HOME` fallback, failure listener semantics, replay-on-construction for `--resume`, step cache, and `hasStepOutputs` path resolution.
- [x] Existing `resume-fallback.test.ts` (including `'should warn once when append fails'`) still passes — optimistic `writable=true` when the unwritable path is the only candidate preserves lazy-failure semantics that test relies on.
- [x] Existing `builder-resume-persistence.test.ts` passes.
- [x] Ran the full SDK vitest suite; the two failures that remain (`models.test.ts` package-entry, `workflow-runner.test.ts > should record review completion in trajectory…`) also fail on unmodified `main` — pre-existing flakes, not caused by this PR.
- [ ] After merge + release: rerun `hi-interactive.ts` in cloud; runner should return `status: 'completed'` and the bootstrap should not report FAILED.

## Follow-up (not in this PR)

The cloud orchestrator agent still has no `fs:write` scope on `/project/.agent-relay/`, so `--resume` can't round-trip even with this fix — the jsonl isn't persisted across processes. Three options: (1) grant the scope in the workspace ACL, (2) have the cloud bootstrap construct `JsonFileWorkflowDb` with `homeFallback: true`, (3) accept `--resume` as local-only. Captured in the trail trajectory; will revisit after this lands.

<!-- devin-review-badge-begin -->

---

<a href="https://app.devin.ai/review/agentworkforce/relay/pull/757" target="_blank">
  <picture>
    <img src="https://static.devin.ai/assets/gh-open-in-devin-review-light.svg?v=1" alt="Open in Devin Review">
  </picture>
</a>
<!-- devin-review-badge-end -->

### DIFF
--- a/packages/sdk/src/__tests__/file-db.test.ts
+++ b/packages/sdk/src/__tests__/file-db.test.ts
@@ -134,6 +134,30 @@ describe('JsonFileWorkflowDb', () => {
     expect(resolved).toContain(path.join('.agent-relay', 'workflow-runs-workflow-runs.jsonl'));
   });
 
+  // Regression for PR #757 Codex review feedback: the primary path's
+  // directory can be writable while the jsonl file itself is read-only
+  // (relayfile-mount chmods synced files to 0o444 while leaving the
+  // parent dir at 0o755). The old dir-only probe would accept the
+  // primary path, every append would lazy-fail, and homeFallback
+  // would never kick in despite the caller explicitly opting in.
+  it('opt-in homeFallback: true → read-only file with writable dir still falls back', () => {
+    const writableDir = path.join(tmpDir, 'project');
+    mkdirSync(writableDir, { recursive: true });
+    const primaryPath = path.join(writableDir, 'workflow-runs.jsonl');
+    writeFileSync(primaryPath, ''); // create the file so chmod targets it
+    chmodSync(primaryPath, 0o444); // file read-only; dir still 0o755
+
+    const db = new JsonFileWorkflowDb({
+      filePath: primaryPath,
+      homeFallback: true,
+    });
+
+    const resolved = db.getStoragePath();
+    expect(db.isWritable()).toBe(true);
+    expect(resolved.startsWith(os.homedir())).toBe(true);
+    expect(resolved).not.toBe(primaryPath);
+  });
+
   it('notifies onWriteFailure on every failed append', async () => {
     const dbPath = path.join(tmpDir, 'workflow-runs.jsonl');
     const failures: Array<{ err: unknown; filePath: string }> = [];
--- a/packages/sdk/src/__tests__/file-db.test.ts
+++ b/packages/sdk/src/__tests__/file-db.test.ts
@@ -1,0 +1,210 @@
+/**
+ * Tests for JsonFileWorkflowDb — in-memory cache is authoritative.
+ *
+ * Regression: before this file existed, `getRun` re-read the jsonl from
+ * disk on every call. If a write failed (EACCES in cloud, ENOSPC, etc.)
+ * the cache-less implementation would return stale data, which in turn
+ * caused `WorkflowRunner.execute()` to report a completed run as
+ * `status: 'running'` to callers.
+ */
+
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { chmodSync, existsSync, mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
+import os from 'node:os';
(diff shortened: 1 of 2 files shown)


khaliqgant:
Addressed Codex's P2 feedback on `resolveStoragePath`:

The non-last-candidate probe now also checks file-level writability when the jsonl already exists. The review scenario — writable directory but read-only file (exact shape of a relayfile-mount-synced file with workspace ACL enforcing `0o444` on the file while leaving the parent `.agent-relay/` at `0o755`) — now correctly throws out of the probe and falls through to the next candidate.

Behavior matrix:
| dir | file (if present) | outcome |
|---|---|---|
| writable | missing | accept |
| writable | writable | accept |
| writable | read-only | **try next candidate** ← review fix |
| read-only | n/a | try next candidate |
| last candidate | n/a | optimistic accept |

Regression test added: constructs the writable-dir + read-only-file combination and asserts the storage path resolves to the `$HOME` fallback, not the primary path. 18/18 tests in the file-db + resume suites pass.

khaliqgant:
Addressed Devin's two findings:

1. **Stale class-level JSDoc** — corrected to 'opt-in, default false' (matches the option's own JSDoc and the constructor default).

2. **Aliasing hazard** — legitimate. `InMemoryWorkflowDb` shallow-copies on insert; `JsonFileWorkflowDb` was storing the caller's reference directly. The runner keeps each inserted row in its own `stepStates` map and later mutates `state.row.status` directly before calling `updateStep`/`updateRun` — through shared-reference aliasing, that mutation was landing in the cache and bypassing the append + timestamp path, undermining the exact invariant this cache was built to enforce.

Both `insertRun` and `insertStep` now shallow-copy on insert, matching the reference implementation. Added a regression test that inserts, mutates the caller's object post-insert, and asserts the cache is unaffected.

19/19 pass across file-db + resume test files.

PR #758, opened 2026-04-20
feat(sdk): add onAgentActivityChanged hook

noodleonthecape:
## Summary
- add a first-class `onAgentActivityChanged` hook to `AgentRelay`
- derive activity state from existing broker delivery and lifecycle events
- add focused SDK tests and docs for rendering a simple thinking badge

## Semantics
- `active=true` when the SDK sees an in-flight delivery for an agent
- activity starts on `delivery_queued`, `delivery_injected`, `delivery_active`, or `delivery_ack`
- activity ends on `relay_inbound`, `agent_idle`, `agent_exited`, `agent_released`, or `delivery_failed` when no pending deliveries remain
- concurrent deliveries are tracked per agent so the hook does not flip inactive too early
- duplicate active transitions are suppressed when derived state does not change

## Testing
- attempted: `npm test -- --run packages/sdk/src/__tests__/agent-activity.test.ts`
- blocked in this environment because the repo build depends on `turbo`, which is not installed here

Closes #681

### DIFF
--- a/packages/sdk/README.md
+++ b/packages/sdk/README.md
@@ -38,6 +38,9 @@ const relay = new AgentRelay();
 // Event hooks
 relay.onMessageReceived = (msg) => console.log(`${msg.from}: ${msg.text}`);
 relay.onAgentIdle = ({ name, idleSecs }) => console.log(`${name} idle for ${idleSecs}s`);
+relay.onAgentActivityChanged = ({ name, active, pendingDeliveries }) => {
+  updateThinkingBadge(name, active ? `Thinking (${pendingDeliveries})` : 'Idle');
+};
 
 // Spawn agents using shorthand spawners
 const worker = await relay.claude.spawn({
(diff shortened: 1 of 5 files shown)


noodleonthecape:
Follow-up: I installed dependencies locally and verified the targeted SDK test passes with:

`cd packages/sdk && npx vitest run --config vitest.config.ts src/__tests__/agent-activity.test.ts`

Result: 6 tests passed.

noodleonthecape:
Good catch. I patched the PR to treat `delivery_ack` as a no-op for activity tracking, updated the idle test to start from `delivery_queued`, and added a regression test covering a late `delivery_ack` after `agent_idle`.

Verified with:
`cd packages/sdk && npx vitest run --config vitest.config.ts src/__tests__/agent-activity.test.ts`

Result: 7 tests passed.

willwashburn (approved):

PR #763, opened 2026-04-21
feat(sdk): widen @relayfile/sdk dep range (unblocks cloud 0.3.x)

khaliqgant:
## Summary

\`@agent-relay/sdk\` pins \`@relayfile/sdk\` at \`^0.1.2\`. Downstream cloud workspace packages (\`@cloud/core\`, \`@cloud/web\`, \`@cloud/relayfile\`, \`@cloud/specialist-worker\`) bumped to \`^0.3.0\` to pick up the \`contentIdentity\` field on write ops. The two caret ranges can't dedupe, so npm installs two copies of \`@relayfile/sdk\` (0.1.x at top level, 0.3.x nested), producing duplicate \`RelayFileClient\` type identity errors during \`tsc\` in cloud.

## Change

Widen \`@relayfile/sdk\` in \`packages/sdk/package.json\` from \`^0.1.2\` to \`>=0.1.2 <1\`. No runtime change — this package only imports \`RelayFileClient\` (provisioner/seeder.ts), whose interface is stable 0.1.x through 0.3.x. Patch bump 4.0.34 → 4.0.35.

## Why a range, not a peer dep

\`@agent-relay/sdk\` is widely consumed as a regular dep. Converting to peer would force every consumer to install \`@relayfile/sdk\` explicitly — a breaking distribution change. Range-widening in \`dependencies\` is backward-compatible: existing consumers keep getting a working install (latest 0.1.x if nothing else constrains it); consumers with their own 0.3.x pin dedupe against that instead of pulling a second nested copy.

## Unblocks

- [AgentWorkforce/cloud#231](https://github.com/AgentWorkforce/cloud/pull/231) currently carries a root \`overrides\` block as a workaround for this exact issue. After this PR merges + \`@agent-relay/sdk@4.0.35\` publishes, that override can be dropped.

## Test plan

- [x] \`npm install\` in this repo resolves clean — default picks 0.1.2 (unchanged).
- [x] \`npm run build -w @agent-relay/sdk\` passes.
- [ ] Publish 4.0.35.
- [ ] Downstream cloud drops the \`overrides\` block and types resolve to 0.3.x without the nested-install split.

## Followup (not this PR)

If peer-style distribution is desired long-term, that's a separate 5.x change: remove \`@relayfile/sdk\` from \`dependencies\`, add as \`peerDependencies\`, doc the breaking change in the changelog, consumers update their package.json. Out of scope here.

---

<a href="https://app.devin.ai/review/agentworkforce/relay/pull/763" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "agent-relay",
-  "version": "4.0.30",
+  "version": "4.0.34",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "agent-relay",
-      "version": "4.0.30",
+      "version": "4.0.34",
       "bundleDependencies": [
         "@agent-relay/cloud",
         "@agent-relay/config",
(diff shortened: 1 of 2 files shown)


khaliqgant:
Addressed Devin + Codex feedback in force-push 6b25da3a:

Reverted the premature version bump (kept at 4.0.34) per Option 1 in Devin's suggestion. Workspace consumers pin `@agent-relay/sdk: 4.0.34` exactly, and bumping the sdk package alone would have made npm fall back to the registry copy instead of linking locally. The `chore(release)` workflow already handles atomic version bumps across all workspace packages, so this PR is now purely the dep-range widening.

Diff is 1 file meaningful change: `packages/sdk/package.json` — `@relayfile/sdk: '^0.1.2'` → `'>=0.1.2 <1'`. Plus refreshed lockfile.

Build + install verified:
- `npm install` — workspace consumers link to local 4.0.34 correctly
- `npm run build -w @agent-relay/sdk` — clean

PR #783, opened 2026-04-24
fix(sdk): skip node:sqlite fallback under Bun runtime

khaliqgant:
## Summary

The codex and opencode CLI-session collectors (`packages/sdk/src/workflows/collectors/codex.ts`, `opencode.ts`) try `better-sqlite3` first and fall back to the Node 22+ `node:sqlite` builtin. **Bun does not implement `node:sqlite`** — attempting `await import('node:sqlite')` (codex) or `require('node:sqlite')` (opencode) both reject, but Bun also emits cosmetic stderr diagnostic output

```
error: Registry URL must be http:// or https://
Received: "node:sqlite"
```

*before* the promise rejects. The try/catch catches the throw, but the stderr line has already been written.

Under cloud's Bun-driven `.relay-standalone-wrapper.mjs`, this surfaces in `runner.log` after every successful workflow completion (between "PASSED" and the patch-generation step) and looks like a real error to anyone reading the logs.

## Fix

Detect the Bun runtime via `process.versions.bun` and skip the `node:sqlite` fallback when present. Under Bun, if `better-sqlite3` isn't available the collector returns null (empty `CliSessionReport`), which is the correct behavior — there's no other SDK path today that can read the state DBs from Bun.

## Example of the noise this fixes

From a prod cloud workflow run (`d4f9f0ab-80d2-4ab6-b426-26e14de11948`), in `runner.log`:

```
[workflow 01:24] Shutting down broker...
Run status: completed
PASSED
error: Registry URL must be http:// or https://
Received: "node:sqlite"
[bootstrap] 11 files changed, 0 files deleted since baseline
```

After this PR + next SDK publish + bump in cloud, the two `error:` lines go away.

## Test plan
- [ ] Merge
- [ ] Publish @agent-relay/sdk (Khaliq handles the version bump in publish step)
- [ ] Bump cloud's snapshot to pull the new SDK
- [ ] Re-run a workflow and confirm the stderr lines no longer appear after "PASSED"

---

<a href="https://app.devin.ai/review/agentworkforce/relay/pull/783" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/sdk/src/workflows/collectors/codex.ts
+++ b/packages/sdk/src/workflows/collectors/codex.ts
@@ -3,11 +3,7 @@ import os from 'node:os';
 import path from 'node:path';
 import { createRequire } from 'node:module';
 
-import type {
-  CliSessionCollector,
-  CliSessionQuery,
-  CliSessionReport,
-} from '../cli-session-collector.js';
+import type { CliSessionCollector, CliSessionQuery, CliSessionReport } from '../cli-session-collector.js';
 
 const require = createRequire(import.meta.url);
 const CODEX_HOME = path.join(os.homedir(), '.codex');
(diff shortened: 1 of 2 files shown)


xkonjin (commented):
**Code Review: fix(sdk): skip node:sqlite fallback under Bun runtime**

Smart, targeted fix. Bun + node:sqlite impedance mismatch is a real annoyance, and the `process.versions.bun` guard is the cleanest approach.

**What works:**
- The `isBunRuntime()` check is minimal, correct, and the right level of abstraction.
- Reusing it in both codex and opencode collectors is DRY enough (these are separate modules with separate imports, so it makes sense).
- The try/catch already catches the thrown import, but the cosmetic stderr still leaked. This cuts it off at the source.

**Nit:**
- Consider adding a tiny test case or at least a comment that the `isBunRuntime` guard should be revisited once Bun implements node:sqlite. It's not a blocker, but it prevents a future dev from removing what looks like a redundant guard.

No security or functional issues found. Approve with the nit above.

PR #785, opened 2026-04-24
fix(broker): unbreak Windows build

willwashburn:
## Summary

The Publish Package workflow [failed on its Windows broker build](https://github.com/AgentWorkforce/relay/actions/runs/24891319229/job/72884138536) because three unix-only APIs were called unconditionally:

- `tokio::signal::unix::signal(SignalKind::window_change)` — SIGWINCH — in `src/wrap.rs` and `src/pty_worker.rs`
- `tokio::signal::unix::signal(SignalKind::terminate)` — SIGTERM — in `src/main.rs::run_init`
- `get_terminal_size()` was `#[cfg(unix)]` but invoked without cfg guards from `pty_worker.rs:935` and `wrap.rs:1305`

On Windows, `tokio::signal::unix` doesn't exist, leaving `sigwinch`/`sigterm` with unresolved types that cascaded through the giant `tokio::select!` blocks as `E0277: the size for values of type [u8] cannot be known at compilation time` errors. `get_terminal_size` surfaced as `E0425: cannot find function ... in this scope`.

## Fixes — functional parity, not cfg-skips

- **`get_terminal_size()`** now uses `crossterm::terminal::size()` (TIOCGWINSZ on unix, `GetConsoleScreenBufferInfo` on Windows). Dropped the `#[cfg(unix)]` gate.
- **`wrap.rs`** polls `get_terminal_size()` every 200ms to detect resize. `crossterm::event::EventStream` was *not* viable — it reads `/dev/tty`/`CONIN$` and would steal keystrokes from the stdin passthrough that forwards user input to the child PTY. Polling uses a pure size query, no input consumption, imperceptible latency.
- **`pty_worker.rs`** removes the SIGWINCH handler entirely. The worker's stdout is a pipe, so `get_terminal_size()` always returned None and the arm was already a no-op on unix. Resize already flows in correctly via the `resize_pty` protocol frame from the broker (handled at the existing frame dispatcher).
- **`run_init`** splits SIGTERM: unix keeps `SignalKind::terminate()`; Windows uses `tokio::signal::windows::ctrl_shutdown()`. Ctrl+C continues to work cross-platform via the existing `tokio::signal::ctrl_c()` arm.

## Test plan

- [x] `cargo check --target x86_64-pc-windows-gnu` — clean (was: 9 errors)
- [x] `cargo build` on darwin — clean
- [x] `cargo test --lib` — 232 passed
- [ ] CI Windows MSVC build passes
- [ ] Manual: resize terminal while running `agent-relay-broker wrap ...` on macOS/Linux, verify child process sees new dimensions within ~200ms

---

<a href="https://app.devin.ai/review/agentworkforce/relay/pull/785" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -4864,7 +4864,9 @@ fn record_thread_history_event(history: &mut VecDeque<Value>, event: Value) {
 /// Uses `crossterm::terminal::size()`, which is cross-platform:
 /// TIOCGWINSZ on unix, GetConsoleScreenBufferInfo on Windows.
 fn get_terminal_size() -> Option<(u16, u16)> {
-    crossterm::terminal::size().ok().map(|(cols, rows)| (rows, cols))
+    crossterm::terminal::size()
+        .ok()
+        .map(|(cols, rows)| (rows, cols))
 }
 
 /// Detect Claude Code auto-suggestion ghost text.
--- a/src/main.rs
+++ b/src/main.rs
@@ -1172,7 +1172,13 @@ async fn run_init(cmd: InitCommand, telemetry: TelemetryClient) -> Result<()> {
     let mut lease_check = tokio::time::interval(Duration::from_secs(10));
     lease_check.set_missed_tick_behavior(MissedTickBehavior::Skip);
 
+    // Graceful-shutdown signal: SIGTERM on unix, Ctrl+Break/Close on Windows.
+    // `tokio::signal::ctrl_c()` is handled in its own select! arm below and
+    // works on both platforms.
+    #[cfg(unix)]
     let mut sigterm = tokio::signal::unix::signal(tokio::signal::unix::SignalKind::terminate())?;
+    #[cfg(windows)]
+    let mut sigterm = tokio::signal::windows::ctrl_shutdown()?;
 
     while !shutdown {
         tokio::select! {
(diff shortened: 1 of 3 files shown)


barryollama (dismissed):

xkonjin (commented):
**Code Review: fix(broker): unbreak Windows build**

Solid cross-platform fixes. A few notes:

1. **crossterm::terminal::size() swap — good.** The platform abstraction is cleaner than the raw libc/unsafe block.
2. **pty_worker.rs — removing SIGWINCH is the right call.** The worker stdout is a pipe, so the signal arm was already dead code on all platforms. The resize_pty protocol frame is the correct mechanism.
3. **wrap.rs — the 100ms Windows polling thread is pragmatic.** Consider bumping to 200ms to match the PR description (says 200ms, but code is 100ms). Either works, but consistency would be nice.
4. **Typo in PR description:** mentions `wrap.rs` polling at 200ms, but the code uses 100ms. Not blocking, but fix one or the other.
5. **Missing:** No tests added for the Windows path. The manual test checklist is fine, but automated regression coverage would be stronger against future platform breakage.

Overall: ship it after a minor test gap fix.

Issue #770, opened 2026-04-22
TS SDK: split broker binaries into per-platform optional-dependency packages


willwashburn:
## Problem

The `agent-relay` npm tarball currently bundles all four platform-specific broker binaries at its top-level `bin/` (`agent-relay-broker-darwin-arm64`, `-darwin-x64`, `-linux-x64`, `-linux-arm64` — each ~13–15 MB). Every user downloads ~55 MB of broker binaries and uses one.

Costs and constraints:
- **Bandwidth / install time** — 4× the binary payload every user has to download.
- **No room for Windows** — adding `win32-x64` pushes the bundle to ~70 MB; adding musl/arm32/freebsd compounds.
- **Postinstall fragility** — today, `scripts/postinstall.js` copies from the bundled top-level `bin/` into `packages/sdk/bin/agent-relay-broker` so `getBrokerBinaryPath()` can find it. When postinstall is skipped (`--ignore-scripts`, Bun, corporate policies) the binary becomes unfindable. This is exactly what bit Falko (see #768 for the mitigation: resolver now walks ancestor `bin/` dirs). The underlying fragility remains until postinstall is gone.
- **Coupling** — broker patches and CLI patches ride the same tarball. Shipping a broker-only fix means republishing ~60 MB.
- **Nothing Windows-friendly** — postinstall has macOS/Linux-specific logic (`xattr`, `codesign`), and the bundled binaries don't cover win32.

The industry-standard fix is what esbuild, `@swc/core`, `@next/swc-*`, `@parcel/*`, `lightningcss`, `sharp`, and `@tauri-apps/cli` all do: **platform-specific sibling packages declared as `optionalDependencies`, with `os`/`cpu`/`libc` constraints so the package manager installs exactly one.**

## Goal

After this change:
- `npm i @agent-relay/sdk` (or `bun add`, `pnpm add`, `yarn add`) installs the SDK plus exactly one broker binary package matching the user's platform.
- `getBrokerBinaryPath()` resolves via `require.resolve('@agent-relay/broker-<platform>-<arch>')`; no postinstall, no runtime download, no ancestor walking.
- `scripts/postinstall.js`'s broker logic can be deleted.
- Adding Windows (or musl, or arm32) is a one-line addition to the release matrix, not a bundle-size decision.
- Each user downloads ~13 MB of broker instead of ~55 MB.

### In scope
- New per-platform npm packages under `@agent-relay/broker-*`.
[...]

PR #788, opened 2026-04-24
feat!: unbundle @agent-relay/* so per-platform broker binaries actually install

willwashburn:
## TL;DR

`agent-relay@6.0.0` ships with a broken broker binary on every platform. This PR fixes that by unbundling the SDK and its dep tree from the CLI tarball — finally making the per-platform optional-dep pattern from #770 / #772 work as designed.

**This is a major version bump.** The CLI tarball drops from 24.2 MB to 1.5 MB.

---

## The actual production bug

`npm install agent-relay@6.0.0` ends up with empty broker bin dirs:

```
node_modules/agent-relay/node_modules/@agent-relay/broker-darwin-arm64/bin/
  -rw-r--r--  .gitkeep      ← 0 bytes, no binary
```

The separately-published `@agent-relay/broker-darwin-arm64@6.0.0` tarball does contain the real 11.7 MB binary — but it never reaches the user. Spawn fails at runtime.

## Why it broke

PR #772 introduced the per-platform optional-dep pattern (the same shape esbuild, sharp, swc, lightningcss, @next/swc all use). It worked for the SDK install path (`npm install @agent-relay/sdk`) but silently broke for the CLI install path. Three layers conspired:

1. **`bundledDependencies` and `optionalDependencies` are fundamentally incompatible** for the same chain of deps. Bundling freezes resolution at pack time; optional-deps require resolution at user-install time. PR #772 kept `@agent-relay/sdk` bundled while making the broker per-platform packages optional-deps of the SDK — the bundled SDK shipped with the broker packages frozen at their pack-time state (empty `bin/.gitkeep`), and npm refused to re-resolve them at install time because they were "already satisfied" via the bundle.

2. **`@agent-relay/hooks` depended on `@agent-relay/sdk`.** Even when I tried to remove SDK from `bundledDependencies`, hooks transitively pulled it back in. Cutting the chain required unbundling everything in the SDK's transitive closure.

3. **A duplicate legacy `bundleDependencies` field** (no 'd') sat at the bottom of `package.json` from before npm normalized the spelling. It was still listing every `@agent-relay/*` package and silently overriding edits to the new-spelling field above. This made the bug look unfixable from the obvious-place view.

### The main change: unbundle the SDK chain, embrace optional-deps

`@agent-relay/sdk` and its transitive `@agent-relay/*` deps (`cloud`, `config`, `hooks`, `telemetry`, `trajectory`, `user-directory`, `utils`) are no longer bundled into the CLI tarball. They install from the npm registry exactly like every other Node ecosystem tool with native binaries:

- `agent-relay` → declares `@agent-relay/sdk` as a regular dependency
- `@agent-relay/sdk` → declares `@agent-relay/broker-{darwin-arm64,darwin-x64,linux-arm64,linux-x64,win32-x64}` as optional-deps with `os`/`cpu` constraints
- `npm install agent-relay` → npm filters the broker optional-deps by the user's platform, installs exactly one to `node_modules/@agent-relay/broker-<plat>/`, broker binary lands at `bin/agent-relay-broker`
- SDK's `getBrokerBinaryPath()` resolves it via standard `require.resolve('@agent-relay/broker-<plat>/package.json')`

This is the esbuild / sharp / swc / lightningcss model. It's well-understood, the failure modes are well-known, and engineers debugging cross-platform binary install issues have prior art.

`@relaycast/sdk` and `@relayfile/local-mount` (external, not ours) stay bundled — they're not in the broker dep chain, so bundling is fine.

### Import surface (breaking change)

The previous `agent-relay/broker/<x>` subpath exports (which pointed at literal files inside the CLI tarball) are removed. Going forward there are exactly two ways to use the SDK:

```js
// Main entry — already exported by agent-relay's `src/index.ts`
import { AgentRelayClient } from 'agent-relay';

// Granular — @agent-relay/sdk is hoisted in node_modules as a
// transitive dep of agent-relay, so direct imports work without
// adding it to your own package.json
import { AgentRelayClient } from '@agent-relay/sdk';
import { AgentRelayClient } from '@agent-relay/sdk/client';
import { type BrokerEvent } from '@agent-relay/sdk/browser';
```

`@agent-relay/sdk` remains a published, public package — install it standalone if you don't want the CLI. This matches the convention every Node ecosystem package follows: one main package entry, granular paths from the actual sub-packages.

### Other cleanup included in this PR

- **Drop `packages/*/dist`, `packages/*/package.json`, `packages/sdk/bin`, `packages/sdk/README.md` from `files`.** Tarball ships only the CLI's own `dist/` plus runtime resources. 24.2 MB → 1.5 MB. 9197 files → 1160 files.
- **Delete `setupWorkspacePackageLinks` from `scripts/postinstall.js`** (~130 lines). It existed solely to bridge bundled-workspace-packages → `node_modules/@agent-relay/*` symlinks for global installs. With unbundling, npm's normal hoisting handles this.
- **Update JSDoc examples in `packages/sdk/src/{shadow,logs,browser}.ts`** to use `@agent-relay/sdk` (the canonical SDK import for SDK-source readers), not the removed `agent-relay/broker*`.
- **Update `verify-publish.yml` and `scripts/post-publish-verify/verify-install.sh`** to use the SDK's own `getBrokerBinaryPath()` resolver — the canonical way clients locate the binary. Replaces ~300 lines of stale path-probing checks that were written for the pre-#772 bundled layout (`node_modules/agent-relay/{bin,packages/sdk/bin}` — neither exists anymore).

### Adjacent CI fixes also included

- **Drop the `darwin-x64` smoke leg.** GitHub's `macos-13` (Intel) runner pool is so capacity-constrained that the smoke job was queueing for 2+ hours while every other smoke leg finished in under 90s. The `x86_64-apple-darwin` broker binary is still cross-compiled, published, and verified downstream by `Verify Standalone (macOS)` — we just don't run the end-to-end smoke on an actual Intel Mac anymore.
- **Fix PyPI publish retry logic.** `pypa/gh-action-pypi-publish` writes `<wheel>.publish.attestation` next to each wheel on its first attempt. If upload then fails, the attestation file is left behind and the next attempt aborts with `already have publish attestations`. Clear stale attestations between retries and add `skip-existing: true` so retries also no-op cleanly when the wheel did make it through.
- **Fix Windows path check in `verify-publish-sdk.yml`.** `p.includes(expected)` was comparing a backslash-separated path against a forward-slash package name. Normalize backslashes before the check. The resolver was always correct — false negative in the test.
- **Drop the `pull_request` trigger from `verify-publish.yml`.** These tests install the published `agent-relay` from the npm registry. With unbundling, a locally-packed PR-mode tarball can't install standalone (it references `@agent-relay/*@<ver>` not yet on the registry). Manual `workflow_dispatch` and `workflow_call` from `publish.yml` still work.

## Trade-offs we accept

- **First install needs network for transitive deps.** No more single-tarball offline install.
- **Race window between `publish-packages` and `publish-main` is now real.** Mitigated by `needs:` ordering in the workflow, but a few seconds of theoretical risk exist where someone could `npm i agent-relay@<new>` before `@agent-relay/sdk@<new>` propagates the CDN.
- **`agent-relay/broker*` import paths are removed** (breaking). Users importing from `agent-relay/broker/client` etc. need to switch to `@agent-relay/sdk/client`. Major version bump covers this.

## Verification

- Local pack of this branch: 1.5 MB tarball, 1160 entries, 0 validator violations.
- Simulated install layout: broker binary found via SDK resolver with no special-casing.
- Architecture matches industry pattern (esbuild, sharp, swc, lightningcss, @next/swc, parcel — all use unbundled per-platform optional-deps).

## Test plan

- [ ] Re-dispatch `Publish Package` after merge; confirm tarball at registry installs cleanly with non-empty broker bin dir.
- [ ] On each of macOS arm64, macOS x64, Linux x64, Linux arm64, Windows x64: `npm install agent-relay@<new>` in a scratch project, verify `node_modules/@agent-relay/broker-<plat>/bin/agent-relay-broker` exists, run `npx agent-relay --version` (loads CLI), run a spawn through the SDK to confirm broker binary actually runs.
- [ ] Verify `import { AgentRelayClient } from 'agent-relay'` and `import { AgentRelayClient } from '@agent-relay/sdk'` both resolve at type-check and runtime.
- [ ] Verify `npm install @agent-relay/sdk` (SDK-only path) still installs broker optional-deps correctly.
- [ ] Watch the Smoke Broker Packages and Verify Standalone macOS jobs in the next publish run; both should pass.


xkonjin (commented):
I like the direction here, especially moving verification onto the SDK resolver instead of hand-inspecting bin paths. One concern though: adding skip-existing: true to all three PyPI publish attempts can hide a real partial-publish failure. If attempt 1 uploads some artifacts and then fails mid-run, attempts 2 and 3 may quietly skip the already-seen files instead of proving the full release set is consistent. That makes the workflow more retry-friendly, but it also weakens the signal that the publish step actually recovered cleanly. It may be safer to scope skip-existing only where duplicate uploads are expected, or add an explicit post-publish verification that every expected artifact/version exists in PyPI before declaring success.

Test coverage also gets thinner on the darwin-x64 path. This PR removes a lot of direct binary/smoke assertions and replaces them with resolver-based checks, which is good, but I do not see an equivalent darwin-x64 verification for the new optional-dependency model. A minimal resolver test for that platform, or a short note on why dropping that smoke path is safe, would make this change more convincing.

PR #790, opened 2026-04-25
fix: bundle @agent-relay/cloud (not published at lockstep version)

willwashburn:
## Summary

`agent-relay@6.0.1` was published, but `npm install agent-relay@6.0.1` fails on every platform with:

```
npm error code ETARGET
npm error notarget No matching version found for @agent-relay/cloud@6.0.1
```

`@agent-relay/cloud` is the only workspace package whose npm namespace isn't owned by this repo's publish flow. The `publish-packages` job in `.github/workflows/publish.yml` ships every other internal package (sdk, hooks, config, telemetry, trajectory, user-directory, utils) but cloud is absent from the matrix. The latest registry version of `@agent-relay/cloud` is `2.0.23`, not 6.x.

Under the previous bundled-everything model this was invisible: the workspace cloud at the lockstep version got bundled into the agent-relay tarball and users never tried to fetch from registry. After #788 unbundled `@agent-relay/*`, agent-relay's regular dep on `@agent-relay/cloud@6.0.1` causes `npm install` to fail because that version doesn't exist on npm.

## Fix

Keep `@agent-relay/cloud` bundled. It's safe — cloud's only `@agent-relay` dependency is `@agent-relay/config` (also at 6.0.1, also published normally), and cloud is not in the SDK chain, so bundling it doesn't re-introduce the broker shadow-copy problem #788 just spent five commits eliminating.

## Verification

Locally packed this branch and installed into a scratch project. Tree:

```
node_modules/agent-relay/node_modules/@agent-relay/cloud/    ← bundled
node_modules/agent-relay/node_modules/@agent-relay/config/   ← bundled (transitively via cloud)
node_modules/@agent-relay/sdk/                               ← hoisted from registry
node_modules/@agent-relay/broker-darwin-arm64/bin/agent-relay-broker ← real Mach-O executable, hoisted via SDK optional-dep
```

Tarball size: 1.5 MB → 4.0 MB. Most of the increase is cloud's `@aws-sdk/client-s3` transitive deps. Still 6× smaller than the original 24.2 MB pre-#788.

## Long-term follow-ups

- Either publish `@agent-relay/cloud` from this repo at the lockstep version (figure out the namespace ownership story), or rename the workspace to avoid the collision. For now bundling restores correctness without picking a side.

## Test plan

- [ ] After merge + republish: confirm `npm install agent-relay@<new>` succeeds on a fresh machine and the broker spawns.
[...]

PR #791, opened 2026-04-25
fix(ci): re-add @agent-relay/cloud to publish-packages matrix

willwashburn:
## Summary

`agent-relay@6.0.1` shipped broken because installing it failed with:

```
npm error notarget No matching version found for @agent-relay/cloud@6.0.1
```

`@agent-relay/cloud` is a workspace in this repo (\`packages/cloud/\`) that was historically published to npm at the lockstep version. Commit \`c8c8dad5\` ("additional clean", Jan 27 2026) silently dropped \`- cloud\` from the \`publish-packages\` matrix in \`.github/workflows/publish.yml\`. The workspace kept getting version-bumped to stay in lockstep with every release but was never republished, so the registry's latest \`@agent-relay/cloud\` froze at \`2.0.23\` while the workspace climbed to \`6.0.1\`.

Under the previous bundled-everything architecture this was invisible — the workspace cloud got bundled into the agent-relay tarball at pack time, and users' \`npm install\` got it from the bundle without ever consulting the registry. After #788 unbundled \`@agent-relay/*\`, the CLI's regular dep on \`@agent-relay/cloud@<lockstep>\` began going to the registry like every other internal dep, where the lockstep version didn't exist.

## Fix

One line: re-add \`- cloud\` to the publish-packages matrix.

```diff
   matrix:
     package:
       - policy
       - memory
       - utils
       - trajectory
       - hooks
       - user-directory
       - config
+      - cloud
       - sdk
       - telemetry
       ...
```

## Why this is the right fix (not bundling)

The npm registry's \`@agent-relay/cloud@2.0.23\` declares its \`repository\` as this repo, \`packages/cloud\`. It's the same code base, just stale. There's no namespace collision and no other repo owns this name — the only thing that broke was the publish step.

[#790](https://github.com/AgentWorkforce/relay/pull/790) tried to fix this by adding \`@agent-relay/cloud\` back to \`bundledDependencies\`. That works around the symptom but freezes the bundling architecture for one workspace, costs ~2.5 MB of tarball size (cloud's @aws-sdk/client-s3 transitive deps), and reintroduces a special case we just spent #788 eliminating. Closing #790 in favor of this.

## Test plan

- [ ] Re-dispatch \`Publish Package\` after merge with version 6.0.2.
- [ ] Confirm \`Publish cloud\` job appears in the run and succeeds.
- [ ] Confirm \`@agent-relay/cloud@6.0.2\` shows up on the registry alongside the other \`@agent-relay/*@6.0.2\` packages.
- [ ] Confirm \`npm install agent-relay@6.0.2\` succeeds on a fresh machine and the broker spawns.

---

<a href="https://app.devin.ai/review/agentworkforce/relay/pull/791" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -884,6 +884,7 @@ jobs:
           - hooks
           - user-directory
           - config
+          - cloud
           - sdk
           - telemetry
           - acp-bridge


barryollama (approved):
LGTM

One-line fix restoring @agent-relay/cloud to the publish matrix. The CI-only change correctly addresses the root cause — the package stopped publishing at 2.0.23 while the workspace continued versioning to 6.0.1.

Approving the targeted fix over #790's approach of re-bundling, which would perpetuate tech debt and add 2.5MB to the tarball.

Good catch on the silent regression from c8c8dad5.

PR #804, opened 2026-04-29
fix(ci): pack github-primitive + workflow-types in smoke; publish workflow-types

khaliqgant:
## Summary

The v6.0.3 publish run ([25091201135](https://github.com/AgentWorkforce/relay/actions/runs/25091201135)) shipped a half-broken release: `agent-relay@6.0.3` is on npm but pinned to `@agent-relay/sdk@6.0.3`, `@agent-relay/cloud@6.0.3`, broker packages, etc. that were never published. Users now get \`ETARGET No matching version found for @agent-relay/cloud@6.0.3\` / \`@agent-relay/github-primitive@6.0.3\` on install.

Two underlying bugs:

1. **Smoke test couldn't install the SDK tarball.** \`smoke-broker-packages\` packs SDK + broker + config locally and \`npm install\`s them into a scratch project, expecting npm to resolve everything else from the registry. But \`@agent-relay/sdk@6.0.3\` added \`@agent-relay/github-primitive\` and already pinned \`@agent-relay/workflow-types\` at the about-to-be-published version. Those aren't on the registry yet at this point in the workflow → ETARGET → smoke fails on every platform → publish-broker-packages / publish-packages / publish-sdk-only all skipped on \`needs:\`. \`Publish Main Package\` has no such gate, so it shipped anyway, pointing at the missing deps.
2. **\`@agent-relay/workflow-types\` was never in the publish matrix.** \`npm view @agent-relay/workflow-types\` returns 404 — the package has never been on npm — but the SDK declares it as an exact-version hard dependency. Even after fix #1, installs would still fail on workflow-types until it's published.

## Fix

- Pack \`github-primitive\` and \`workflow-types\` in \`smoke-broker-packages\` alongside config, and pass their tarballs to \`npm install\` (both the main install and the linux-x64 negative smoke). The SDK now resolves entirely from local tarballs without registry round-trips.
- Add \`workflow-types\` to the \`publish-packages\` matrix.

The pack list is documented as needing to stay in sync with \`packages/sdk/package.json\` \`dependencies\`.

## Recovery

This PR fixes the workflow but doesn't republish the broken 6.0.3. Once merged, we still need to either (a) bump to v6.0.4 and re-run the publish workflow, or (b) deprecate \`agent-relay@6.0.3\` and republish the missing sub-packages at 6.0.3. (a) is cleaner.

## Test plan

- [ ] Manually run the publish workflow with \`dry_run=true\` against this branch and confirm \`smoke-broker-packages\` passes on all 4 platforms.
- [ ] Confirm \`publish-broker-packages\`, \`publish-packages\`, and \`publish-sdk-only\` are no longer skipped (assuming \`package=all\`).
- [ ] On the next real release, verify \`npm install agent-relay@<version>\` succeeds on a fresh machine.

---

<a href="https://app.devin.ai/review/agentworkforce/relay/pull/804" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -633,11 +633,15 @@ jobs:
           mkdir -p packages/sdk/bin
           (cd packages/sdk && npm pack --ignore-scripts --pack-destination "$TARBALLS")
           (cd "packages/${{ matrix.broker_pkg }}" && npm pack --ignore-scripts --pack-destination "$TARBALLS")
-          # The build job bumped every workspace to NEW_VERSION. The SDK's
-          # `dependencies` pins `@agent-relay/config` at that version, which
-          # is not on the registry yet at this point in the workflow — so
-          # pack it locally and install it alongside the SDK.
+          # The build job bumped every workspace to NEW_VERSION. The SDK pins
+          # several internal @agent-relay/* packages at that exact version,
+          # none of which are on the registry yet at this point in the
+          # workflow — so pack them locally and install them alongside the
+          # SDK. Keep this list in sync with packages/sdk/package.json's
+          # `dependencies` whenever a new internal dep is added.
           (cd packages/config && npm pack --ignore-scripts --pack-destination "$TARBALLS")
+          (cd packages/github-primitive && npm pack --ignore-scripts --pack-destination "$TARBALLS")
+          (cd packages/workflow-types && npm pack --ignore-scripts --pack-destination "$TARBALLS")
           ls -lh "$TARBALLS"
 
       - name: Install tarballs into scratch project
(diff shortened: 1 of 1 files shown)


barryollama (approved):
## 🦫 Barry Ollama PR Review

Thanks for this fix, @khaliqgant! This is a solid post-mortem and remediation. I've reviewed the GitHub workflow changes.

### ✅ What Looks Good

1. **Root cause analysis is thorough** — The description clearly explains the race condition between the smoke test and the publish workflow. The "Publish Main Package" running without proper gate checks while dependencies weren't yet published is a classic CI/CD trap.

2. **Adding `workflow-types` to the publish matrix** is the right fix. Missing internal dependencies in the publish matrix is a common oversight when monorepos grow.

3. **Packing internal deps locally for smoke tests** is the correct approach. Installing from tarballs avoids the registry lookup that was causing ETARGET errors.

4. **The comment updates** explaining why we pack these locally ("Keep this list in sync with packages/sdk/package.json's dependencies") will prevent future regressions.

### 📝 Code Review Notes

**`.github/workflows/publish.yml`**

| Line | Note |
|------|------|
| 638-650 | Good explanatory comment update — makes the "why" clear |
| 658-665 | Variable unpacking with `GITHUB_PRIMITIVE_TGZ` and `WORKFLOW_TYPES_TGZ` is clean |
| 898 | Adding `workflow-types` to matrix — essential fix ✅ |

### 💭 Suggestions (Non-blocking)

1. **Consider a check** to validate that the packed tarball list in `smoke-broker-packages` matches what's in `packages/sdk/package.json` at CI time. A simple grep/compare could catch future drift.

2. **The negative smoke test** (lines ~712-727) now installs all internal deps — this is correct because the test intent is "no broker binary" not "no internal pkgs". Good catch updating that too.

### 🎯 Approval

CI-only changes, fixes a real incident, test plan is reasonable. LGTM! 👍

/approve

PR #836, opened 2026-05-10
fix: resolve #802 — broker: add VT grid via alacritty_terminal (steal from ht, don't use libghostty)

### LINKED ISSUE #802: broker: add VT grid via alacritty_terminal (steal from ht, don't use libghostty)
## Problem

The broker currently has no real VT parser. We `strip_ansi` the raw byte stream and substring-match against text (`src/helpers.rs`, `src/pty_worker.rs`). That's fragile in real use:

- Alternate-screen TUIs and partial repaints can produce text that looks "right" after stripping but isn't actually on screen
- We can't ask "what's at row R, column C?" — only "does this substring appear somewhere in the recent stream?"
- Cursor position is invisible to us, so any wait condition that depends on cursor (#800's `Cursor` variant) is impossible
- We hand-rolled a `TerminalQueryParser` to fake DSR/DA1/CPR responses so children don't hang querying the terminal — a real VT parser handles this for free
- Unicode width on emoji/CJK is best-effort; cell alignment in snapshots is wrong

## Proposal

Add `alacritty_terminal` as a dependency and run a parser per PTY session. Keep raw bytes flowing to subscribers and the replay buffer untouched — the grid is for matching, snapshots, and ready-detection, not for displacing the existing pipe.

```rust
// src/pty_worker.rs (sketch)
use alacritty_terminal::{Term, event::VoidListener, vte::ansi::Processor};

let mut parser = Processor::new();
let mut term = Term::new(config, size, VoidListener);

// in the read loop, in addition to forwarding bytes downstream:
parser.advance(&mut term, &chunk);
// term.grid() and term.cursor() are now queryable
```

Unlocks:

- `WaitCondition::Cursor { row, col }` from #800
- Snapshot-as-text and snapshot-as-ANSI commands (`agent-relay-broker dump-pty <worker>`) for repro bugs
- Drop the hand-rolled `TerminalQueryParser` — alacritty_terminal answers DSR/DA1/CPR natively
- Cleaner ready detection: instead of substring-match-on-stripped-bytes, ask the grid directly

## Why alacritty_terminal and not libghostty / vt100

- **vt100 (Rust crate)**: lightly maintained, lags behind modern TUI behavior (synchronized output, ZWJ emoji width). Would work today, but we'd hit the wall the first time Claude updates its rendering.
- **libghostty-vt (Ghostty's parser, what ht uses)**: most complete parser available, but: (a) Zig 0.15.2 + CMake in our build chain on every CI runner across 5 platform targets, (b) unstable C API requiring SHA-pinning, (c) no Windows support — we'd ship two parsers anyway, (d) cgo-style FFI from Rust, panics across FFI are nasty, (e) ~6MB extra in every npm tarball. Capability we'd gain: sixel, kitty graphics, kitty keyboard protocol — none of which Claude/Codex/Gemini emit.
- **alacritty_terminal (Alacritty's parser)**: pure Rust, `cargo add` integration, production-grade (backs Alacritty itself), used by Zellij and others, modern Unicode + synchronized output handling, stable enough API. Cross-compiles cleanly to all our targets including win32. Covers 100% of the value for driving coding-agent CLIs.

Reference: `wezterm-term` is the other reasonable choice (more actively developed, slightly more features), but its API churns more. Default to alacritty_terminal unless we hit a concrete blocker.

## Files to touch

- `Cargo.toml` — add `alacritty_terminal`
- `src/pty.rs` — own a `Term` per session, advance the parser on every read chunk
- `src/pty_worker.rs` — query the grid for ready detection; drop or shrink `TerminalQueryParser`
- `src/helpers.rs` — `detect_cli_ready` becomes a thin wrapper over grid queries (text + cursor)
- New: `src/snapshot.rs` (optional) — render grid as plain text or ANSI for debug commands

## Effort

Medium. Main risk is API surface area and resize handling — alacritty_terminal expects `SizeInfo` updates on resize, which we already get from portable-pty. Tests can replay recorded byte streams from real Claude sessions and assert grid state.

## Why now

Prerequisite for the `Cursor` wait variant in #800. Pairs naturally with #801 (paced injection) — together they replace the fragile parts of our PTY interaction layer with the same primitives a real terminal uses.

## Related

- #800 — composable wait-conditions (the `Cursor` variant lands once this is in)
- #801 — escape-aware paced injection

SlncTrZ:
## Summary

fix: resolve #802 — broker: add VT grid via alacritty_terminal (steal from ht, don't use libghostty)

## Problem

**Severity**: `Medium` | **File**: `Cargo.toml`

Add the `alacritty_terminal` crate to enable a real VT100‑compatible parser and grid. This dependency is pure Rust, cross‑platform, and production‑grade (used by Alacritty itself). No additional features are required beyond the default set, as we only need the terminal emulation state, not rendering or windowing.

## Solution

Under `[dependencies]`, add:

## Changes

- `Cargo.toml` (modified)
- `src/pty.rs` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced"


willwashburn:
Pushed `4a4c05ad` addressing the remaining review feedback. Recap by reviewer:

**@coderabbitai**

- ✅ **(Major) Serialize `resize()` with parser advancement** — Fixed. `resize` now acquires `(processor, term)` in the same order the reader thread uses, holds both through `master.resize()` and `term.resize()`. Matching lock order also forecloses the deadlock class. Inline comment on the reader loop now documents the convention.
- ✅ **(Critical) `vte::Parser` vs `Processor` mismatch** — addressed in `43389b0` (switched to `alacritty_terminal::vte::ansi::Processor`).
- ✅ **(Critical) `Term::new` signature + by-value `SizeInfo`** — addressed in `43389b0` (3-arg `Term::new(Config::default(), &size, VoidListener)`; replaced `SizeInfo` with a local `GridSize` implementing `Dimensions`).
- ⏭️ **Add `vte` as an explicit Cargo.toml dep** — N/A in current code. No direct `use vte::...` anywhere; we go through `alacritty_terminal::vte::ansi::Processor` (alacritty re-exports vte at its module path), so there's no transitive-dep fragility to guard against.
- ⏭️ **(Nitpick) drop the `processor` field** — kept on purpose. The field is now load-bearing for the resize-race fix: `resize` takes the same lock the reader uses, which requires the lock to live somewhere accessible from `&self`. Doc comment updated to reflect this.

**@cubic-dev-ai**

- ✅ **(P1) Validate/clamp VT grid dimensions** — Fixed. New `GridSize::from_pty(rows, cols)` clamps each axis to `max(1, n)`. Used by both `spawn` and `resize`. Two regression tests: `grid_size_clamps_zero_dimensions_to_one` (unit) and `resize_to_zero_does_not_panic` (live PTY).

**@devin-ai-integration**

- ✅ **`SizeInfo` zero `cell_width`/`cell_height` overflow** — already self-resolved when the code switched to a `Dimensions`-based `GridSize` (no float math involved anymore). Confirmed in your follow-up comment.

`cargo fmt` clean. Full broker suite: **581 passed, 0 failed** (was 579 before this commit — +2 for the new clamp tests).

willwashburn (approved):

PR #872, opened 2026-05-17
broker: make detect_cli_ready grid-aware

### LINKED ISSUE #868: broker: make detect_cli_ready grid-aware
## Context

Follow-up to #802 / PR #836 (the alacritty VT grid integration) and #800 / PR #837 (the composable wait primitive).

Today `helpers::detect_cli_ready` runs against `strip_ansi(output)` — a best-effort recreation of "what would have been written if the terminal naively played the byte stream." It then substring-matches that text. PR #837 made it delegate to `wait::for_cli::*` for the per-CLI rules, but those rules still operate on the same stripped-byte view.

That view has known false-positive paths:

- A TUI that clears the screen (`ESC[2J`) and redraws produces stripped text containing **both** the cleared content and the redraw. Substring matches the cleared text.
- "Is `❯` on its own line at the prompt?" can't be expressed precisely against a byte stream — we approximate with regex on the last 10 lines.
- Cursor-aware conditions (`WaitCondition::Cursor` in #800 / PR #837) are still a stub because the grid lookup wasn't wired into the snapshot.

Now that the broker owns a real `Term` per session, the wait evaluator can query *the actual visible screen*.

## Proposal

Two changes, in order:

1. **Extend `WaitSnapshot` with grid access.** Add a `grid: Option<&dyn Grid>` (or a tiny trait wrapping the methods `wait` needs — `cursor_position`, `cell_at`, `screen_text`). Make `WaitCondition::Cursor` non-stub: evaluate `cursor.position == (row, col)` against `snap.grid`. Existing snapshot-style callers that don't have a grid keep passing `None` and `Cursor` stays unsatisfied for them.
2. **Refactor `detect_cli_ready` (and the per-CLI `wait::for_cli` builders) to query the grid.** Specifically:
   - Claude: "cursor is on a row whose only non-whitespace character is `❯` or `>`" — express via `Cursor` row + `cell_at(cursor_row, 1)` instead of regex on stripped tail.
   - Gemini: keep the compose-prompt text match (it's already cheap and unambiguous), but evaluate it against `term.screen_text()` (the *current* visible screen) instead of accumulated stripped bytes — drops the alt-screen false-positive class.
   - Codex / generic: same — `screen_text()` instead of stripped bytes.

The `wait::for_cli` builders stay the same shape; their `Text` conditions now match against `WaitSnapshot::grid_text()` (current screen) when a grid is available, and fall back to `WaitSnapshot::screen` (the stripped accumulated bytes) when it isn't. This keeps the offline / replay path working without a grid.

## Files to touch

- `src/wait.rs` — extend `WaitSnapshot`, fill in `Cursor` evaluation.
- `src/helpers.rs` — `detect_cli_ready` builds a grid-backed `WaitSnapshot` from `PtySession::screen_text()` + `cursor_position()` and evaluates the right `wait::for_cli` set against it.
- `src/pty_worker.rs` — pass the live `PtySession` (or the relevant snapshot fields) into `detect_cli_ready`.

## Effort

Medium. Most of the surface area is in the snapshot wiring; the per-CLI rules themselves become *simpler* once they can ask the grid.

## Test plan

- All 35+ existing `detect_cli_ready` tests must still pass (they use accumulated-bytes input — grid is `None`, fallback path active).
- New tests: drive a `Term` directly with recorded byte streams from real Claude / Gemini / Codex sessions, build a grid-backed `WaitSnapshot`, assert the cursor / cell predicates fire correctly. The `parse_into` helper in `src/pty.rs::tests` already does the parser-driving half.
- Regression: alt-screen TUI (e.g., `claude` in mid-menu) that previously tripped a false positive should now stay un-ready.

## Related

- #802 — parent issue ("`detect_cli_ready` becomes a thin wrapper over grid queries").
- #836 — landed the grid.
- #837 — landed `wait::WaitSet` + the per-CLI builders this proposal would make grid-aware (and would fill in the `Cursor` variant stub from that PR).
- Sibling follow-ups: TerminalQueryParser replacement, snapshot module.

willwashburn:
## Summary

Stacked on #837. Depends on #837 landing first.

Fixes #868.

- add a rendered-grid readiness snapshot built from `PtySession::screen_text()` and `PtySession::cursor_position()`
- route production startup readiness through the live VT grid instead of accumulated stripped bytes
- keep the relaycast boot gate strict: a prompt must appear after boot and be visible in the current grid
- keep the old accumulated-byte detector as test-only fallback coverage

## Testing

- `cargo fmt`
- `cargo check`
- `cargo test detect_cli_ready`
- `cargo test startup_gate`
- `cargo test`


willwashburn:
Superseded by the correctly stacked PR targeting #837: https://github.com/SlncTrZ/relay/pull/1

PR #879, opened 2026-05-18
broker(pty): replace TerminalQueryParser with alacritty EventListener

### LINKED ISSUE #867: broker: replace hand-rolled TerminalQueryParser with an alacritty EventListener
## Context

Follow-up to #802 / PR #836. The VT grid is now wired into every `PtySession`, but we still keep the hand-rolled `TerminalQueryParser` (`src/helpers.rs:303-377`) to answer DSR / DA1 / CPR queries that children send to their terminal. It works for the five sequences it handles, but it's a parallel state machine to the real VT parser we just added and it covers a tiny slice of what a real terminal answers.

PR #836 passes `VoidListener` to `Term::new`, which means alacritty *parses* query sequences but discards the responses. The replacement is to pass a real `EventListener` that captures terminal-query events and writes the responses back to the PTY's stdin.

## Proposal

Replace `VoidListener` with a `RelayEventListener` that owns an `mpsc::UnboundedSender<Vec<u8>>` connected to the existing `PtySession::writer`. Implement `EventListener::send_event` to translate the relevant alacritty events into bytes:

- `Event::PtyWrite(bytes)` — DSR (`ESC[5n` → `ESC[0n`), DA1 (`ESC[c` → `ESC[?1;2c`), DA2 (`ESC[>c`), CPR (`ESC[6n` → `ESC[<row>;<col>R`). alacritty fills these in based on real grid state, so CPR responses become *accurate* instead of always reporting `1;1`.
- Optionally forward `Event::Title(...)` / `Event::ResetTitle` as broker events on `/ws` so dashboard / `view` clients can display the agent's window title.
- Optionally forward `Event::ColorRequest` / `Event::TextAreaSizeRequest` (same pattern as `PtyWrite`).

Then:

- Delete `TerminalQueryParser`, `TerminalQueryState`, and `terminal_query_responses` from `src/helpers.rs`.
- Remove `let mut terminal_query_parser = TerminalQueryParser::default();` and its `feed` call site in `src/pty_worker.rs:180`.

## Files to touch

- `src/pty.rs` — new `RelayEventListener` type, wire it into `Term::new`, plumb the writeback channel into `PtySession::writer`.
- `src/helpers.rs` — delete `TerminalQueryParser` + tests.
- `src/pty_worker.rs` — drop the parser call site.

## Effort

Small / medium. ~80–120 lines net (the listener is more capable than the state machine it replaces). Main risk is making sure the writeback doesn't deadlock against the reader thread that already locks the same PTY writer — should be fine because writes are short and the listener fires from inside `processor.advance(...)` which holds `processor` and `term` locks but not `writer`.

## Test plan

- Unit: feed `ESC[5n` → assert `ESC[0n` lands at the writer.
- Unit: write text to row 3 col 5, then feed `ESC[6n` → assert the CPR response reflects the real cursor (`ESC[3;5R`), not the hardcoded `1;1` the old parser returned.
- Regression: spawn a CLI that issues a DA1 query at startup and confirm it doesn't hang.

## Related

- #802 — parent issue (this is the "drop or shrink `TerminalQueryParser`" follow-up).
- #836 — landed the alacritty integration this builds on.
- Sibling follow-ups: detect_cli_ready grid refactor, snapshot module.

willwashburn:
Closes #867.

## Summary
- Replaces `VoidListener` with a new `RelayEventListener` that captures alacritty `Event::PtyWrite` (DSR / DA1 / DA2 / CPR responses) and writes them back to the PTY's stdin via an `std::sync::mpsc` channel drained by a dedicated thread.
- CPR responses now reflect the real cursor position from the VT grid instead of the hardcoded `1;1` the hand-rolled parser always returned.
- Deletes `TerminalQueryParser`, `TerminalQueryState`, and `terminal_query_responses` from `helpers.rs`, plus all call sites in `pty_worker.rs`, `wrap.rs`, and the `main.rs` test module.
- Makes `Snapshot::from_term` generic over `EventListener` so offline snapshot tests still use `VoidListener` while live capture uses `RelayEventListener`.

## Why this shape
- `send_event` runs inside `Processor::advance` (sync), so the channel must be non-blocking. `std::sync::mpsc` keeps the listener `Sync`, never blocks, and avoids requiring a tokio runtime in `PtySession::spawn`. The drainer is a `std::thread` matching the existing reader-thread shape; it exits when `term` (and its listener / sender) is dropped at session teardown.
- The drainer takes the `writer` lock, but the reader thread does *not* hold `writer` while in `processor.advance`, and `resize` doesn't touch `writer` either — so no deadlock with the existing reader.

## Test plan
- [x] `cargo build` clean.
- [x] `cargo test` — 615 pass / 2 ignored / 0 fail.
- [x] `cargo clippy --all-targets -- -D warnings` clean.
- [x] New unit: `listener_answers_dsr_with_terminal_ok` — `ESC[5n` → `ESC[0n` lands at the writer.
- [x] New unit: `listener_answers_cpr_with_real_cursor_position` — `ESC[3;5H ESC[6n` → `ESC[3;5R` (proves CPR is grid-accurate, not hardcoded).


willwashburn:
Thanks for the reviews — pushed `71358b1d` addressing the actionable items.

**Fixed:**
- **[codex P1 — bounded writeback queue]** Switched `std::sync::mpsc::channel` → `sync_channel(128)` and the listener to `try_send`. A full queue logs a `tracing::warn!` and drops the response rather than growing unbounded if the drainer is stuck behind a blocked `write_all`. Listener stays non-blocking inside `Processor::advance`. (`src/pty.rs`)
- **[CodeRabbit Minor — missing DA1 regression test]** Added `listener_answers_da1_with_vt102_ident`. alacritty identifies as a VT102 (`ESC[?6c`) rather than xterm's `ESC[?1;2c`; test asserts the actual reply so DA1 regressions surface immediately. (`src/pty.rs`)
- **[codex P2 / CodeRabbit Major — absolute paths in trajectory metadata]** Replaced `/Users/will/...` in `.trajectories/index.json` `path` and `traj_piik8r6zu3i7.json` `projectId` with repo-relative values.

**Declined (with reason):**
- **[CodeRabbit Heavy lift — drainer/`write_all` share writer mutex]** Real concern, but the fix (route all writes through a unified priority queue so protocol replies preempt bulk stdin) is a separate refactor that touches every `write_all` caller and should not be bundled into this listener-swap PR. In practice the risk is bounded: the drainer's writes are tiny (≤8 bytes for DSR/DA1, ≤16 for CPR) so even if it has to wait behind a bulk stdin write, the wait is short and the bounded queue above prevents memory blowup during the wait. Will file a follow-up issue for the unified writer queue.

CI green again (cubic + CodeRabbit), 616 tests pass, clippy clean.

PR #882, opened 2026-05-18
refactor(broker): route all PTY writes through one FIFO drainer

willwashburn:
## Summary

Finding from review of the freshly-merged #867 / #879 work: `PtySession::write_all` and the alacritty listener's writeback path both took the same writer mutex, but the mutex only provides per-call mutual exclusion — not FIFO ordering across producers. A DSR/DA1/CPR reply emitted by the listener can land between two consecutive `write_all` calls (e.g. between an injection body and its trailing `\r`), splicing the reply into the middle of an otherwise-atomic user write.

## Before vs after

**Before:**

```rust
// listener path
self.tx.try_send(text.into_bytes())        // → drainer thread → writer.lock().write_all
// user path
self.writer.lock().write_all(bytes)        // → SAME writer, different path. Mutex provides
                                           //   exclusion per-call, NOT FIFO across producers.
```

**After:**

```rust
enum WriteMsg { Reply(Vec<u8>), UserInput(Vec<u8>) }

// listener path
self.tx.try_send(WriteMsg::Reply(bytes))    // non-blocking, drops on full
// user path
self.write_tx.send(WriteMsg::UserInput(b))  // blocking, backpressure
                                            //
// drainer thread (single owner of the writer):
while let Ok(msg) = write_rx.recv() {
    writer.write_all(msg.bytes())?;
    writer.flush()?;
}
```

## What changed

- New `WriteMsg` enum with `Reply` / `UserInput` variants.
- Single `sync_channel<WriteMsg>` (depth `WRITE_QUEUE_DEPTH = 128`, renamed from `WRITEBACK_QUEUE_DEPTH`).
- `RelayEventListener::send_event` wraps its bytes in `WriteMsg::Reply` — still non-blocking, still drops on full with the existing `tracing::warn!`.
- `PtySession::write_all` sends `WriteMsg::UserInput` blocking. Backpressure flows to the caller when the queue fills. Returns `Err` only on channel disconnect (drainer exited) — same fatal-for-session semantics as the prior synchronous error path.
- `PtySession::writer: Arc<Mutex<Box<dyn Write + Send>>>` field is gone — the drainer thread now owns the `Box<dyn Write + Send>` directly, no shared lock anywhere.
- Drainer thread loop matches on `WriteMsg::{Reply,UserInput}` and writes/flushes for both.

## Test plan

- [x] New `write_queue_preserves_fifo_across_user_and_reply` test: interleaves `UserInput` / `Reply` / `UserInput` / `Reply` on the channel and asserts the drainer pops them in the same order — the regression the original concern called out.
- [x] All 20 existing `pty::tests` continue to pass (helper updated to match on `WriteMsg`).
- [x] Full broker test suite: **617 passed, 0 failed**.
- [x] `cargo fmt --check` and `cargo clippy --all-targets --all-features -- -D warnings` clean.

## Behavior note worth flagging

`PtySession::write_all` is now **blocking** (sync channel `send`) where it previously was a quick mutex grab. Backpressure semantic: if the PTY's slave side is slow to drain, the worker's tokio task can block on `write_all`. That's arguably more correct — if the PTY is wedged, blocking the producer until the consumer catches up is right — but it's a behavior delta. No callers in `pty_worker.rs` seemed to rely on the previous quick-grab semantic (they all check the `Result`), so this should be transparent.

## Related

- Builds on the alacritty `EventListener` integration from #867 / PR #879.
- Found while reviewing that PR post-merge.

### DIFF
--- a/src/pty.rs
+++ b/src/pty.rs
@@ -20,13 +20,47 @@ use parking_lot::Mutex;
 use portable_pty::{native_pty_system, CommandBuilder, PtySize};
 use tokio::sync::mpsc;
 
-/// Upper bound on queued writeback responses (DSR/DA1/DA2/CPR replies).
+/// Upper bound on queued PTY writes (both terminal-query replies from
+/// alacritty AND user/injection writes from `PtySession::write_all`).
 /// Bounded so a misbehaving child that floods query sequences while the
 /// drainer is stuck on `write_all` cannot grow the queue without limit
 /// and OOM the broker. 128 entries is well past any real burst — a
-/// healthy terminal interaction sees fewer than a handful of pending
-/// replies at a time.
-const WRITEBACK_QUEUE_DEPTH: usize = 128;
+/// healthy interaction sees fewer than a handful of pending writes at
+/// a time.
+const WRITE_QUEUE_DEPTH: usize = 128;
+
+/// Single FIFO command for the PTY write drainer. Both query-reply
+/// writebacks (from `RelayEventListener`) and user-input/injection
+/// writes (from `PtySession::write_all`) go through the same queue so
+/// the drainer thread is the **only** thing that touches the writer.
+/// That guarantees a global FIFO ordering across both producers — a
+/// terminal-query reply can no longer splice between two consecutive
+/// user writes (e.g. between an injection body and its trailing `\r`).
+enum WriteMsg {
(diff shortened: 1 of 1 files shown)


willwashburn:
@copilot fix issues raised in this PR by comments

Issue #881, opened 2026-05-18
Fix telemetry: unify projects, add missing dimensions, extract keys from source


willwashburn:
## Summary

Audit of telemetry across `relay` (this repo) and `relaycast` (hosted side) revealed three classes of problems:

1. **Data is fragmented** across two PostHog projects with no shared identity, so we can't answer basic usage questions.
2. **Missing dimensions** mean even within each project we can't segment by harness, can't split CLI vs SDK, and can't count unique humans.
3. **PostHog keys are hardcoded in source**, so forks pollute our data and we have no dev/staging separation.

This issue tracks the full fix plan, prioritized.

---

## What we can answer today

| Question | Status | Notes |
|---|---|---|
| Total users of agent-relay | ⚠️ Partial | Two PostHog projects; client uses hashed machine-ID, server uses workspace-ID. No join. |
| **CLI vs SDK split** | ❌ No | SDK (`packages/sdk/`) emits zero telemetry. |
| Agents spun up | ✅ Yes (client) | Broker `agent_spawn` event in `src/telemetry.rs` + `src/main.rs:~1643`. |
| OS distribution | ✅ Captured, ❌ not surfaced | `os`/`os_version`/`arch` on every event (`packages/telemetry/src/events.ts:55-76`); no dashboard segments on it. |
| Messages sent | ✅ Yes | Client: `message_send`. Server: `usageRecords.messagesSent` + `messageLogs`. |
| **Which harnesses used most** | ❌ No | Broker tracks the *spawned child's* CLI (`normalize_cli_name`, `src/main.rs:6666`), not the *parent harness* (Claude Code vs Cursor vs Codex vs Gemini). |

---

## Gaps

1. **No SDK telemetry** — `packages/sdk/` has no `track()` calls. Can't measure SDK adoption.
2. **No orchestrator/harness identification** — broker doesn't record what's driving it.
3. **No install / update / first-run events** — `runInstallCommand` in `src/cli/lib/core-maintenance.ts` is silent. No adoption curve.
4. **No broker crash/panic telemetry** — Rust panics fire nothing.
5. **No unique-user metric on the server** — `distinct_id = workspace_id` (`relaycast/packages/server/src/lib/telemetry.ts:18`). 1 workspace with 10 humans looks like 10 workspaces with 1 human.
6. **No cross-workspace admin dashboard** — `/console/stats` (`relaycast/packages/server/src/routes/console.ts`) is per-workspace only.
[...]

PR #883, opened 2026-05-18
chore(telemetry): inject PostHog key at build time (P0.5 of #881)

willwashburn:
## Summary

Closes P0.5 of #881. Removes the hardcoded PostHog write key (`phc_2uDu01…`) from both the Rust broker and the TypeScript telemetry package, and wires the release pipelines to inject the key from a GitHub Actions repository **variable** at build time.

The motivation is **fork pollution + staging separation**, not secrecy — PostHog `phc_*` keys are write-only ingest keys, public-by-design (same category as a Sentry DSN or a Stripe publishable key). Today every fork that runs the test suite (and every dev who builds locally) ships events to our production project; that ends with this change.

### Rust (`src/telemetry.rs`)
- `POSTHOG_API_KEY` becomes `option_env!("AGENT_RELAY_POSTHOG_KEY")` — baked in at compile time when the env var is set, `None` otherwise.
- New private `TelemetryClient::disabled()` constructor consolidates every no-op path (env opt-out, prefs-file opt-out, missing build-time key) so they all behave identically.
- A quiet `tracing::debug!` breadcrumb is left when the build-time key is absent; no panic, no scary warning, no first-run notice.
- New unit test covers the empty-string-as-unset normalization.

### TypeScript (`packages/telemetry/src/posthog-config.ts`)
- Drops the `PROD_API_KEY` constant.
- `getPostHogConfig()` now reads `AGENT_RELAY_POSTHOG_KEY` from `process.env` at runtime; the existing `POSTHOG_API_KEY` override keeps working for local debugging / staging.
- Empty/whitespace values normalize to `null` so a blank value doesn't trip us into hitting PostHog with an invalid key.
- The package builds with `tsc` (no build-time `define`), so runtime env reading is the natural fit. The bun-compiled standalone binaries get the value baked in via `--define` (see below).

### GitHub Actions
- `.github/workflows/publish.yml`:
  - `AGENT_RELAY_POSTHOG_KEY: ${{ vars.POSTHOG_PROJECT_KEY }}` added to the unix and windows Rust broker build steps.
  - The standalone `agent-relay` and `relay-acp` `bun build --compile` steps now pass `--define="process.env.AGENT_RELAY_POSTHOG_KEY=\"${AGENT_RELAY_POSTHOG_KEY:-}\""` so end-user-shipped binaries embed the key without depending on user env.
- `.github/workflows/build-broker-binary.yml`: same variable mapping added to the `Build static binary` step.
- `Cross.toml`: `[build.env] passthrough = ["AGENT_RELAY_POSTHOG_KEY"]` so the aarch64 `cross` container sees the host env var the same way native `cargo build` does. (This name is the code-side contract; it stays `AGENT_RELAY_POSTHOG_KEY`.)
- CI test workflows are **intentionally left untouched** — we want them to run without the key so test events don't pollute the dashboard.

**Naming note:** the GitHub-side identifier is `POSTHOG_PROJECT_KEY` (unified across the relay, relaycast, and cloud repos so the user sets one value per repo). The code-side env var stays `AGENT_RELAY_POSTHOG_KEY` — workflows just map one to the other.

## Action required after merge

**Add `POSTHOG_PROJECT_KEY` as a repository *Variable* — Settings → Secrets and variables → Actions → *Variables* tab, NOT *Secrets*.** Without it, shipped artifacts will have telemetry disabled.

Why a **var** and not a **secret**:
- PostHog `phc_*` keys are write-only ingest keys — publishing them grants no read access to data, only the ability to send events. They're the same category as a Sentry DSN or a Stripe publishable key.
- Calling it a `secret` was misleading and made it less discoverable. The actual structural concern is fork pollution, and that's already solved two ways: (a) telemetry no-ops when the value is missing, and (b) GitHub gates both `vars` and `secrets` from fork PRs by default — neither leaks via untrusted contributors.
- Repository `vars` are visible to anyone with read access to the repo, which matches the key's "public-by-design" status and lets contributors see what's wired without admin access.

## Out of scope

This PR only addresses P0.5 of #881. P0, P1, and P2 are deliberately left for follow-ups. The relaycast extraction is a separate parallel PR.

## Test plan

- [x] `cargo build --bin agent-relay-broker` with no env set → binary has no `phc_*` strings and emits the no-op debug log.
- [x] `AGENT_RELAY_POSTHOG_KEY=phc_test cargo build --bin agent-relay-broker` → binary embeds `phc_test` (confirmed via `strings`).
- [x] `cargo test --lib telemetry` → 12/12 passing, including the new `posthog_api_key_treats_empty_as_unset` test.
- [x] `cargo test --lib` (full broker suite) → 253/253 passing.
- [x] `cargo clippy -- -D warnings` clean; `cargo fmt --check` clean.
- [x] `npm run build` for `@agent-relay/telemetry` → compiled `dist/posthog-config.js` contains no hardcoded key.
- [x] YAML lint on the two modified workflow files.
- [ ] Once the variable lands, kick off a dry-run publish to confirm the standalone bun-compiled binaries actually embed the key (verify with `strings`).
- [ ] After first real release, confirm events with the new `broker_version` start arriving in the dashboard.

### DIFF
--- a/.github/workflows/build-broker-binary.yml
+++ b/.github/workflows/build-broker-binary.yml
@@ -60,11 +60,16 @@ jobs:
         uses: taiki-e/install-action@cross
 
       - name: Build static binary
-        # AGENT_RELAY_POSTHOG_KEY is read by `option_env!` in src/telemetry.rs
-        # and baked into the binary. Leave it unset (no secret configured) to
-        # ship a broker with telemetry hard-disabled.
+        # POSTHOG_PROJECT_KEY is a GitHub Actions repository *variable*
+        # (Settings → Variables, NOT Secrets) because the PostHog ingest
+        # key is a public-by-design write-only key — same category as a
+        # Sentry DSN. We map it into the code-side env var name
+        # `AGENT_RELAY_POSTHOG_KEY`, which `option_env!` in
+        # src/telemetry.rs reads and bakes into the binary. Unset
+        # variable → telemetry ships disabled (acceptable for forks and
+        # pre-release pipelines).
         env:
-          AGENT_RELAY_POSTHOG_KEY: ${{ secrets.AGENT_RELAY_POSTHOG_KEY }}
+          AGENT_RELAY_POSTHOG_KEY: ${{ vars.POSTHOG_PROJECT_KEY }}
         run: |
           if [[ "${{ matrix.target }}" == "aarch64-unknown-linux-musl" ]]; then
             RUSTFLAGS="-C target-feature=+crt-static" cross build --release --target ${{ matrix.target }} --bin agent-relay-broker
--- a/.github/workflows/build-broker-binary.yml
+++ b/.github/workflows/build-broker-binary.yml
@@ -60,6 +60,11 @@ jobs:
         uses: taiki-e/install-action@cross
 
       - name: Build static binary
+        # AGENT_RELAY_POSTHOG_KEY is read by `option_env!` in src/telemetry.rs
+        # and baked into the binary. Leave it unset (no secret configured) to
+        # ship a broker with telemetry hard-disabled.
+        env:
+          AGENT_RELAY_POSTHOG_KEY: ${{ secrets.AGENT_RELAY_POSTHOG_KEY }}
         run: |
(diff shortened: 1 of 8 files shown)


willwashburn:
@copilot resolve the merge conflicts in this pull request

PR #902, opened 2026-05-18
Unify inbound dispatch through pending queue

### LINKED ISSUE #885: broker: unify inbound dispatch behind always-queue, mode = drain policy
## Context

Surfaced during review of PR #884 (sub-PR 2 of #864). Today the broker has two dispatch paths for inbound relay messages:

```rust
match gate_inbound_for_session_mode(...) {
    GateOutcome::Inject  => { /* Relay mode: straight to queue_and_try_delivery_raw */ }
    GateOutcome::Queued  => { /* Human mode: parked in per-worker pending queue */ }
    GateOutcome::WorkerMissing => { /* fall through to existing not-found accounting */ }
}
```

The current split was chosen to preserve byte-for-byte pre-#864 behavior for the common case (Relay mode = no behavioral change vs main as of #864 sub-PR 1). It works, but creates two parallel code paths reasoning about the same kind of state.

## Proposal

Collapse to one path: **every inbound message goes through the per-worker pending queue.** `SessionMode` becomes a *drain policy*, not a gate:

```rust
enum SessionMode {
    Relay,   // auto-drain inline as part of the same broker tick
    Human,   // hold until explicit /flush or human → relay transition
    Paused,  // (future) hold indefinitely, no auto-drain even on mode flip
}
```

In `Relay` mode the broker enqueues then immediately drains in the same `tokio::select!` iteration — net latency roughly unchanged (~µs) because no actual yield happens between enqueue and drain.

`PendingRelayMessage` keeps the routing-metadata structure landed in PR #884 (`target`, `thread_id`, `workspace_id`, `workspace_alias`, `priority`, `mode`) so the uniform queue replay reproduces the original delivery byte-for-byte.

## Why this is interesting (not why it's urgent)

This is **architectural cleanup**, not a correctness fix. The races you might worry about are already handled:

| Race | Already protected by |
|------|---------------------|
| Two `/api/send` calls + an inbound relaycast at the same instant | `run_broker`'s single `tokio::select!` serializes events at the broker |
| Two `deliver_relay` frames at the worker | Worker stdin pipe is a serial FIFO |
| User-write splicing with alacritty's writeback at the PTY | PR #882's FIFO drainer (single writer-owner) |

So the proposal isn't motivated by a bug. The case for it:

1. **One dispatch path** is simpler to reason about than two.
2. **`GET .../pending` becomes universal** — today it only shows human-mode workers; with always-queue it shows every inbound on its way through, which is useful for "did this message get delivered?" debugging.
3. **Foundation for future features.** Audit log, dedup, ordering / prioritization beyond the existing `priority` field, message TTL, `paused` mode — all become easier when there's one buffer to reason about.
4. **Multi-driver coordination.** When #864 sub-PR 3 (`drive` client) ships and two clients can both drive the same agent (CLI + web dashboard + Mac app), "who owns the queue, who can flush" semantics need a single authoritative buffer to define against.

## Costs

1. **Latency.** Today: `/api/send` → broker tick → inject. Proposed: `/api/send` → broker tick → enqueue → drain → inject. Should be the same wall-clock cost (no extra yields) but it adds two hot-path operations per message.
2. **Two queue concepts already exist.** `pending_deliveries` (in-flight deliveries awaiting echo verification) + `pending` (per-worker inbound waiting for drain). They do different things but the naming overlap will trip people up. Worth renaming one or both as part of this work.
3. **Memory** — bounded by the existing per-worker cap (`MAX_PENDING_PER_WORKER = 256`), but every Relay-mode message touches the queue even when it'd drain in 1 ms.

## When to actually do this

**Not now.** Three reasons:

1. The races this would "fix" are already handled (`tokio::select!` + worker FIFO + PR #882's PTY drainer). No correctness pressure.
2. PR #884 is already large; bundling architectural unification on top would bury the routing-metadata fix it was scoped around.
3. **Sub-PR 3 (`drive` client) is the natural forcing function.** When real concurrent drivers exist (CLI + web + Mac app), the "who owns the queue" semantics need a single authoritative buffer. Doing the unification then, informed by actual usage patterns, beats guessing now.

So the right time is *after* sub-PR 3 ships and we've seen real drive sessions in use.

## Out of scope

- Persistence of the pending queue across broker restart (current PR doesn't persist; this unification wouldn't either).
- Cross-worker queue inspection / global pending view (per-worker is the right granularity for both today's use case and the unified version).
- A `paused` mode itself (the proposal mentions it as a natural fit once the unification lands; landing `paused` is its own decision).

## Files this would touch

- `src/main.rs` — collapse `GateOutcome::{Inject,Queued}` to a single enqueue, follow with conditional inline drain based on mode.
- `src/types.rs` — `SessionState` already holds the queue; minor restructuring at most.
- `src/listen_api.rs` — `GET .../pending` semantics broaden (now shows transient relay-mode entries too, which raises a question: should they be filterable?).
- Possibly rename `pending_deliveries` → `in_flight_deliveries` to disambiguate from the inbound `pending` queue.

## Related

- PR #884 — sub-PR 2 of #864; introduced the current two-path dispatch.
- #864 — parent issue (the broker-owned agent model).
- Sub-PR 3 of #864 — `drive` client; will be the forcing function for this work.
- PR #882 — PTY-level FIFO drainer; closed the only real "messages overwriting" race the codebase actually had.

willwashburn:
## Summary
- Route both /api/send and relaycast inbound deliveries through the per-agent pending queue.
- Treat auto_inject as an immediate drain policy while manual_flush keeps holding for flush or mode transition.
- Update queue semantics docs, changelog, and focused queue tests.

Closes #885

## Tests
- cargo check
- cargo test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Broker inbound delivery now uses one per-agent queue so `auto_inject` and `manual_flush` handle ordering consistently.
 - Inbound delivery APIs use `/delivery-mode` with `auto_inject` and `manual_flush` names before their first release.
 - CLI attach commands share SDK-backed broker snapshots, delivery mode changes, streams, and flushes.
 - PTY terminal query replies use the live VT grid, so cursor-position responses reflect the actual screen.
--- a/src/listen_api.rs
+++ b/src/listen_api.rs
@@ -152,7 +152,8 @@ pub enum ListenApiRequest {
         reply: tokio::sync::oneshot::Sender<Result<SetInboundDeliveryModeOk, DeliveryRouteError>>,
     },
     /// `GET /api/spawned/{name}/pending` — snapshot the per-worker
-    /// pending-message queue (FIFO, head first).
+    /// pending-message queue (FIFO, head first). Auto-inject workers usually
+    /// report an empty queue because they drain in the same broker turn.
     GetPending {
         name: String,
         reply: tokio::sync::oneshot::Sender<Result<Vec<PendingRelayMessage>, DeliveryRouteError>>,
@@ -1165,11 +1166,11 @@ async fn listen_api_snapshot(
 }
 
 // ---------------------------------------------------------------------------
-// Inbound delivery mode (per-agent inject vs. queue, plus pending-queue inspection)
+// Inbound delivery mode (per-agent drain policy plus pending-queue inspection)
 //
-// The broker keeps an `InboundDeliveryMode` per worker; `manual_flush`
-// mode parks inbound relay messages in a FIFO `pending` queue instead
-// of injecting them.
+// The broker keeps an `InboundDeliveryMode` per worker. All inbound relay
+// messages pass through a FIFO `pending` queue; `auto_inject` drains it
+// immediately, while `manual_flush` parks messages until the caller flushes.
 // These four routes are the server-side surface the `agent-relay drive`
 // client calls to flip modes, inspect the queue, and drain it.
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


willwashburn:
Addressed the CodeRabbit review on the pending-drain injection path in d69863b9: `try_inject_pending_relay_message` now bounds `queue_and_try_delivery_raw` with `timeout(...)` and returns a timeout error instead of being able to stall flush/mode-transition drains indefinitely. Also reworded the changelog bullet to be impact-first.

PR #929, opened 2026-05-20
Add streaming PTY input API

### LINKED ISSUE #924: Expose a low-latency streaming PTY input API
## Context

Pear currently drives broker PTYs through the SDK's `sendInput`, which maps to `POST /api/input/{name}`. For interactive terminal typing, every keypress goes through:

`renderer -> Electron IPC -> main process -> SDK HTTP request -> broker -> worker stdin -> PTY echo -> broker websocket event -> renderer store -> xterm.write`

That makes terminal input feel noticeably more sluggish than a direct local PTY, especially in Electron where keyboard events are frequent and the visible echo only arrives after the broker round trip.

Pear can queue and fire-and-forget input on its side, but that still ultimately becomes HTTP requests into `/api/input/{name}`. The more effective fix appears to be a broker/SDK-level persistent input path.

## Request

Could Relay expose a low-latency streaming PTY input channel for dashboard/Electron clients?

Possible API shapes:

- Extend the authenticated dashboard `/ws` connection to accept inbound frames, e.g. `{ "type": "pty_input", "name": "codex-1", "data": "..." }`.
- Add a dedicated authenticated websocket such as `/api/input/{name}/stream`.
- Add an SDK helper such as `openInputStream(name)` or `sendInputStream(name, data)` that keeps a connection open and preserves per-agent ordering/backpressure.

## Requirements / questions

- Preserve input ordering per agent.
- Handle backpressure or at least avoid unbounded buffering during paste/key repeat.
- Return/emit a clear error when the agent is missing or exits.
- Work with existing API-key auth.
- Avoid one HTTP request per keypress while still supporting larger paste chunks.
- Clarify whether bidirectional use of the existing `/ws` dashboard stream is acceptable, or whether input should be isolated on a dedicated websocket.

## Why

This would let GUI clients wire xterm input to Relay without routing every typed byte through a separate request/response path. It should make embedded terminals feel much closer to native PTY latency while keeping Relay as the source of truth for agent I/O.

willwashburn:
## Summary
- Add authenticated GET /api/input/{name}/stream websocket PTY input with ready, ack, and error frames.
- Add TypeScript SDK openInputStream() with ordered writes and bounded client-side buffering/backpressure.
- Document the broker API and SDK helper, update OpenAPI, changelog, and trajectory notes.

Fixes #924

## Tests
- cargo test -p agent-relay-broker listen_api
- npm --workspace @agent-relay/sdk run build
- npx vitest run --config vitest.config.ts src/__tests__/transport.test.ts
- npx vitest run --config vitest.config.ts src/__tests__/orchestration-upgrades.test.ts -t "AgentRelayClient orchestration payloads"


willwashburn:
@copilot resolve the merge conflicts in this pull request

PR #981, opened 2026-05-25
fix(broker): paced, escape-aware PTY injection

### LINKED ISSUE #801: broker: escape-aware paced injection (steal from ht)
## Problem

Today, the broker injects messages into a CLI by writing the entire payload in one `write_all`, sleeping 50ms, then writing `\r` (`src/pty_worker.rs:788–798`). That works most of the time, but we have a recurring class of bug where Claude appears to batch or eat characters during injection — likely a consequence of writing too fast for the child to process between read calls.

Symptoms include: occasional dropped characters at the start of an injection, chat input that arrives as a single garbled blob, and inconsistencies under heavy concurrent load.

## Prior art

[montanaflynn/headless-terminal](https://github.com/montanaflynn/headless-terminal) defaults to **pacing**: writing one keystroke at a time with a 20ms gap (configurable `--rate`, `--rate 0` disables). The clever bit is `escapeSeqEnd` in `internal/session/session.go` — the pacer is *escape-aware*, so VT control sequences (CSI `ESC [ ... cmd`, SS3 `ESC O cmd`, OSC `ESC ] ... ST`) and multi-byte UTF-8 codepoints are kept atomic instead of being split mid-sequence. Without that, you'd corrupt arrow keys, modified function keys, etc.

## Proposal

Port `escapeSeqEnd` to Rust and add `write_paced` next to the current `write_all` path. Replace the bulk-write-then-sleep injection in pty_worker with paced writes. Keep a knob (`paced_inject` or `inject_rate_ms`) so we can toggle it off if we discover a workload that needs raw speed.

```rust
// src/pty.rs (sketch)
pub async fn write_paced(&self, bytes: &[u8], rate: Duration) -> io::Result<()> {
    let mut i = 0;
    while i < bytes.len() {
        let end = next_atom_end(&bytes[i..]);  // ports escapeSeqEnd
        self.write_all(&bytes[i..i + end]).await?;
        i += end;
        if i < bytes.len() { tokio::time::sleep(rate).await; }
    }
    Ok(())
}
```

`next_atom_end` returns the length of the next "atom" — one ASCII char, one CSI/SS3/OSC sequence, or one UTF-8 codepoint.

## Files to touch

- `src/pty.rs` — add `write_paced` + `next_atom_end` helper
- `src/pty_worker.rs:788–798` — replace `write_all` + `sleep(50ms)` + `\r` with `write_paced(payload + b"\r", 20ms)`
- Tests: feed a payload containing CSI/SS3/OSC + multi-byte UTF-8 and assert the chunks emitted are atomic

## Effort

Small. ~50 lines of logic + tests. Pure Rust, no new deps.

## Why now

Likely fixes the "Claude occasionally batches/eats characters" class of bug we have today. Also makes our injection observably closer to a real human typing, which keeps Claude (and any other prompt-detection in CLIs) on its happy path.

willwashburn:
## Summary

Closes #801. Replaces the broker's bulk-write injection (`write_all(payload)` → `sleep(50ms)` → `write_all("\r")`) with a paced, escape-aware writer that emits one atom at a time — fixing the "Claude occasionally batches/eats characters" class of bug we see when injecting into a busy CLI.

- New `next_atom_end` in `crates/broker/src/pty.rs` — port of `escapeSeqEnd` from [montanaflynn/headless-terminal](https://github.com/montanaflynn/headless-terminal). Returns the length of the next atom: a single ASCII byte, a full UTF-8 codepoint (1–4 bytes), or a complete CSI / SS3 / OSC sequence. Truncated sequences are returned as one chunk so partial writes still make forward progress.
- New `PtySession::write_paced(bytes, rate)` — walks `next_atom_end` and enqueues each atom through the existing FIFO write queue with `rate` between atoms. Honours `Duration::ZERO` (no sleep, still atom-by-atom). Interleaves correctly with terminal-query replies because it uses the same queue.
- `pty_worker.rs` injection site now calls `write_paced(body + b"\r", 20ms)` instead of the bulk write + sleep + `\r` pair. Constant `INJECT_PACE_MS = 20` lives next to the other module constants.

The 20 ms cadence matches the headless-terminal default — fast enough to feel instant, slow enough that the child reliably reads each chunk between writes.

## Test plan
- [x] `cargo build -p agent-relay-broker` — clean
- [x] `cargo clippy -p agent-relay-broker --lib --tests -- -D warnings` — clean
- [x] `cargo test -p agent-relay-broker --lib pty::` — 33 pass (11 new, covering plain ASCII, CSI, SS3, OSC (BEL + ST), multi-byte UTF-8, mixed payloads, truncated CSI, lone ESC, and an end-to-end live-PTY round-trip through `cat` with a 4-byte codepoint)
- [x] `cargo test -p agent-relay-broker --lib pty_worker::` — 6 pass (no regressions)
- [ ] Smoke-test against a real Claude/Codex worker under load (recommend before merge)

https://claude.ai/code/session_01EWZheXxPE8ykajhPzqMn1H

---
_Generated by [Claude Code](https://claude.ai/code/session_01EWZheXxPE8ykajhPzqMn1H)_

### DIFF
--- a/crates/broker/src/pty.rs
+++ b/crates/broker/src/pty.rs
@@ -8,6 +8,7 @@ use std::{
         mpsc as std_mpsc, Arc,
     },
     thread,
+    time::Duration,
 };
 
 use alacritty_terminal::event::{Event, EventListener};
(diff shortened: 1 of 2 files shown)


willwashburn:
Closing without merging.

On reflection, the cost/benefit doesn't hold up:

**What it gets us**
- Escape-sequence atomicity on the wire. But `write_all` already loops until flushed, and the child's VT parser handles fragmented reads — so this only matters if you believe the child has a read-mid-sequence bug. That's the issue's hypothesis, not a measured fact.
- "Looks like typing" — anecdotal, not demonstrated here.

**What it costs us**
- Per-injection latency goes from ~50 ms (bulk write + 50 ms sleep + `\r`) to **20 ms × atom_count**. A 200-byte body becomes ~4 s; a 1 KB body becomes ~20 s before the trailing `\r` lands. That stretches verification windows and any downstream throttle timing.

The reference (`headless-terminal`) is a testing tool whose job is to *simulate a human*. The broker's job is to *deliver a message* — different goals, and we shouldn't pay typing-speed latency for it.

Better path: wait until we can reproduce the "Claude eats characters" bug and confirm bulk writing is actually the cause. If it is, revisit with either a much faster rate (1–2 ms or atom-by-atom with no sleep) or a flag-gated knob, with evidence to justify the trade.

---

PR #1075, opened 2026-06-10
fix(cli): capture orchestrator harness for the detached broker

willwashburn:
## Problem

[relay#1069](https://github.com/AgentWorkforce/relay/pull/1069) made the broker forward its detected harness to the relaycast backend, but server telemetry (PostHog 296966) **still records `harness: "unknown"` for ~100% of Rust-broker traffic** — verified on v8.3.2/v8.3.3 brokers (`@relaycast/sdk-rust 2.3.0`).

Root cause: the broker is launched **detached** (its own session via `setsid` — `broker-lifecycle.ts` spawns it with `detached: true`). By the time the broker runs `detect_orchestrator_harness`, its parent is the init process, so the parent-PID walk finds no `claude-code` / `codex` / `cursor` ancestor. #1069's forwarding works correctly — it just forwards `"unknown"`.

The asymmetry this explains:

| Path | Runs… | Detection | Result |
|---|---|---|---|
| JS / MCP (`@relaycast/sdk`) | in-process under the harness | parent walk sees the harness ✓ | `claude-code`, `codex` ✅ |
| Rust broker daemon | detached, own session | parent walk sees `launchd`/init ✗ | `unknown` ❌ |

## Fix

The `agent-relay` CLI **is** still in-process under the harness when `up` is invoked. Detect the harness there and stamp `AGENT_RELAY_ORCHESTRATOR_HARNESS` into the env handed to the broker — the established pattern for passing values to the broker (cf. `AGENT_RELAY_STATE_DIR`). The detached broker checks env **before** walking the process tree (`detect_orchestrator_harness` in `crates/broker/src/telemetry.rs`), so it now forwards the real harness, and the agents it spawns inherit it too.

- New `ensureOrchestratorHarnessEnv(env, detect?)` helper, called at the top of `runUpCommand` so both the background (detached re-invoke) and foreground paths inherit it.
- An explicit value already in env wins — including the re-invoked detached `up` inheriting ours, so detection runs once and isn't clobbered.
- `detect` is injectable → unit-tested without depending on the real process tree.

## Tests

- `broker-lifecycle.test.ts`: +4 cases (stamps detected harness; ignores `unknown`; respects explicit value; threads env to the real detector via `RELAYCAST_HARNESS`). Suite green (17 tests).
- CLI `tsc --noEmit` clean; `eslint` clean (the 6 pre-existing warnings are unrelated, in `refreshDashboardAssetsIfStale`).

## Rollout

Ships in the next `agent-relay` release. After adoption, `sdk-rust` rows in 296966 should flip from `unknown` to the real harness. This is the second half of the harness fix (forwarding plumbing #1069 + this detection capture).

### DIFF
--- a/packages/cli/src/cli/lib/broker-lifecycle.test.ts
+++ b/packages/cli/src/cli/lib/broker-lifecycle.test.ts
@@ -1,6 +1,12 @@
 import { describe, expect, it } from 'vitest';
 
-import { classifyBrokerStartError, classifyBrokerStartStage, describeError } from './broker-lifecycle.js';
+import {
+  classifyBrokerStartError,
+  classifyBrokerStartStage,
+  describeError,
+  ensureOrchestratorHarnessEnv,
+} from './broker-lifecycle.js';
+import { ORCHESTRATOR_HARNESS_ENV } from '../telemetry/orchestrator-harness.js';
 
 describe('describeError', () => {
   it('returns plain message for a bare Error', () => {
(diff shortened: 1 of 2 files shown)


willwashburn:
Closing as redundant. `bootstrap.ts` (`runCli` → `propagateTelemetryContextToChildren`) already detects the orchestrator harness in the non-detached outer CLI and sets `process.env.AGENT_RELAY_ORCHESTRATOR_HARNESS` before any command runs. Since `deps.env === process.env` (core.ts:277) and the broker is spawned with `env: deps.env`, the broker already inherits it — the guard in `ensureOrchestratorHarnessEnv` short-circuits, so this is a no-op. The `unknown` traffic is genuinely harness-less sessions, not a detection gap. Pursuing per-worker harness attribution instead.

PR #1078, opened 2026-06-10
feat(broker): per-worker harness attribution for relaycast telemetry

willwashburn:
## Problem

`relaycast_server_*` telemetry attributes **every** spawned agent to the session **orchestrator** harness — the human's `claude-code`/`codex` that ran `agent-relay up`. That's because `AGENT_RELAY_ORCHESTRATOR_HARNESS` is set once by the outermost CLI (`bootstrap.ts`) and inherited by the broker and all workers.

So a broker orchestrating a **heterogeneous swarm** (a `codex` worker spawned under a `claude-code` orchestrator, a `gemini` worker, …) collapses every worker's relaycast telemetry to the single orchestrator value. We want **per-worker** attribution: each agent's traffic tagged with the harness it actually runs.

## Approach

The broker knows each worker's CLI at spawn time (`spec.cli` / `params.cli`), so derive the worker's harness there and stamp `AGENT_RELAY_HARNESS` — the **highest-priority** key the JS SDK reads (`AGENT_RELAY_HARNESS` → `AGENT_RELAY_ORCHESTRATOR_HARNESS` → …), so it overrides the inherited orchestrator value.

Workers talk to relaycast via their **own** `@relaycast/sdk` (the MCP server constructs `RelayCast` directly against the gateway), and that SDK resolves harness from env — so setting the worker's env is the correct lever.

Covers all three worker entry modes:

| Mode | Path | Change |
|---|---|---|
| `wrap` | `spawn_env_vars` (via `run_wrap`) | pass `infer_harness_from_command(params.cli)` instead of the broker's orchestrator harness |
| `pty` / `app-server` | `WorkerRegistry::spawn` | stamp `AGENT_RELAY_HARNESS` from `infer_harness_from_command(spec.cli)` into the worker env (unless `harness_config` already set it) |

- Reuses `telemetry::infer_harness_from_command` (already used for orchestrator process-tree detection) — now `pub(crate)` — for cli→harness mapping.
- An **unrecognized** CLI sets nothing, so the worker gracefully falls back to the inherited orchestrator value rather than a wrong attribution.
- The broker's **own** relaycast traffic (session WS, agent registration) keeps the orchestrator harness — the correct scope for broker-level events.

## Result

- Per-worker traffic → the worker's harness (`codex`, `claude-code`, `gemini`, …).
- Broker's own traffic → the session orchestrator.

## Tests

- cli→harness mapping locked for the bare names the broker passes (`claude`→`claude-code`, `codex`, `cursor`, `gemini`, and unknown→`None`).
- 709 broker lib tests pass; `clippy` + `cargo fmt --check` clean.

## Context

Follow-up to #1069 (broker forwards harness) and the closed #1075 (which was redundant — `bootstrap.ts` already propagates the orchestrator harness). This PR is the deliberate move from *orchestrator-only* to *per-worker* attribution.


willwashburn:
Superseded by #1090, which folds the per-worker attribution into the full origin_actor model — the broker now sets AGENT_RELAY_ORIGIN_ACTOR=agent-relay-cli/agent/<harness> (the full path), consumed by the JS SDK 3.1.1.

PR #1095, opened 2026-06-11
feat(broker): resume missed events on reconnect

willwashburn:
## Problem

The Rust broker holds a WebSocket to the hosted Relaycast engine and was the last client that reconnected blind: the reconnect path had no cursor and no backfill, so every disconnect window silently lost messages that should have been injected into agent PTYs.

## Which socket the broker holds (investigation result)

The broker authenticates the WS with the **workspace key** (`rk_live_...`, see `WsClientOptions::new(self.workspace_http.api_key...)` plus `ensure_workspace_stream_enabled`). Server side, `rk_live_` tokens route to the **workspace-stream firehose** (`attachWorkspaceSocket` in the engine node entrypoint; `WorkspaceStreamDO` in cloud) — *not* a per-agent socket.

The `resync` / `resync_ack` protocol (relaycast PR #176) lives only on per-agent sockets: `agent_seq` stamping, the 500-event ring, and the `resync` frame handler exist in `attachAgentSocket` / `onAgentMessage` (engine) and the AgentDO (cloud). `WorkspaceStreamDO` and `attachWorkspaceSocket` have no resync handling and workspace-stream events carry no `agent_seq`. The Rust SDK (`relaycast 3.0.0`) has no resync support either.

## Path taken: client-side cursor + backfill

Since the broker's socket cannot resync server-side, this PR implements the cursor + backfill alternative:

- **Cursor**: `BackfillCursor` tracks the newest server-side `created_at` observed on message-bearing frames (tolerant extraction across top-level / payload-wrapped shapes; absent or malformed timestamps never break it or regress it). Seeded to connect time on first connect so a backfill before any traffic replays nothing.
- **Backfill on reconnect** (both the outer reconnect loop and SDK-internal re-open via `WsLifecycleEvent::Open`): fetches subscribed-channel history and all-workspace DM history via the existing REST APIs (`get_channel_messages`, `get_all_dms`), converts messages strictly newer than the cursor into synthetic `message.created` events, and sends them down the **normal delivery path** (bridge → routing → pending delivery). Replayed events carry the original message ids, so the runtime `DedupCache` (keyed `{workspace_id}:{event_id}`, 300s TTL) suppresses anything already delivered. No changes to PTY injection or routing.
- **Observability**: structured `tracing` log plus a `relay_resync` `BrokerEvent` (`{workspace_id, replayed, gap_detected}`) emitted through the runtime, mirrored in `packages/harness-driver/src/protocol.ts`.

## Known limitations / follow-up

- Thread replies created during a disconnect are only recovered if they appear in channel history; a dedicated replies backfill was left out to keep the diff focused.
- The complete fix for long gaps under heavy multi-channel traffic is server-side resync on the workspace stream (seq stamping + ring + DB replay in `WorkspaceStreamDO` / `attachWorkspaceSocket`, plus `resync` support in the Rust SDK `WsClient`). That is a relaycast/cloud follow-up, intentionally not built here.
- Backfill fetches up to 50 messages per channel / DM conversation per reconnect; gaps larger than that window are truncated (oldest messages dropped).

## Tests

- `backfill_cursor_tracks_newest_timestamp_tolerantly` — seq/cursor tracking from frames, malformed-field tolerance, no regression on older frames.
- `collect_backfill_events_filters_by_cursor_and_orders_oldest_first` — httpmock-backed filtering + ordering across channel and DM history.
- `replay_missed_messages_replays_gap_and_emits_resync_frame` — backfill trigger emits replayed events then the `broker.resync` frame, and advances the cursor.
- `replay_missed_messages_is_noop_without_cursor`
- `replayed_backfill_event_dedupes_against_live_delivery` — synthetic replay maps through the bridge to the same dedup key as the live event and is suppressed.

Verification: `cargo fmt --check` clean; `cargo clippy -p agent-relay-broker --all-targets` clean except one pre-existing `get_first` warning in `snippets.rs` (present on main); `cargo test -p agent-relay-broker` — 727 lib + 12 integration passed, 0 failed (4 pre-existing ignores + 2 pre-existing ignored ws fixtures). `protocol.ts` typechecks standalone; full package typecheck not run (workspace deps not installed in the worktree).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `agent-relay-broker` recovers messages missed during Relaycast WebSocket disconnects: on reconnect it backfills channel and DM history past the last seen message and routes the gap through normal delivery with dedup, emitting a `relay_resync` broker event (`{replayed, gap_detected}`) mirrored in `@agent-relay/harness-driver` protocol types.
 - `@agent-relay/sdk` wires the durable delivery surface to the Relaycast backend: `inbox.list`/`inbox.subscribe` replay and stream the per-recipient delivery ledger, `inbox.ack/fail/defer` and `deliveries.ack/fail/defer` apply real server transitions, capabilities report `serverDeliveryState: true` for agent-scoped clients, and `DeliveryRunner` now works against the hosted backend.
 - `agent-relay-broker` and `@agent-relay/harness-driver` accept explicit workspace keys and broker instance names, so local and cloud brokers can join the same Relay workspace with stable, addressable names.
 - `@agent-relay/harnesses` adds a `grok` PTY harness for the Grok CLI, including Relaycast MCP support for spawned agents.
--- a/crates/broker/src/protocol.rs
+++ b/crates/broker/src/protocol.rs
@@ -522,6 +522,11 @@ pub enum BrokerEvent {
         name: WorkerName,
         channels: Vec<ChannelName>,
     },
+    RelayResync {
+        workspace_id: WorkspaceId,
+        replayed: usize,
+        gap_detected: bool,
+    },
 }
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
--- a/crates/broker/src/relaycast/ws.rs
+++ b/crates/broker/src/relaycast/ws.rs
@@ -1,6 +1,7 @@
 use std::{collections::HashMap, sync::Arc, time::Duration};
 
 use anyhow::{Context, Result};
+use chrono::{DateTime, Utc};
 use parking_lot::Mutex;
 use relaycast::{
     agent::DmOptions, format_registration_error,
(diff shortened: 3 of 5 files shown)


willwashburn:
Closing as **superseded** — the problem this PR fixes no longer exists on `main`.

This PR added a client-side cursor + history backfill to the broker's **workspace-stream WebSocket** (`/v1/ws`, `rk_live_` workspace key), which reconnected blind with no server-side resync. Since this was written, **#1201 (node-only delivery, relaycast v5.0.1)** re-architected broker delivery:

- `RelaycastWsClient` and the entire workspace-stream socket this PR patches were removed. `relaycast/workspace.rs` now notes the legacy `/v1/ws` socket is "observer-only … no longer opened."
- Inbound delivery + PTY injection moved to `/v1/node/ws` (`node_control.rs`), which already has the server-side resync the workspace stream lacked: per-agent monotonic `seq`, a `FleetDeliveryBook` cursor with `Deliver/Duplicate/Stale/Gap` gating, ack-withheld-on-injection-failure (at-least-once redelivery), and a `resume_cursor` on reconnect.

That's the "server-side resync" this PR's description itself deferred to a follow-up — main shipped it, in a better place. The code region this PR touches (`ws.rs`, `runtime/relaycast_events.rs`) has been rewritten/removed, so it can't be rebased without re-implementing a client-side backfill on a socket that's no longer opened. No code change is warranted.

---

Issue #1056, opened 2026-06-07
Design RFC: Fleet Delivery — Agents, Nodes, and Reliable Messaging


willwashburn:
> **Design RFC — for discussion, not yet scheduled work.**
> Captures a design session on running agents across a fleet of machines and making messaging reliable. Source of truth lives at `specs/fleet-delivery.md` (branch `feature/messaging-protocol-evals`); this issue mirrors it so others can read and comment. Please leave feedback inline.

---

# Fleet Delivery — Agents, Nodes, and Reliable Messaging

**Status**: Draft
**Date**: 2026-06-06
**Author**: Design session (Will + Claude)

---

## 1. Vision

Run agents across many machines. Each machine (a **node**) can run a specific set of compute — some can spawn Claude agents, some Codex, some both. Relaycast is the control plane: agents are equal peers in a flat messaging fabric, nodes advertise what they can spawn, and Relaycast routes messages to agents wherever they live and places new agents onto nodes that can run them.

The goal is the simplest deployable unit that maximizes the environments a node can run in, with messaging that survives flaky networks and agent restarts.

## 2. The frame: two planes

Everything below lives on one of two planes. Keeping them separate is what keeps the model simple.

- **Messaging fabric — flat, all equal `agents`.** Every agent is a peer: a stable identity that sends, receives, and may expose actions. No agent is above or routed "through" another.
- **Compute layer — `nodes`.** A node is a machine where some agents run. It has a **broker** runtime and a set of **capabilities** (what it can spawn). Agents sharing a node is a deployment fact, not a relationship in the fabric.

There is **no "participant" umbrella and no agent subtype.** "Orchestrate vs communicate" is not a type distinction — it's just whether an agent is colocated on a node-with-broker or self-connected (see *location*, §5).

## 3. Core concepts

| Concept | What it is | Plane |
|---|---|---|
| **Agent** | a peer in the fabric: identity, send/receive, exposed actions | messaging |
| **Node** | a named machine that runs agents and advertises capabilities | compute |
| **Broker** | a node's runtime/delivery engine — **infra, not a peer** (not in the agent roster) | compute |
| **Location** | where Relaycast routes an agent's **inbound** | routing detail |
| **Capability** | what a node can spawn (e.g. `spawn:codex`) | compute |
| **Action** | something an agent exposes/invokes in the fabric | messaging |

[...]

PR #1107, opened 2026-06-12
feat(broker): fleet node control plane (Phase 3)

willwashburn:
Summary
- Adds the Rust broker fleet node control client for /v1/node/ws with node registration, heartbeat/load reporting, inventory sync, reconnect/backoff, delivery acking, handler dispatch, and agent token minting.
- Adds broker local sidecar WS dispatch for register_node, register_handlers, handler_result, invoke_handler, and handler-initiated spawn_agent using the existing AgentSpec spawn shape.
- Wires fleet delivery into existing wait/steer PTY injection while guarding fleet-mode PTY agents from the legacy workspace inbound path.

Refs AgentWorkforce/relay#1056.

Stacking / merge order
- Stacked on feat/fleet-wire-serde and feat/fleet-local-protocol.
- Merge order: feat/fleet-wire-serde -> feat/fleet-local-protocol -> feat/fleet-control-plane.

Validation
- cargo fmt --check
- cargo clippy -p agent-relay-broker -- -D warnings
- cargo test -p agent-relay-broker
- npm --prefix packages/harness-driver run build
- npx vitest run packages/harness-driver/src/protocol.test.ts
- npm run build


willwashburn (commented):
Round 1 review verdict: NO-GO. I attempted to submit this with `--request-changes`, but GitHub rejected it because this account is considered the PR author (`Review Can not request changes on your own pull request`). Treat this as a request-changes review.

Validation I ran in a fresh `/tmp/review-Rev3x1` worktree:
- `cargo fmt --check` passed.
- `cargo clippy -p agent-relay-broker -- -D warnings` passed.
- `cargo test -p agent-relay-broker` passed: 745 unit tests, 12 continuity tests, 1 fleet fixture test, 4 ignored.
- `npm ci` completed, with existing audit warnings.
- `npx vitest run packages/harness-driver/src/protocol.test.ts` passed.
- `npm run build` passed.
- `npm --prefix packages/harness-driver run build` passes after `npm run build` has generated `@agent-relay/sdk` dist. It fails in a clean worktree before that build because `@agent-relay/sdk/actions` has no generated dist yet, so the validation order matters.

Findings:

1. [blocker] `node.register.capabilities` does not match the Relaycast fleet wire contract. This PR serializes each capability as an object (`NodeCapability { name, kind, metadata }`) and the Rust fixture contains object entries at `crates/broker/src/fleet_wire.rs:55` / `crates/broker/tests/fixtures/fleet-wire/node.register.json:6`. The current Relaycast PR #192 contract still validates `capabilities: z.array(z.string())` and its `node.register.json` fixture is `['spawn:codex', 'spawn:claude', 'run:test']`. A broker built from this PR will send object capabilities and be rejected by the engine before the node can register. Align the shared wire type and fixtures across both repos before merging: either keep string capabilities here, or update Relaycast's zod schema/types/fixtures and tests to the object shape in the same stack.

2. [blocker] A failed fleet delivery injection can be acked as delivered on retry. `handle_fleet_deliver` calls `self.fleet_delivery_book.observe(&deliver)` before writing to the worker (`crates/broker/src/runtime/fleet.rs:245`), and `observe` immediately inserts the `msg_id` and advances `up_to_seq` for the next contiguous seq (`crates/broker/src/node_control.rs:117`, especially `:138-:139`). If `workers.deliver` then returns an error, the code withholds the ack and returns (`crates/broker/src/runtime/fleet.rs:250-260`), but the cursor has already advanced. When Relaycast retries the same `msg_id`, `observe` returns `Duplicate`, and `handle_fleet_deliver` sends `delivery.ack` for that advanced `up_to_seq` without ever injecting the message (`crates/broker/src/runtime/fleet.rs:264-274`). Move the cursor/msg_id mutation behind successful injection, or split the delivery book into a pure decision plus explicit commit step. Add a test where the first injection fails and the retry is actually injected before the cumulative ack advances.

3. [major] Fleet inventory never removes agents or publishes the removal path. Agents are inserted into `fleet_inventory` after spawn (`crates/broker/src/runtime/fleet.rs:400`) and `publish_fleet_inventory` blindly sends that map on reconnect (`crates/broker/src/runtime/fleet.rs:508-514`), but release and exit cleanup only remove `delivery_states`, result tokens, and persisted state (`crates/broker/src/runtime/api.rs:425-450`, `crates/broker/src/runtime/maintenance.rs:148-195`, `crates/broker/src/runtime/maintenance.rs:202-248`). There is no `fleet_inventory.remove`, no `inventory.sync` update, and no `agent.deregister` on release/death, despite the wire variant existing. A node reconnect after an agent release/crash will re-announce dead agents as live, defeating the Phase 3 inventory reconciliation requirement. Remove entries and publish inventory/load on every release, permanent death, unsupervised exit, and successful restart transition; send `agent.deregister` where the engine contract expects it.

4. [major] Spawn attribution drops `invocation_id` under concurrent handler invocations. `handle_fleet_spawn_agent` guesses the spawn's source with `self.fleet_handlers.single_in_flight_invocation()` (`crates/broker/src/runtime/fleet.rs:385`), and that helper returns `None` unless exactly one action is in flight (`crates/broker/src/node_control.rs:195-199`). If two node actions are dispatched concurrently and both sidecar handlers call `spawn_agent`, both `agent.register` and subsequent `inventory.sync` records can omit `invocation_id`, so Relaycast cannot resolve/reconcile the open spawn invocations. Make the local `spawn_agent` protocol carry the current `invocation_id` explicitly (or otherwise bind the sidecar handler context to the spawn request) and reject ambiguous spawns instead of silently registering them without attribution.

5. [major] Generated resumable session refs are registered too early and can be lost. `handle_fleet_spawn_agent` computes `session_ref` from the incoming `AgentSpec` before calling `spawn_from_agent_spec` (`crates/broker/src/runtime/fleet.rs:386-394`), but `WorkerRegistry::spawn` can generate or discover the actual session id later for Claude/Codex PTY agents (`crates/broker/src/worker.rs:295-328` and again for the non-harness PTY path at `:499-:530`). Those generated session ids never make it into the earlier `agent.register` request or the inventory entry created from the pre-spawn `session_ref` (`crates/broker/src/runtime/fleet.rs:400-407`). Pre-resolve the effective session id before `agent.register` and before starting the harness, or require the sidecar to supply the session ref explicitly, so resumable spawned agents are registered and reconciled with the real `session_ref`.

6. [major] The handler send-failure path leaves the invocation marked in-flight after reporting `handler_unavailable`. `handle_invoke` inserts the invocation into `in_flight` before the broker sends `invoke_handler` to the sidecar (`crates/broker/src/node_control.rs:225-236`). If the outbound send fails, `handle_fleet_action_invoke` sends an `action.result {error: handler_unavailable}` directly but only calls `disconnect_sidecar`; it does not remove or complete that specific in-flight entry (`crates/broker/src/runtime/fleet.rs:356-367`). Until the later disconnect event drains the map, duplicate `action.invoke` messages for the same id hit `AlreadyInFlight` (`crates/broker/src/node_control.rs:216-217`) and no result is resent; when the disconnect drain arrives, the same unavailable result can be emitted again. Add a `fail_in_flight(invocation_id, handler_unavailable)` path that removes the entry and records/replays the terminal result before sending it, and cover duplicate invoke after a failed forward.

willwashburn (commented):
Round 2 adversarial review: NO-GO. I attempted to submit this with --request-changes, but GitHub rejected it because this identity is treated as the PR author. The requested-change verdict still stands.

Findings:
1. [major] packages/harness-driver/src/protocol.ts:161 / crates/broker/src/runtime/fleet.rs:107 - spawn_agent attribution is still optional. The TS contract exposes invocation_id?: string and the broker accepts Option<String>, then sends agent.register with invocation_id=None. Phase 3 and round-1 require spawn attribution strictly via spawn_agent.invocation_id; an omitted value creates via-node agents that the engine cannot bind to the action invocation. Change the local protocol and Rust serde to require invocation_id for fleet spawn_agent, reject missing or blank values before spawning, and add negative tests on both TS and Rust paths.
2. [major] crates/broker/src/runtime/fleet.rs:391 - agent.register is sent before the local spawn succeeds, and the error path from spawn_from_agent_spec at line 396 returns without agent.deregister or inventory cleanup. A bad cwd, bad CLI, or duplicate name can mint an agent token and leave Relaycast believing a via-node agent exists even though no worker was started. Change this to either defer registration until after a successful spawn via a token reservation flow, or send agent.deregister plus an immediate inventory.sync on every spawn failure before returning the error.
3. [major] crates/broker/src/runtime/fleet.rs:516 - sidecar supervision bypasses the existing RestartPolicy and restarts forever. On child exit it sets restart_at to Instant::now() and a successful restart clears all state; repeated immediate exits will loop without max_restarts, max_consecutive_failures, or backoff. Wire supervision through the existing supervisor policy or add equivalent bounded restart accounting for sidecars, and emit handlers_live=false/permanent failure when exhausted.
4. [major] crates/broker/src/node_control.rs:101 - FleetDeliveryBook keeps an unbounded seen_msg_ids set per agent and there is no cleanup when agents are released or reaped. Release only prunes inventory at crates/broker/src/runtime/api.rs:456; death only prunes inventory at crates/broker/src/runtime/maintenance.rs:203 and crates/broker/src/runtime/maintenance.rs:262. Busy agents accumulate every msg_id forever, and released agent names can retain stale cursors that may treat a later same-name agent seq=1 delivery as stale. Bound or TTL the dedup set and remove/reset delivery cursors on agent deregister/release/permanent death/non-supervised exit, or scope cursors by stable agent_id/location instead of name.
5. [minor] crates/broker/src/node_control.rs:436 - reconnect backoff has a cap but no jitter and never resets after a successful connection. A Relaycast deploy or network flap will keep all nodes retrying in lockstep, and a node that has accumulated 30s backoff keeps that delay for later independent disconnects. Add randomized jitter and reset to INITIAL_RECONNECT_DELAY after a connection survives registration/heartbeat, with a test that samples non-identical delays.
6. [minor] crates/broker/src/node_control.rs:181 - HandlerDispatchState.completed is an unbounded replay cache keyed by invocation_id. A long-running node handling many action invocations will retain every completed result even though duplicates only need a short idempotency window. Add a bounded TTL/LRU or tie eviction to an engine ack/timeout.

Validation:
- PASS: cargo fmt --check
- PASS: cargo clippy -p agent-relay-broker -- -D warnings
- FAIL: cargo test -p agent-relay-broker failed in the full suite at relaycast::auth::tests::workspace_name_conflict_retries_with_fresh_suffix with RELAY_WORKSPACES_JSON must be valid JSON. The same test passed when run alone with Relay env vars unset, so this looks like an existing process-env race rather than a fleet regression, but the required full command did not pass here.
- PASS: npm ci
- PASS after SDK build: npm run build:sdk && npm --prefix packages/harness-driver run build. The package-only harness-driver build failed first in the fresh worktree because @agent-relay/sdk/actions resolves through SDK dist before the SDK is built.
- PASS: npx vitest run packages/harness-driver/src/protocol.test.ts
- PASS: npm run build

willwashburn (commented):
Round 3 adversarial review: NO-GO. I attempted to submit this with `--request-changes`, but GitHub rejected it because this identity is treated as the PR author (`Review Can not request changes on your own pull request`). Treat this as a request-changes review.

Validation I ran in a fresh `/tmp/review-Rev3x3` worktree at `642ad769f3ec57a6d931d51243365761e17e441f`:
- PASS: `cargo fmt --check`
- PASS: `cargo clippy -p agent-relay-broker -- -D warnings`
- PASS: `cargo test -p agent-relay-broker` (756 unit tests, 12 continuity tests, 1 fleet fixture test, 4 ignored, doc tests ok)
- PASS: `npm ci` (completed; existing audit warnings reported)
- PASS after `npm ci`: `npx vitest run packages/harness-driver/src/protocol.test.ts`
- PASS: `npm run build`
- PASS: `npm --prefix packages/harness-driver run build`

Findings:

1. [major] `crates/broker/src/runtime/fleet.rs:543` / `crates/broker/src/node_control.rs:631` - `agent.register` timeouts can leave Relaycast with a ghost via-node agent. `register_fleet_agent_token` times out the oneshot after 30s and returns `agent_register_timeout`, so `handle_fleet_spawn_agent` abandons the local spawn before any worker exists. However, the node-control task keeps the request in `pending_agent_registrations` after sending the wire `agent.register`; if Relaycast replies successfully after the runtime timeout, `complete_agent_registration` removes the pending entry and tries to send the minted `{agent_id, token}` into a dropped receiver, ignoring the failure (`crates/broker/src/node_control.rs:737`). At that point Relaycast has minted and registered a via-node agent, but the broker never spawns it and cannot send the spawn-failure `agent.deregister` cleanup because the runtime no longer has the late `agent_id`. Move the timeout/cancellation handling into node_control so a late success after caller cancellation is immediately deregistered, or avoid abandoning the registration until there is a terminal wire reply/disconnect. Add a test where an `agent.register` reply arrives after the caller-side timeout and assert no registered agent remains without a local spawn.

2. [major] `crates/broker/src/runtime/fleet.rs:480` / `crates/broker/src/worker.rs:295` - generated resumable session refs still miss the `agent.register` frame. Phase 3 requires `agent.register` to include `invocation_id + session_ref` for resumable harnesses, but `handle_fleet_spawn_agent` snapshots `initial_session_ref` from the incoming `AgentSpec` before spawn and sends it in `AgentRegister` at `crates/broker/src/runtime/fleet.rs:531`. For common PTY spawns without an explicit session id, `WorkerRegistry::spawn` generates or discovers the actual Claude/Codex session later (`crates/broker/src/worker.rs:295-328`), so the wire `agent.register` goes out with no `session_ref` and no `resumable`, and only a later `inventory.sync` carries the discovered session. Pre-resolve the effective session id before sending `agent.register` (or require the sidecar spawn request to supply it) so the first registration frame matches the fleet contract; add a regression test for a Claude/Codex fleet spawn with no caller-supplied session id.

willwashburn (commented):
**Round 4 (FINAL VERIFY) — GO.** Adversarial pass over tip `aae4d28e`. Built and ran everything locally: `cargo test -p agent-relay-broker` → **760 passed / 0 failed** + `fleet_wire_fixtures` pass; `tsc -p packages/harness-driver/tsconfig.json` → clean (so the `@ts-expect-error` metadata guards still bite); `vitest protocol.test.ts` → 5 passed.

### Round-3 fixes — all verified
1. **Register-timeout compensation** — `node_control.rs:531` keeps `pending_agent_registrations` keyed by request-id; `expire_agent_registrations` (`:788`) prunes at the 300s TTL on each heartbeat tick (`:654`); a late *success* whose caller already gave up hits the `Err(Ok(token))` arm (`:752-773`) and emits a correlated `agent.deregister{id, agent_id, name}`. The test `node_control_agent_register_timeout_late_success_deregisters` (`:1399`) genuinely simulates timeout-then-late-reply — it `drop(reply_rx)`, *then* releases the late `reply`, and asserts the compensating `agent.deregister` carries `agent_id="agt-late"` + `name="agent-a"`. Not vacuous. (One residual gap — minor #1 below.)
2. **Session discovery → immediate `inventory.sync`** — `worker_events.rs:384-392` calls `refresh_fleet_inventory_session_ref` on `worker_ready` (not only on reconnect); `spec.session_id` is already populated at spawn by `worker.rs` (`:297/:426/:501`), so the discovered ref is real. `refresh_fleet_inventory_session_ref_publishes_immediate_sync` asserts the immediate publish, the trim, and the no-op-on-unchanged short-circuit. Spawn path also publishes inventory immediately (`fleet.rs:517`).
3. **Object-record capability metadata** (relay#1105 `abfd5ff0`) — Rust `FleetCapability.metadata: Option<BTreeMap<String,Value>>` (`fleet_wire.rs:70`); TS `Record<string, JsonValue>` with compile-time `@ts-expect-error` rejection of scalar/array metadata (`protocol.ts:93-102`), confirmed by clean `tsc`. Fixtures + `register_node` test exercise object metadata.

### Rounds 1–2 spot-checks — intact
- **Injection-gated ack**: `fleet.rs:346-357` withholds the `delivery.ack` and `return`s when `workers.deliver` errors; only commits + acks on `Ok`.
- **RestartPolicy bounds**: `fleet.rs:30-55` enforces both `max_restarts` and `max_consecutive_failures`; `sidecar_restart_state_obeys_restart_policy_bounds` green.
- **Cursor pruning**: both worker-exit arms (`maintenance.rs:204` supervised-dead, `:264` plain-exit) call `prune_fleet_agent_state` → `remove_agent` + inventory diff. Spawn-failure uses explicit `agent.deregister` (`fleet.rs:763`) because the agent was never inserted into inventory — correct asymmetry.
- **deregister_node clean-shutdown vs crash**: `deregister_node` clears supervision + restart state and sends wire `node.deregister` → no restart (`fleet.rs:87-103`); a crash/disconnect keeps supervision and schedules a bounded restart (`fleet.rs:75-85`). Legacy workspace-stream inbound is short-circuited in fleet mode (`relaycast_events.rs:626-634`) so PTY agents aren't double-delivered.

### Minor findings (non-blocking — fine to land, fix in a follow-up)
1. **`node_control.rs:730-737` — late agent.register reply after TTL leaves an orphan with no compensating deregister.** Once `expire_agent_registrations` drops the pending entry at the 300s TTL, a subsequent `reply{ok:true}` for that id falls into the "did not match a pending registration" branch and is only logged. But the engine has already minted the agent + bound its location, and because the spawn caller already failed at the 30s `FLEET_AGENT_REGISTER_TIMEOUT` the agent was never inserted into `fleet_inventory` — so it appears in no `inventory.sync` and can linger/hold the name until an engine-side liveness sweep. The reply still carries `data.agent_id`. Fix: in the `None` arm of `complete_agent_registration`, try `reply.validate_agent_register_data()` and, on success, send a compensating `agent.deregister{agent_id, name}` (same as the dropped-receiver path). Low likelihood (needs an engine reply >300s on a still-open socket), hence minor.
2. **`node_control.rs:110-165` — `AgentDeliveryCursor.seen_msg_ids` grows unbounded.** One entry is inserted per committed delivery and is only cleared by `remove_agent`. In `observe`, any redelivered *committed* message has `seq <= up_to_seq` and would already be caught by the `Stale` check, so the msg_id set never changes ack behavior (it only re-labels `Stale`→`Duplicate`) yet leaks memory for long-lived high-volume fleet PTY agents. Recommend bounding it (retain only the last committed msg_id) or dropping it in favor of the seq cursor.

Wire format (snake_case fields, `type` discriminant, `v:1` envelope, exhaustive `NodeToServer`/`ServerToNode` coverage, `deny_unknown_fields`, explicit-null rejection, finite/nonnegative load, `ok` literal enforcement) is consistent and well-tested. No blocker or major. **GO.**

PR #1108, opened 2026-06-12
feat(fleet): @agent-relay/fleet SDK, fleet serve CLI, MCP query_nodes/spawn + §5 resource cleanup (Phase 4)

willwashburn:
## Summary

- Add `@agent-relay/fleet` with `defineNode`, `action`, `spawn`, `onMessage`, manifest helpers, trigger sync inputs, and default local node support.
- Add `agent-relay fleet serve|nodes|status`, TS node loading through `jiti`, fleet sidecar broker registration, handler dispatch, trigger sync, status reporting, and `examples/relay-node.ts`.
- Extend the SDK with `nodes` and `triggers` surfaces used by the CLI and MCP tools.
- Update MCP by adding `query_nodes` and fleet `spawn`. The §5 MCP resource surface (`RealtimeResourceBridge`, resource registrations, and `enableInboxPiggyback`) is **retained** — the earlier plan to remove it was superseded by the accepted control-plane stack, which keeps this surface. (Round-4 review finding 3.)
- Preserve `local up` compatibility by starting an implicit default fleet node for `claude`, `codex`, `gemini`, plus additional `teams.json` CLIs.

## Stack

Refs AgentWorkforce/relay#1056.

Stacked after AgentWorkforce/relay#1105 (`feat/fleet-local-protocol`) and AgentWorkforce/relay#1107 (`feat/fleet-control-plane`). This branch is based on `feat/fleet-control-plane`, which already contains the local protocol merge in the stack.

## Validation

- `npm run typecheck`
- `npm run build` (incl. release Rust broker)
- `npm test` (64 files passed, 824 tests passed; 1 file/5 tests skipped)
- `cargo test -p agent-relay-broker` (760 unit + 12 continuity + fleet wire fixtures, 0 failed)
- Sidecar↔broker WS smoke against the rebuilt merged broker: `hello`/`register_node`/`register_handlers`/`deregister_node` + spawn `invocation_id` attribution all pass.


willwashburn (commented):
1. major packages/fleet/src/index.ts:155-175 - defineNode() normalizes capability names into an object map but never checks for collisions after trimming. A definition like foo plus foo  silently overwrites one handler, so the node surface can change without any error. Reject duplicate capability names before building the map.

2. blocker packages/cli/src/cli/lib/fleet-sidecar.ts:231-236 - the SIGTERM/abort path only closes the websocket (optionally after sending an empty register_handlers frame) and never emits the defined node.deregister frame. That leaves deregistration dependent on transport loss instead of an explicit teardown, so node/handler state can linger after a controlled shutdown. Send node.deregister before close and have the broker treat it as the authoritative shutdown signal.

willwashburn (commented):
Round 2 review: NO-GO. I found blocking/major issues. GitHub would not allow this identity to submit a request-changes review (`Review Can not request changes on your own pull request`), so this is posted as a review comment instead.

1. [blocker] Clean shutdown deregistration is not wired to the real broker. The sidecar sends `node.deregister` before close (`packages/cli/src/cli/lib/fleet-sidecar.ts:231`), but the broker deserializes every local sidecar frame into `SdkToBroker` (`crates/broker/src/runtime/fleet.rs:47`) and `SdkToBroker` has no `node.deregister` variant in the local protocol enum (`crates/broker/src/protocol.rs:408`). The error is swallowed by `.catch(() => undefined)` at `packages/cli/src/cli/lib/fleet-sidecar.ts:233`, so the round-1 test only proves a stub saw the frame; the real broker rejects it and the control loop shuts down without sending fleet-wire `node.deregister` (`crates/broker/src/node_control.rs:551`). Add a Rust local-protocol variant and runtime arm that sends `BrokerToRelaycast::NodeDeregister { v: FLEET_WIRE_VERSION }` before socket close/control shutdown, and add an integration/unit test against the broker parser/runtime, not just a stub.

2. [major] `spawn_agent` lost `invocation_id` from the stacked local protocol. Commit `0c2c2899` had `payload.invocation_id?: string`; this branch now only allows `{ agent, initial_task?, skip_relay_prompt? }` (`packages/harness-driver/src/protocol.ts:151`), and the sidecar omits the current invoke id when calling `spawn_agent` (`packages/cli/src/cli/lib/fleet-sidecar.ts:332`). The broker then guesses with `single_in_flight_invocation()` (`crates/broker/src/runtime/fleet.rs:385`), which returns `None` as soon as two spawn actions are in flight, so `agent.register`/inventory lose the action invocation binding and resumable placement cannot reconcile reliably. Restore optional `invocation_id` in TS/Rust protocol, thread the current `invoke_handler.payload.invocation_id` through `FleetActionContext.spawnAgent`, and add a concurrent spawn-action test.

3. [major] Trigger sync is not idempotent. `syncTriggers` runs after every registration (`packages/cli/src/cli/lib/fleet-sidecar.ts:264`) and always calls `relay.triggers.create` for each descriptor (`packages/cli/src/cli/lib/fleet-sidecar.ts:347`). Since `serveFleetSidecar` reconnects indefinitely (`packages/cli/src/cli/lib/fleet-sidecar.ts:121`), restarting or reconnecting a node duplicates the same message triggers, causing one message to invoke the same action multiple times. Use the trigger CRUD surface to list existing triggers and update/create by stable key, or delete stale ones, and test serving the same definition twice does not increase trigger count.

4. [major] MCP resource cleanup is unguarded despite the upstream audit saying consumers still exist. The relaycast#193 Section 11 audit says active MCP resource/subscription consumers remain and actual deletes are gated. This PR codifies no resources and no subscription handlers (`packages/cli/src/cli/agent-relay-mcp.startup.test.ts:297`, `packages/cli/src/cli/agent-relay-mcp.startup.test.ts:553`), with no feature flag or compatibility path. Restore the legacy MCP resource/subscription surface until the gate is satisfied, or add an explicit opt-in flag/migration story and update the PR body to justify the consumer break.

Validation run locally in `/tmp/review-Rev4x2`: `npm run typecheck` passed; `npm run build` passed; `npm test` passed (64 files, 819 tests, 5 skipped); `cargo test` passed; targeted fleet/local/MCP Vitest subset passed (7 files, 57 tests).

willwashburn (commented):
NO-GO. I attempted to submit this as `--request-changes`, but GitHub rejected it with: `Review Can not request changes on your own pull request`. Posting as a review comment with the same request-changes verdict.

I found real protocol/runtime defects against the now-frozen v3.1 contract (relaycast#191 @ ee2c001, relay#1106 @ 7fd17659, local protocol relay#1105 @ b13782fa).

1. [blocker] `agent.register` token authority is still wired to a non-canonical `agent.registered` server message. `ServerToNode` defines `agent.registered` at `crates/broker/src/fleet_wire.rs:330` / `crates/broker/src/fleet_wire.rs:371`, and the control client only completes pending registrations when it sees that variant at `crates/broker/src/node_control.rs:594`. The accepted v3.1 wire sends `type: "reply"` / `type: "error"` with a request `id` for `agent.register`; this PR's outgoing `AgentRegister` also has no request id (`crates/broker/src/fleet_wire.rs:106`). Result: fleet spawn cannot receive the minted token from the relaycast engine and will sit until `FLEET_AGENT_REGISTER_TIMEOUT`. Port the frozen request/reply/error wire mirror: add request ids, parse `reply`/`error` by id, complete/reject the pending registration from `data`, and remove `agent.registered`.

2. [blocker] Clean shutdown still does not emit the real remote `node.deregister` frame. The sidecar sends a local `deregister_node` request before close (`packages/cli/src/cli/lib/fleet-sidecar.ts:233`), but the Rust local protocol enum has no `DeregisterNode` variant around `crates/broker/src/protocol.rs:408`, so that frame is rejected and swallowed. Broker shutdown also just returns on `FleetControlCommand::Shutdown` at `crates/broker/src/node_control.rs:551` without sending `BrokerToRelaycast::NodeDeregister`. Add the local protocol variant from relay#1105, handle it in `runtime/fleet.rs` by forwarding `node.deregister`, and on control-client shutdown send `node.deregister` before closing the websocket. Add a test that the mock relaycast socket receives `type: "node.deregister"` before close.

3. [blocker] `spawn_agent.invocation_id` is dropped at the Rust boundary, so concurrent spawn attribution is broken. The TS local protocol includes `invocation_id` in `spawn_agent` (`packages/harness-driver/src/protocol.ts:151`), but the Rust `SdkToBroker::SpawnAgent` payload omits it (`crates/broker/src/protocol.rs:413`) and `runtime/fleet.rs` dispatches spawn without it (`crates/broker/src/runtime/fleet.rs:107`). The broker then guesses with `single_in_flight_invocation()` at `crates/broker/src/runtime/fleet.rs:385`; with two concurrent spawn invokes this returns `None`, so `agent.register` and `inventory.sync` lose the originating invocation id. Add `invocation_id?: string` to the Rust protocol, pass it through `handle_fleet_spawn_agent`, and cover two concurrent spawn requests with distinct ids.

4. [major] The Rust fleet fixtures are not byte-identical to the accepted v3.1 fixtures. For example, this PR models `agent.deregister` as `{ name }` (`crates/broker/src/fleet_wire.rs:131`) and the fixture contains only `name`, while the frozen schema requires `agent_id` with optional `name`. The Rust fixture set also lacks the frozen `reply.json`, `reply.agent_register.json`, and `error.json` frames, and request fixtures omit `id` fields. Replace the local fixtures/types with the accepted v3.1 mirror from relay#1106/7fd17659 and make the parity test fail on directory diffs against the canonical fixtures, not just self-round-trips.

5. [major] Trigger sync does not re-enable an existing disabled trigger. `syncTriggers()` only calls `update()` when `!triggerEquals(primary, trigger)` (`packages/cli/src/cli/lib/fleet-sidecar.ts:378`), but `triggerEquals()` never reads `left.enabled` and only checks that the desired trigger is enabled (`packages/cli/src/cli/lib/fleet-sidecar.ts:428` / `packages/cli/src/cli/lib/fleet-sidecar.ts:433`). If Relaycast already has the same trigger with `enabled: false`, fleet serve treats it as equal and leaves the node trigger inert. Include `left.enabled === right.enabled` in equality and add coverage for a disabled existing trigger being re-enabled.

6. [minor] The PR body still says the MCP §5 resources/subscription surface was removed, but the code has restored it (`packages/cli/src/cli/agent-relay-mcp.ts:431`, `packages/cli/src/cli/agent-relay-mcp.ts:1751`, `packages/cli/src/cli/agent-relay-mcp.ts:1872`). Update the PR body to document the round-2 scope decision honestly: legacy resources/subscriptions are restored and deletes are deferred to the flag-gated rollout/audit phase.

Validation run on `/tmp/review-Rev4x3` at `1bf0f917`:
- `npm run build` initially failed because the fresh worktree had no dependencies (`tsc: command not found`); after `npm install`, `npm run build` passed.
- `npx vitest run` passed: 64 files / 822 tests passed, 1 file / 5 tests skipped.
- `cargo test -p agent-relay-broker fleet` passed.
- `cargo test -p agent-relay-broker node_control` passed.

willwashburn (commented):
Round 4 adversarial review result: NO-GO.

Note: I attempted to submit this with `gh pr review --request-changes`, but GitHub rejected it because this authenticated identity is treated as the PR author. Posting as a review comment with the same NO-GO findings.

Validation I ran locally on a fresh worktree at `625eed06298c774a282ed927c87e4047a52e63f6`:

- `git merge-base --is-ancestor 642ad769 HEAD` passed.
- `npm ci` was needed because the fresh worktree had no local toolchain.
- `npm run typecheck` passed.
- `npm run build` passed, including the release Rust broker build.
- `npm test` passed: 64 files passed, 824 tests passed, 5 skipped.
- `cargo test` passed: 756 unit tests passed, 12 continuity tests passed, fleet wire fixture test passed, 4 ignored.

Findings:

1. **major** `packages/harness-driver/src/protocol.ts:82`, `crates/broker/src/protocol.rs:248`, `crates/broker/src/node_control.rs:324` - capability metadata is still typed as arbitrary JSON on the local protocol, then the broker silently drops non-object metadata before emitting `node.register`. The accepted relay#1105 round-6 contract tightened manifest capability metadata to `Record<string, JsonValue>`, and `@agent-relay/fleet` already follows that shape. But `NodeCapabilityManifest.metadata?: JsonValue` still lets direct harness-driver consumers send `"metadata": "gpu"` or arrays, and Rust accepts `Option<Value>` and uses `.as_object()`, so the field disappears instead of being rejected or preserved. Please make the local protocol object-only on both sides (`metadata?: Record<string, JsonValue>` in TS, object map type/deserializer in Rust), reject scalar/list metadata, and add TS/Rust tests for the rejected cases.

2. **major** `packages/fleet/src/index.ts:100`, `packages/fleet/src/index.ts:343`, `packages/sdk/src/messaging/types.ts:70`, `packages/fleet/src/index.test.ts:117` - `onMessage` accepts a JavaScript `RegExp`, but trigger sync serializes only `RegExp.source`, dropping flags. The current test even passes `/ship/i` and expects `pattern: "ship"`, which means a declared case-insensitive trigger is synced as case-sensitive and can silently miss matching messages. The Relay trigger input only has `pattern`, with no flags field, so either serialize in a backend-supported form that preserves flags or reject non-empty `RegExp.flags` with a clear error. Update the tests so flagged regexes cannot silently degrade.

3. **minor** `packages/cli/src/cli/agent-relay-mcp.ts:393`, `packages/cli/src/cli/agent-relay-mcp.ts:431`, `packages/cli/src/cli/agent-relay-mcp.ts:579`, `packages/cli/src/cli/agent-relay-mcp.ts:1871` - the PR summary and Phase 4 plan still say this removes the §5 MCP resource/subscription surface (`RealtimeResourceBridge`, resource registrations, and `enableInboxPiggyback`), but the implementation keeps and registers all of it, with tests asserting legacy resource behavior. If this restore is intentional from the accepted control-plane stack, please update the PR body/plan references so reviewers do not keep checking for a deletion that is no longer desired. If the Phase 4 cleanup is still in scope, remove these surfaces and the legacy assertions.

willwashburn:
## Round-4 fixes — pushed at `c20cd840`

Stack-synced both dependency tips into this branch (`feat/fleet-local-protocol`@`abfd5ff0`, `feat/fleet-control-plane`@`aae4d28e`), both merged cleanly with no conflicts.

**Finding 1 (major) — capability metadata must be object-only on both sides.** Resolved via the synced stack:
- TS: `NodeCapabilityManifest.metadata?: Record<string, JsonValue>` (`packages/harness-driver/src/protocol.ts:82`), with compile-time `@ts-expect-error` assertions rejecting scalar and array metadata.
- Rust: `metadata: Option<HashMap<String, Value>>` (`crates/broker/src/protocol.rs:248`) — deserialization now *errors* on a scalar/list instead of `.as_object()` silently dropping it. Covered by `sdk_register_node_capability_metadata_requires_object`, which asserts `serde_json::from_value::<SdkToBroker>(... "metadata": ["not","an","object"] ).is_err()`.
- `crates/broker/src/node_control.rs:320` now iterates the typed `HashMap` map rather than re-inspecting arbitrary JSON, so non-object metadata can never reach the `node.register` emit — it's rejected at parse time. No silent drop remains.

**Finding 2 (major) — `onMessage` RegExp flags silently dropped.** Fixed in `defineNode` (`packages/fleet/src/index.ts:186`): a trigger whose `match` is a flagged `RegExp` is now **rejected** with a clear error — `trigger regex flags are not supported yet; match is case-sensitive — use character classes like [Ss]hip` — rather than serializing only `.source`. Flags support is a planned fast-follow; the trigger wire schema is unchanged. Added a validation test, updated the serialization tests off `/ship/i`→`/[Ss]hip/` and the CLI sync tests off `/hello/i`→`/hello/` so no flagged regex can silently degrade, and documented the constraint in the new `packages/fleet/README.md`.

**Finding 3 (minor) — stale §5 MCP-removal claim.** The implementation intentionally **retains** `RealtimeResourceBridge`, the resource registrations, and `enableInboxPiggyback` (confirmed present/registered at `agent-relay-mcp.ts:393/579/1806/1871`) — the removal was superseded by the accepted control-plane stack. Updated the PR body to describe the surface as retained rather than removed.

### Validation
- `npm run typecheck` ✅
- `npm test` ✅ — 64 files / 824 tests passed, 5 skipped
- `cargo test -p agent-relay-broker` ✅ — 760 unit + 12 continuity + fleet wire fixtures, 0 failed
- Sidecar↔broker WS smoke against the **rebuilt merged** broker binary: `hello` → `register_node` → `register_handlers` → `spawn_agent`(with `invocation_id`) → `deregister_node` all pass; `invocation_id` is threaded into `handle_fleet_spawn_agent`/`register_fleet_agent_token` before any spawn-exec step.

willwashburn (commented):
## Round 5 — FINAL VERIFY: GO ✅

Reviewed at tip `7a525960` (origin advanced past the `c20cd840` named in the round-5 brief; the only delta is a 2-line Prettier reformat of `packages/fleet/README.md`, no logic change). Ran typecheck, full build (incl. release Rust broker), `npm test`, and `cargo test -p agent-relay-broker` myself — all green. Walked each round-5 focus item against the running scenario.

### Local validation (re-run, not trusted from the PR body)
- `npm run typecheck` → pass
- `npm run build` → pass; release broker built (`target/release/agent-relay-broker`)
- `npm test` → **64 files / 824 tests passed**, 1 file / 5 skipped
- `cargo test -p agent-relay-broker` → **760 unit + 12 continuity + 1 fixture-parity, 0 failed**

### (1) Stack contains both dep tips ✅
`git merge-base --is-ancestor abfd5ff0 HEAD` and `aae4d28e HEAD` both return ancestor. The control-plane and local-protocol tips are merged in (`cea4f77e`, `a10f8aec`).

### (2) Capability metadata is object-only on both sides, errors (not silently drops) on non-object ✅
- TS: `NodeCapabilityManifest.metadata?: Record<string, JsonValue>` with real negative `@ts-expect-error` tests for scalar (`metadata: 'v1'`) and array (`metadata: ['v1']`) — `packages/harness-driver/src/protocol.ts:93-102`.
- Rust SDK frame: `NodeCapabilityManifest.metadata: Option<HashMap<String, Value>>` (`protocol.rs:248`); negative test `sdk_register_node_capability_metadata_requires_object` asserts `serde_json::from_value::<SdkToBroker>(… metadata:["not","an","object"]).is_err()` (`protocol.rs:849-861`). Serde fails the whole frame rather than dropping the field.
- Wire layer: `fleet_wire::FleetCapability.metadata: Option<BTreeMap<String, Value>>` with `#[serde(deny_unknown_fields)]` (`fleet_wire.rs:55-71`). Manifest→wire mapping preserves every key (`node_control.rs:326-331`).

### (3) Flagged regexes rejected at `defineNode` with clear error + test + README ✅
`defineNode` throws on any flagged `RegExp` trigger — `if (trigger.match instanceof RegExp && trigger.match.flags) throw …` (`packages/fleet/src/index.ts:186-190`), message points users at `[Ss]hip` character classes. Not silently stripped. Test: `onMessage({ match: /ship/i }, 'run')` ⇒ `toThrow(/trigger regex flags are not supported yet/)` (`index.test.ts:168-170`). README documents it (`README.md:64-71`) with ✅/❌ examples.

### (4) Round-3 items still hold post-merge ✅
- **SIGTERM → deregister → wire `node.deregister`:** `fleet.ts:144-145` wires SIGINT/SIGTERM to `controller.abort()` → `serveFleetSidecar` `abort()` → `close()` sends `deregister_node` (only when `nodeRegistered`) (`fleet-sidecar.ts:232-243`). Broker emits `BrokerToRelaycast::NodeDeregister { v:1 }` and clears inventory/handlers_live (`node_control.rs:606-622`); covered by `node_control_deregister_sends_node_deregister`.
- **spawn `invocation_id` attribution under concurrency:** per-frame `invocation_id` threaded through `handle_fleet_spawn_agent` → `register_fleet_agent_token` → `InventoryAgent` (`fleet.rs:472-516`). TS test "threads invocation ids through concurrent spawn handlers" proves `inv-a→worker-a`, `inv-b→worker-b` under `Promise.all` (`index.test.ts:82-109`).
- **Trigger re-enable:** `syncTriggers` is active and idempotent (dedup by key, update primary, delete duplicates), `enabled: true` (`fleet-sidecar.ts:347-398`, `index.ts:345-355`).
- **Minted-token delivery via canonical v3.1 reply:** `reply`/`error` are `ok`-type-constrained (`deserialize_true_bool`/`deserialize_false_bool`, `fleet_wire.rs:398-466`); `complete_agent_registration` validates `data.{agent_id,token}` non-empty, correlates by `id`, and emits a compensating `agent.deregister` on late success (`node_control.rs:720-776`). Covered by `node_control_agent_register_round_trips_minted_token` and the `reply.agent_register.json` parity assertion.

### Wire-format & exhaustiveness spot-checks
- Envelope `v:1`, discriminant `type`, snake_case fields consistent across all 15 fixtures; `fleet_wire_fixtures_round_trip_semantically` asserts exact file-set + type-set coverage and byte-stable round-trip — a real, non-vacuous parity gate.
- `SdkToBroker` and `RelaycastToBroker` dispatch matches are exhaustive (no catch-all; compiles), so new variants can't be silently ignored.
- `§5` MCP resource surface (`RealtimeResourceBridge`, resource registrations, `enableInboxPiggyback`) retained per accepted control-plane stack; `query_nodes` + `spawn` tools added (`agent-relay-mcp.ts:1220,1669`). The MCP `spawn` tool wraps the legacy broker-owned `spawn` action (plan §4.4), distinct from node `spawn:<cli>` capabilities — verified intentional, not a mismatch.

### Notes (non-blocking, no action required)
- `node.register.json` uses `kind:"run"` for `run:test`; the TS producer emits `kind:"action"` for non-spawn actions. `kind` is free-form `Option<String>` on the wire so this is harmless, but the fixture value is illustrative-only and doesn't mirror a real producer. Nit.
- Live two-process sidecar↔broker smoke was covered via the deterministic node_control/fleet integration tests + fixture parity (which exercise hello/register/deregister/minted-token/invocation_id end-to-end at the wire layer) rather than a fresh live run requiring a workspace key.

No blocker/major/minor defects found. **GO.**

PR #1128, opened 2026-06-14
fix(cloud): serialize auth refresh writes

khaliqgant:
## Summary
- write cloud auth through a pid-scoped sibling temp file, chmod 0600, then atomic rename
- serialize file-backed read-refresh-write with a stale-aware cross-process lock and re-read/re-check stored auth under the lock
- route file-backed CloudApiClient refreshes, including 401-forced refreshes, through the locked refresh path

## Tests
- npx vitest run packages/cloud/src
- npm run build:config && npm --prefix packages/cloud run build

<a href="https://cubic.dev/pr/AgentWorkforce/relay/pull/1128?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `@agent-relay/cloud` now writes cloud auth atomically and serializes file-backed token refreshes across processes, preventing concurrent refreshes from clobbering rotated credentials.
 - `@agent-relay/cloud` refresh now fails with typed, timeout-bounded errors and migrates legacy `~/.agent-relay/cloud-auth.json` credentials into the canonical `~/.agentworkforce/relay/cloud-auth.json` store without dual-writing.
 - `agent-relay-broker` persists pending deliveries on shutdown and on every queue change, redelivers them on restart, reports timeout-fallback verification explicitly, and emits `delivery_dropped` when the per-worker queue cap evicts a message.
 
--- a/packages/cloud/src/api-client.ts
+++ b/packages/cloud/src/api-client.ts
@@ -8,6 +8,10 @@ export type CloudApiClientOptions = {
   accessTokenExpiresAt: string;
   refreshTokenExpiresAt?: string;
   refreshTimeoutMs?: number;
+  refreshAuth?: (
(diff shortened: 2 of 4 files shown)


khaliqgant:
✅ **AR-252 review verdict: APPROVE** (posting as a comment — GitHub blocks a formal approval from the same bot identity that opened the PR).

Reviewed against the daemon-vs-reaper concurrent-refresh failure this targets. Verified independently by checking out the branch: `vitest packages/cloud/src` → **111 passed** (106 baseline + 5 new), `tsc` build clean with `dist/` emitting the lock + atomic rename. CI all green (E2E + Devin still running, no failures).

**Correct on the point that matters.** The cross-process lock spans the **full** read→refresh→write critical section in `refreshStoredAuth`, not just `writeStoredAuth` — with a re-read + `shouldRefresh` re-check under the lock so the race loser returns the freshly-rotated token instead of replaying an invalidated single-use refresh token. The `serializes concurrent file-backed refreshes` test nails it: both callers resolve to the same rotated token **and** `fetch` is asserted to fire exactly once — precisely the assertion that catches a write-only-lock regression.

Other strengths: atomic pid/time/uuid temp + chmod 0600 + rename (temp cleaned in finally); env-backed auth bypasses the lock; native `fs.mkdir` EEXIST lock (no new dep); lock released on throw with a dedicated test; 401-forced refresh routed through the locked path; in-process `refreshPromise` still dedups (no self-deadlock); `onRefresh` dropped so no double-write; stale window 30s ≫ 10s refresh timeout. Preserves pear#323's atomic write; unblocks pear#324.

**Non-blocking** (optional, do not hold merge): (1) `authorizedApiFetch` `force: true` always network-refreshes on 401 even if a peer just rotated — correct/safe, minor optimization to retry with the re-read token first; (2) `writeFile` 0600 + explicit `chmod` 0600 is redundant but fine as umask defense; (3) stale-lock removal has an inherent TOCTOU, acceptable for advisory locking given the timing margin.

Cleared to merge at the maintainer's discretion (per repo policy I'm not merging).

PR #1131, opened 2026-06-14
test(e2e): two-node fleet scenario matrix in CI (Phase 6)

willwashburn:
## Phase 6 — two-node fleet E2E in CI (RFC #1056)

Boots a **real** stack — a relaycast engine (node adapter) plus two
`agent-relay fleet serve` nodes, each its own **Rust broker + TS sidecar** —
and drives the fleet control wire end-to-end. Unlike the unit suites (fake
broker WS / in-process adapter), this exercises the actual broker `node_control`
connection, including the `Authorization: Bearer` node-token handshake and the
`agent.register` token-authority reply (the two engine⇄broker mismatches this
E2E surfaced; fixed in relaycast#194).

### Scenario matrix (13, all green, ~28s locally)
| Scenario | Asserts |
|---|---|
| boot/register | both nodes online + handlers_live via real broker Bearer auth; correct caps |
| negative auth | a node with a bogus token never reaches handlers_live (auth enforced) |
| capability query | roster filtered by capability returns the right node(s) |
| cross-node dispatch | echo→node-a, ping→node-b dispatch over the control connection + ack |
| declarative trigger | fires once; the action-generated reply does not re-trigger — asserted on the `echo:` **prefix** total (a broken guard cascades to distinct strings) |
| spawn completes E2E | targeted spawn mints+injects the token, binds the agent via-node — token-authority regression guard |
| capability-routed spawn | placement picks the only eligible node |
| scheduled spawn | shared-capability spawn routes to the least-loaded node |
| resume | resumable spawn carries session_ref; resume re-targets the origin node |
| placement failure | unsatisfiable capability → `capability_mismatch` (409) |
| reschedule + reconcile | in-flight invocation reruns on the other node; handlers_live drops→restores; the rescheduled invocation is **not re-claimed** after restart |
| delivery seq/dedup | strictly monotonic `agent_seq`, no duplicates; a resync replays only the tail (`gap_detected:false`, no duplicate seqs) — the exactly-once cursor reconcile relies on |
| mailbox TTL | undelivered message dead-letters **and the sender is notified** (`delivery.failed`) |

### Coverage sign-off (not re-asserted here)
- **overflow reject-new** (`belowDepthCapSql` + `notifyDeliveryRejections`): demonstrating
  it E2E needs a recipient whose deliveries QUEUE (a via-node agent on a down node) AND
  an observable ledger — the spawned agent's token is held by the broker; a self-connected
  recipient auto-delivers. **Signed off as covered by the relaycast engine §8.3 mailbox
  conformance matrix.** The monotonic-seq/dedup half of §8 IS asserted here.
- **self spawn** (`target:self`): needs a via-node caller; covered by engine placement conformance.

### CI / reproducibility
`.github/workflows/fleet-e2e.yml` triggers on `feat/fleet-**` PRs (was inert on `[main]`),
builds the relay core + Rust broker, checks out the relaycast engine **pinned to the
relaycast#194 SHA** carrying the compat fixes (the harness resolves the same branch, so
local == CI), and runs the matrix. Matrix ~30s; the 20-min job timeout is the cold-build ceiling.

### Version skew
Old v8.x brokers keep working (fleet flag off); fleet nodes require new broker + engine + flag on.


willwashburn (commented):
## Adversarial review — round 1 (RevE1)

Scope: this is the Phase 6 two-node E2E harness, the phase's exit gate. I read the harness, the two node definitions, the scenario test, the CI workflow, and cross-checked the broker source (`crates/broker/src/node_control.rs`, `telemetry.rs`) for the node-id/Bearer plumbing. I could **not** run the full matrix locally — the relaycast engine repo is not checked out here and neither the broker nor the CLI are built; building the separate engine repo is out of scope for this pass. Findings below are from static analysis + broker-source verification.

Verdict: **request changes.** The headline problem is that the one scenario the orchestrator flagged as critical — the loop guard — is asserted vacuously, and the CI job that is supposed to enforce this gate does not run on this PR at all.

---

### 1. [BLOCKER] Loop-guard assertion is vacuous — the declarative-trigger test passes whether or not the loop guard works
`tests/e2e/fleet/fleet-e2e.test.ts:179-182`

The test posts `please deploy now`; the echo handler emits `echo:please deploy now`. The final assertion is:
```ts
const after = (await listMessages(...)).filter(m => m.text === 'echo:please deploy now');
expect(after.length).toBe(1);
```
But if the loop guard were **broken**, `echo:please deploy now` (which contains `deploy`) re-triggers and the handler emits `echo:echo:please deploy now` — a **distinct** string. Each cascade level is a new unique string, so the count of the *exact* original string is always exactly 1, in both the correct and broken cases. The test can only fail if the trigger never fires (already covered by the `waitFor` at :169-177) — it cannot detect a loop-guard failure. The comment at :167-168 ("echo messages would grow without bound; assert it settles at exactly one") describes the right idea but the assertion doesn't implement it.

Fix: assert the total number of `echo:`-prefixed messages is exactly `before + 1`, e.g. `expect(echoes.length).toBe(before + 1)` using the same prefix filter as the `waitFor`. As written this is the exit criterion "trigger fires exactly once (loop guard verified)" and it verifies nothing.

### 2. [MAJOR] Phase 6 exit matrix is incomplete — six of the required scenarios are absent
Plan `fleet-implementation-plan.md:126` lists the Phase 6 E2E matrix: boot/register, cross-node message, capability query, **targeted + scheduled spawn**, declarative trigger, sidecar crash (drop + **reschedule** + restart), **node restart reconcile**, **mailbox TTL → dead-letter + sender notify**, **overflow reject-new**, **resume via origin-targeted spawn**. This PR implements boot/register, capability query, cross-node dispatch, declarative trigger, sidecar crash (drop+restart only), and targeted/capability-routed spawn placement. Missing: scheduled spawn, self spawn, node-restart reconcile w/ seq+dedup, mailbox TTL dead-letter + sender notify, overflow reject-new, resume via origin-targeted spawn. The PR title/body frame this as *the* Phase 6 E2E gate; as the gate it does not meet the exit criterion. Either land the remaining scenarios or re-scope the PR so it does not claim to close Phase 6.

### 3. [MAJOR] "reschedule" and "no dup deliveries / seq+dedup" are claimed but never asserted
`tests/e2e/fleet/fleet-e2e.test.ts:204-213`

The plan requires sidecar-crash → **reschedule** and node-restart → **reconcile with no duplicate deliveries, seq/dedup asserted**. The crash test kills node-b, restarts it, and fires a *fresh* ping. It never (a) has an in-flight invocation outstanding at crash time to prove rescheduling, nor (b) asserts anything about duplicate delivery or sequence/dedup — the comment at :204 says "dispatch still lands on node-b with no duplication" but there is no dedup/seq assertion in the body. The "no duplication" guarantee is untested.

### 4. [MAJOR] CI job does not run on this PR (or any stacked feature-branch PR)
`.github/workflows/fleet-e2e.yml:20-21` gates `pull_request.branches: [main]`, but this PR's base is `feat/fleet-sdk-cli`. The Fleet E2E job therefore does **not** execute on #1131 or any of the stacked fleet PRs — only after something merges to `main`. The gate that is supposed to protect the feature during development is inert for the entire feature stack. Add the integration base branch(es) (e.g. `feat/fleet-*`) to the trigger, or confirm the intended merge path actually exercises it before relying on it as a gate.

### 5. [MAJOR] Engine pinned to a moving branch, and local vs CI build different engine code
`.github/workflows/fleet-e2e.yml:54` checks out relaycast at `ref: feat/fleet-mailbox` (an unpinned, moving branch — non-reproducible; the E2E can silently change/break when that external branch moves). Meanwhile the PR body states the engine-side Bearer fix (#192) lives on the **rollout-flag** branch, and `tests/e2e/fleet/harness.ts:27` prefers a local `fleet-rollout-flag` worktree *before* `fleet-mailbox`. So a dev runs against rollout-flag (has the fix) while CI runs against fleet-mailbox. If fleet-mailbox lacks the fix, CI is either red or the "Bearer regression guard" is invalid; if it has the fix, the PR body's pointer is stale. Pin to a SHA and make local resolution match what CI builds.

### 6. [MINOR] Bearer "regression guard" is one-directional — no negative auth test
`tests/e2e/fleet/fleet-e2e.test.ts:104-117`. The boot/register happy path *would* fail (via the `beforeAll` `handlers_live` timeout) if the broker stopped sending / mismatched the `Authorization: Bearer` header — good, that catches the historical #1107↔#192 bug. But there is no negative assertion that a node presenting a bad/missing token is **rejected**, so a regression that disables auth enforcement on the engine's node_control connection passes undetected. Add one node that enrolls with a bogus token and assert it never reaches `handlers_live`.

### 7. [MINOR] Unsatisfiable-spawn assertion is too loose
`tests/e2e/fleet/fleet-e2e.test.ts:248`: `expect(spawnNone.status).toBeGreaterThanOrEqual(400)` passes on any 4xx/5xx, including a 500 from an unrelated failure or a generic 404. Assert the specific capability-mismatch status/error code/body so a misroute that 500s doesn't read as a clean placement failure.

### 8. [MINOR] Trigger test relies on a fixed `delay(2_000)`
`tests/e2e/fleet/fleet-e2e.test.ts:178`. Fixed sleeps make the suite slower and are inherently racy for "nothing else happened" assertions. Combined with #1 this is the only thing standing in for loop-guard verification. Prefer a deterministic settle (poll until count stable across N intervals).

### 9. [NIT] Channel `#general` vs `general` inconsistency; node-file triggers are never exercised
`tests/e2e/fleet/nodes/node-a.ts:50` declares `onMessage({ channel: '#general', ... })`, but the test registers the trigger via the engine API with `channel: 'general'` (`fleet-e2e.test.ts:159`) and posts/lists against `general`. Because the sidecar trigger auto-sync is (knowingly) not wired, the node-file `triggers: [...]` block is dead in this test — the firing path under test is the engine-registered one, not the node definition. When auto-sync lands, the `#`-prefix mismatch will bite. Worth normalizing now and noting that the node-file trigger surface is currently uncovered.

### 10. [NIT] Stated runtime budget vs timeout
`.github/workflows/fleet-e2e.yml:43` sets `timeout-minutes: 20` while the PR body and header comment claim "under the 10-minute budget." A cold `rust-cache` + `cargo build --release` + two `npm ci` + engine build will likely exceed 10 min on first run. Either reconcile the claim or document that 10 min is steady-state (warm cache) only.

---

Items 1 and 4 should block: the gate's critical scenario doesn't actually test its feature, and the gate doesn't run on the PRs it's meant to guard. 2/3/5 are required for this to legitimately close Phase 6.

willwashburn (commented):
## RevE2 — adversarial review, round 2 (head bc3be512) — **VERDICT: NO-GO**

(Posting as a comment — GitHub blocks `--request-changes` on a self-owned PR. Treat the 2 MAJORs below as blocking the GO.)

**Build & matrix run locally: GREEN.** Built the head broker (`crates/broker` — has `node_control.rs`) + used the relaycast engine worktree pinned to the CI SHA `b673dfb`; `npm run test:e2e` → **12/12 passed (~29s)**. Confirmed local==CI (my first run failed only because of a stale pre-fleet broker binary; rebuilt from head → all green).

**Round-1 fixes I verified as genuinely fixed:**
- (1) Loop-guard test **non-vacuous** — disabled guard: the re-broadcast `echo:please deploy now` still contains `deploy`, so it cascades to `echo:echo:…` (all `echo:`-prefixed), count grows past `before+1`, `expect(settled).toBe(before+1)` fails. ✅
- (4) CI fires on this PR's base: `pull_request.branches: [main,'feat/fleet-**']` matches base `feat/fleet-sdk-cli`. ✅
- (5) Engine pinned `b673dfb`; local engine worktree at exactly that SHA. ✅
- (6) Negative-auth test present + passing. ✅
- (7) **Spawn-completes-E2E** would catch the `agent.register` regression: broken reply frame hangs the spawn → `waitFor(...'completed',20s)` throws → test fails. ✅

### MAJOR

**1. Reschedule "idempotency" assertion is vacuous — `tests/e2e/fleet/fleet-e2e.test.ts:426-428`**
```ts
const again = await getInvocation(engine, driverToken, 'ping', ping.invocationId!);
expect(again.status).toBe('completed');
expect(again.output).toMatchObject({ pong: 'after-restart' });
```
`again` re-reads the *same* invocation by id right after `pingDone` already saw it `completed`; the stored record can't change on re-read, so these expects can never fail — and a dup dispatch yields identical output under the same id, so it can't catch a dup either. Round-2 required **seq/dedup asserted**; the `seq`/`msg_id` from `listDeliveries` are never asserted anywhere. **Fix:** assert real dedup (side-effect counter on the handler asserted to run exactly once across crash+reschedule, or capture delivery `seq`/`msg_id` and assert no dup / contiguous cumulative ack). If dedup truly isn't observable here, drop the misleading `stays idempotent` claim rather than ship a test that can't fail.

**2. Mailbox TTL asserts dead-letter but not sender-notify — `fleet-e2e.test.ts:462-474`**
Plan Phase-6 matrix (`fleet-implementation-plan.md:126`) is explicit: *"mailbox TTL → dead-letter **+ sender notify**"*. Only `dead_lettered` is asserted. The sender is a controllable registered agent (`sender`, line 451), so the `delivery_failed`→sender event **is** assertable here (unlike overflow). **Fix:** assert the sender observes the dead-letter / `delivery_failed` notification.

### MINOR

**3. Overflow reject-new not asserted — `fleet-e2e.test.ts:477-487`** — also an explicit Phase-6 cell (`plan:126`), deferred to engine §8.3 conformance with a real documented limitation (needs a queuing + observable recipient). Defensible, but either assert it via a via-node recipient on a down node, or get explicit sign-off that engine-conformance coverage satisfies the exit criterion.

### NITS
4. `harness.ts:514` — `getPresence` exported but never used (dead code).
5. `harness.ts:17` — stale docstring says CI uses `feat/fleet-mailbox`, but the workflow pins `feat/fleet-rollout-flag@b673dfb` (relaycast#194).
6. `fleet-e2e.test.ts:163` — negative-auth node-c uses `NODE_B_FILE` (internal name `node-b`) with `--name node-c`; harmless but inconsistent. Fixed `delay(8_000)` adds wall-time / is timing-fragile.
7. `nodes/node-a.ts:50` — node-file `triggers:[onMessage(...,'echo')]` is never exercised (sidecar auto-sync unwired; trigger registered via engine API). Dead fixture that can silently rot.

Net: wire is real, matrix green, round-1 regression guards solid — but two explicitly-required round-2 criteria (seq/dedup asserted; sender-notify) are unmet and one assertion can never fail. Fix #1 and #2 (+ decide #3) and I expect a GO next round.

willwashburn (commented):
## Adversarial review — round 3 (head 61e8bf7b) — **NO-GO**

Exit criterion for this round: every one of the 13 two-node scenarios must have an assertion that *genuinely fails when the feature breaks*. I reasoned each broken-feature case through by hand. One round-2 deliverable — the new exactly-once-over-reconnect assertion — does not fail on the bug it is meant to guard. Details below; the rest of the matrix verified good.

> Note: I could not execute the matrix locally in the timebox. A faithful run needs `npm ci` + `npm run build:core` + `cargo build --release` on this branch, and the only local relaycast engine is `main@ee64c94`, not the pinned relaycast#194 `b673dfb`, so the spawn/handshake scenarios would not run truthfully. All five new files parse clean (esbuild/`node --check`). Findings are from static reasoning.

### 1. [MAJOR] `delivery seq/dedup` replay assertion is vacuous — does NOT fail on a broken exactly-once cursor
`tests/e2e/fleet/fleet-e2e.test.ts:469-472`
```ts
const replayed = replay.seqs().filter((s) => s > cursor);
expect(new Set(replayed).size).toBe(replayed.length);     // no dup (tail only)
expect(replayed).toEqual([...replayed].sort((a,b)=>a-b)); // monotonic (tail only)
expect(Math.min(...replayed)).toBeGreaterThan(cursor);    // "no re-delivery of acked"
```
The property this scenario exists to prove is *no re-delivery of already-acked messages* (seqs ≤ `cursor`) after a resync. But `replayed` is **pre-filtered to `s > cursor`**, so any redelivered head seqs (≤ cursor) are discarded before every assertion. Reason it through:
- **Broken cursor that over-delivers** (ignores `last_seen_seq`, replays the full history `s0..s3`): `replay.seqs()` = `[s0,s1,s2,s3]`, `.filter(s>cursor)` = `[s2,s3]` → all three asserts pass. The over-delivery of the acked `s0,s1` is invisible. **Test passes on the bug.**
- **Broken cursor that under-delivers / replays nothing**: `replayed = []`; `Math.min(...[])` = `Infinity > cursor` ✅, `new Set([]).size === 0 === [].length` ✅, `[].toEqual([])` ✅. **An empty replay also passes.**

So the assertion fires on neither over- nor under-delivery — exactly the failures `resync` must not have. The `resync_ack` matcher (`gap_detected:false`, `current_seq:maxSeq`) only checks the *ack frame*, not the bytes actually streamed. This is the headline round-2 addition ("exactly-once cursor over a WS reconnect") and it is non-asserting.
**Fix:** assert against the *unfiltered* replayed dm stream and pin the exact tail, e.g.
```ts
const replayedDm = replay.ofType('dm.received').map((e) => e.agent_seq as number);
expect(replayedDm).toEqual(seqs.filter((s) => s > cursor)); // exact tail: catches over- AND under-delivery
```

### 2. [MINOR] `reschedule` "not re-claimed" final read happens before a re-run could surface
`tests/e2e/fleet/fleet-e2e.test.ts:423-427` — after restart it waits `delay(2_000)` then asserts `post.output.node === otherName`. The `work` payload is `delayMs: 6_000`. If a broken reconcile re-claims and re-dispatches the in-flight invocation on the restarted home node, that re-run completes ~6s later and only *then* could overwrite `output.node` to `homeName`; the read at +2s still shows `otherName` → false pass. (The narrower failure mode where re-claim immediately resets status away from `completed` *is* caught by `expect(post.status).toBe('completed')`, so the assertion is not fully vacuous.) **Fix:** read after the work delay elapses post-restart (`delay(7_000+)`), or assert no second dispatch occurred (e.g. `dispatched_node_id` unchanged / home node's `active_agents` did not bump for a re-run).

### 3. [MINOR] `resume` assertion is weaker than its claim
`tests/e2e/fleet/fleet-e2e.test.ts:336-365` — the resume invoke passes `target_node: 'node-a'` explicitly (L360), then asserts `handler_node_id === 'node_a'` (L364) under the banner "resume re-targets the **origin** node". With an explicit target, this only proves targeting works, not that the engine re-derives the origin from `session_ref`. To actually exercise the resume contract, omit `target_node` on the resume and assert it still lands on `node_a` from `session_ref` alone.

### 4. [MINOR] CI "Upload node logs on failure" is a no-op — failure diagnostics never reach the artifact
`harness.ts:279` sets `this.logPath = …/serve.log` but nothing ever writes it; sidecar output only accumulates to the in-memory `lastLog` (`harness.ts:332/335`). The workflow step globs `/tmp/fleet-e2e-*/**/serve.log` (`.github/workflows/fleet-e2e.yml:109-115`) — it will always match nothing (`if-no-files-found: ignore` hides it), and `cleanupTmp` removes the dir regardless. So when CI fails there are no logs to debug from. **Fix:** flush `lastLog` to `logPath` on `stop()`/exit (or append on each chunk).

### 5. [MINOR] event-name consistency: plan says `delivery_failed`, test asserts `delivery.failed`
`fleet-e2e.test.ts:497` asserts `ofType('delivery.failed')`; plan §2.1 / §1.1 name the sender event `delivery_failed` (underscore). If the engine truly emits the dotted form the TTL test is fine (and it must, or this scenario would hang) — but please confirm the wire name is canonical and reconcile the spec wording, since a mismatch here is a silent hang, not a clear failure.

### 6. [NIT] resync client frame omits the versioned envelope
`harness.ts` `resync()` sends `{ type: 'resync', last_seen_seq }` with no `v:1` envelope (plan §0.2 "versioned envelope"). Either the engine is intentionally lenient on inbound client frames (fine — note it) or the test isn't exercising the real wire shape.

---
### Verified good (round-2 fixes + prior items)
- **Loop guard (declarative trigger)** non-vacuous: counts the `echo:` **prefix**, so a broken guard cascades to distinct `echo:echo:…` strings and `pollUntilStable` never settles at `before+1`. ✅
- **Negative auth** non-vacuous: polls for 5s and requires the bogus-token node *never* flips `live`/`handlers_live`. ✅
- **Spawn-completes-E2E** regression guard asserts `completed` (not timeout) + `active_agents` increment. ✅
- **Mailbox TTL → sender `delivery.failed`** (round-2 #2): now asserted via a live sender WS (`fleet-e2e.test.ts:489-497`), times out → fails if the sender notification regresses. ✅
- **Overflow reject-new** (round-2 #3): signed off in PR body + test comment as covered by engine §8.3. ✅ (criterion: "asserted or signed-off" — met)
- **Reschedule reruns on the other node** (part a, L385-390): asserts `output.node === otherName`, fails on a missing reschedule. ✅ (only the *post-restart re-claim* sub-assertion is soft — finding #2)
- **CI fires on PR base**: `pull_request: branches: [main, 'feat/fleet-**']` matches base `feat/fleet-sdk-cli`, and the changed paths are in the trigger set. ✅
- **Engine SHA pin** `b673dfbacfd97d47b5b511fa096daa1bb9a60d3c` (relaycast#194) present in the workflow. ⚠️ minor: local resolves the engine by *worktree directory* (`resolveEngineServe`), not by that SHA, so the "local == CI" claim holds only if the local checkout sits on the same commit.

**Verdict: NO-GO** on finding #1 (the round-2 exactly-once-reconnect deliverable is non-asserting). It is a one-line fix; #2–#4 are cheap hardening. Re-run the matrix after.

willwashburn (commented):
## Round-4 FINAL VERIFY — GO ✅

Reviewed at head `8fb1277a`. Checked out the branch in a fresh worktree, built the relay CLI + Rust broker (`cargo build --release`, 4m09s) + the pinned relaycast engine, and **ran the full matrix on the live two-node stack**: `13 passed (13)` in 66.5s. Also reasoned each scenario through its break-mode by hand.

### Round-3 MAJOR (vacuous resync/dedup replay) — FIXED ✅
The round-2 assertion (`61e8bf7b`) filtered the replayed seqs by `> cursor`, which let **both** failure modes pass vacuously: an empty replay made `Math.min(...[]) === Infinity > cursor` true, and a full acked-history replay was masked because the filter stripped the `<= cursor` seqs. The new code (`fleet-e2e.test.ts:469-486`) closes both holes:
- `expectedTail = seqs.filter(s => s > cursor)` is `[3,4]` (non-empty — length 2).
- `waitFor(... dm.received.length >= expectedTail.length ...)` → **under-delivery** (cursor replays nothing) times out and throws → FAIL.
- `delay(1_000)` settle, then `expect(replayedSeqs).toEqual(expectedTail)` on the **unfiltered** seqs → **over-delivery** (replaying acked history `[1,2,3,4]`) fails the deep-equal → FAIL.

Reasoned through both broken-cursor modes; each now trips the assert. The `resync_ack` `{ last_seen_seq: cursor, current_seq: maxSeq, gap_detected: false }` check is intact. Non-vacuous. ✅

### Round-3 minors
1. **Reschedule re-claim reads past the work window — FIXED** (`:428-433`): the no-re-claim loop is 16×500ms = 8s, comfortably past the 6s `work` `delayMs`, so a re-claim that re-dispatches + overwrites on completion would be observed. ✅
2. **CI log-upload step — LIVE** (`fleet-e2e.yml:103-109`): `upload-artifact@v4` on `failure()`, `if-no-files-found: ignore`. ✅
3. **Resume origin — PARTIALLY addressed (minor, non-blocking)**: the *first* spawn is now untargeted so the origin is engine-derived rather than hard-coded — that is the substantive improvement and it's correct. But the resume call (`:361-369`) still passes `target_node: originName`, and `cli:'pool'` is on both nodes, so `handler_node_id === originId` / `dispatched_node_id === originId` are guaranteed by target-honoring alone; `session_ref` is passed but never shown to influence routing. So this scenario verifies "a targeted re-spawn lands on its target" (already covered by spawn-completes-E2E), not "resume re-binds to origin from `session_ref`" as the matrix title claims. The comment honestly notes the engine "does not auto-route from session_ref", so there may be no stronger assertion available today — but if so, consider renaming the scenario to "targeted-origin re-spawn" so the matrix doesn't over-claim. Not a blocker.

### No regressions in prior-round fixes
Verified intact: loop-guard via `echo:`-prefix count (`:241-254`); negative-auth never-flips-live poll (`:180-185`); spawn-completes token-authority guard (`:257-286`); mailbox-TTL dead-letter + sender `delivery.failed` (`:521-549`); CI trigger includes `feat/fleet-**` (`fleet-e2e.yml:18,28`); relaycast engine SHA pin `b673dfba` (`:66`).

### Coverage sign-offs (acknowledged)
overflow reject-new and self-spawn remain signed off to the engine §8.3 conformance matrix (documented in both READMEs). Consistent with prior rounds.

**Verdict: GO.** The exit criterion holds — every scenario that can be broken now fails when its feature breaks, the live matrix is green, and the one residual item (resume) is a documented non-blocking minor.

PR #1164, opened 2026-06-18
feat(cloud): add redacted log streaming flag

khaliqgant:
## Summary
- add `agent-relay cloud logs --redact`
- pass `redact=true` through `@agent-relay/cloud` `getRunLogs`
- add CLI/SDK coverage plus an Unreleased changelog entry

## Notes
- Default is OFF/raw for compatibility and byte-offset follow behavior is unchanged.
- Cloud API companion PR: https://github.com/AgentWorkforce/cloud/pull/2292
- Review-only handoff; no merge requested.

## Verification
- `npx vitest run packages/cli/src/cli/commands/cloud.test.ts packages/cloud/src/workflows.test.ts`
- `npx tsc -p packages/cloud/tsconfig.json --noEmit`
- `npm --prefix packages/cloud run build`
- `cd packages/cli && npx tsc --noEmit`
- `git diff --check`


khaliqgant:
## Review — `ar-305-review` (LGTM, one non-blocking note)

Clean, tightly-scoped client side. Verified:
- **`--redact` CLI flag** defaults `false`; passes `redact: options.redact === true` into `getRunLogs`. ✓
- **SDK `getRunLogs`** sets `redact=true` on the query string *only* when `options.redact === true`; absent otherwise (no behavior change for existing callers). ✓
- **No client-side redaction logic** — correct: the CLI/SDK request server-side redaction (the `?redact=true` path in cloud #2292), they don't reimplement it. Single source of truth for the patterns. ✓
- Default-off end to end; tests assert flag propagation on both layers.

### Non-blocking note — silent no-op against older Cloud
Because redaction is entirely server-side, running `--redact` against a Cloud deploy that predates the `?redact=true` support (cloud #2292) will **silently print raw logs** — the query param is ignored and the user believes output was scrubbed. For a *security* flag that's a meaningful footgun. Options (any, non-blocking): document that `--redact` requires Cloud ≥ the release carrying #2292; or have the server echo a `redacted: true` field the CLI can assert/warn on when missing. Worth a follow-up given the failure mode is "thinks it's redacted, isn't."

Otherwise good to go. Pairs with cloud #2292 (factory/read-path redaction), which I reviewed separately.

khaliqgant:
Closing — this is factory E2E output for AR-305, which was an **ambiguous canary issue** ('add a CLI flag to redact secrets from log output') created to test the factory dispatch loop, not a real feature request. The factory dispatched it, implemented it, and opened this PR — which actually validates hops a→b→e of the pipeline. (The implementer guessed the design rather than asking a clarifying question because the agent's writeback path was unreachable — fixed in factory#11.) Not for merge.

PR #1170, opened 2026-06-19
Fix fleet serve JS node loading in standalone CLI

### LINKED ISSUE #1169: agent-relay fleet serve <node-def> broken in published bun binary: jiti can't find ../dist/babel.cjs
## Summary

`agent-relay fleet serve <node-def-file>` fails immediately on the **published (bun-compiled) CLI**:

```
ResolveMessage: Cannot find module '../dist/babel.cjs' from '/$bunfs/root/agent-relay-darwin-arm64'
```

`loadNodeDefinition` (`packages/cli/src/cli/commands/fleet.ts:123`) uses `jiti` to import the node-def module:
```ts
const jiti = createJiti(pathToFileURL(process.cwd()).href, { interopDefault: true });
const loaded = await jiti.import(absolutePath, { default: true });
```
jiti tries to load its transform backend (`../dist/babel.cjs`), which **isn't bundled into the bun single-file binary** (`/$bunfs/root/…`), so any `fleet serve <file>` (or `@agent-relay/factory/node`) blows up before the broker starts — even when the target is plain compiled JS.

## Repro
```
agent-relay fleet serve ./some.node.js --workspace rk_live_… --base-url https://gateway.relaycast.dev
# → Cannot find module '../dist/babel.cjs' from '/$bunfs/root/agent-relay-darwin-arm64'
```
(agent-relay 8.9.0, darwin-arm64.)

## Workaround
Running the **node-based source CLI** works (jiti + `node_modules/jiti/dist/babel.cjs` resolve):
```
node packages/cli/dist/cli/index.js fleet serve <file> --workspace … --base-url …
# → "Fleet node \"…\" registered with 3 capabilities."
```

## Fix options
- Bundle `jiti`'s `babel.cjs` into the bun binary (or configure jiti's `transform`/`fsCache` so it resolves inside `/$bunfs`), or
- Load node-def files without jiti for the compiled-JS case (plain dynamic `import()`), falling back to jiti only for TS sources.

## Impact
Blocks `agent-relay fleet serve <node-def>` (incl. the documented `@agent-relay/factory/node`) from the published binary — the only way to serve a node-def today is the node source CLI. Surfaced bringing up a factory fleet node.

khaliqgant:
## Summary
- load JavaScript fleet node definition files with native dynamic import instead of constructing jiti
- lazy-load jiti only for TypeScript-like node definition files
- validate loaded fleet node definitions with the local marker check so the Bun-compiled CLI avoids another runtime SDK import failure
- add regression coverage and changelog entry

Closes #1169.

## Validation
- npx vitest run packages/cli/src/cli/commands/fleet.test.ts
- npm run build:sdk
- npm --prefix packages/cli run build
- npm --prefix packages/cli run lint (passes with existing warnings)
- bun build --compile --minify --target=bun-darwin-arm64 --external=better-sqlite3 --external=cpu-features --external=node-pty --define=process.env.AGENT_RELAY_VERSION="8.9.0-test" ./packages/cli/dist/cli/index.js --outfile /tmp/agent-relay-fleet-1169
- /tmp/agent-relay-fleet-1169 fleet serve <temp-js-node-def> --enrollment-url https://agentrelay.com/api/v1/fleet/register -> expected: --enrollment-url requires --enrollment-token.

<a href="https://cubic.dev/pr/AgentWorkforce/relay/pull/1170?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


khaliqgant:
Addressed automated review feedback:

- Added marker-aware default unwrapping for CommonJS compiled `exports.default = defineNode(...)` output.
- Added a Node-only fallback to `jiti` for native import SyntaxError cases, preserving ESM-syntax `.js` node files in CommonJS projects without reintroducing `jiti` into the Bun runtime path.
- Added regression coverage for both cases.

Validation after changes:
- `npx vitest run packages/cli/src/cli/commands/fleet.test.ts`
- `npm --prefix packages/cli run build`
- `npm --prefix packages/cli run lint` (existing warnings only)
- Bun `--compile --minify` smoke for ESM `.js` and CommonJS `exports.default` node definitions.

PR #1212, opened 2026-06-29
fix(cli): make integration subscribe idempotent and per-channel

khaliqgant:
## Why

`agent-relay integration subscribe` was failing on retry with an opaque error:

```
Failed query: insert into "webhooks" (...) values (...)
```

Root cause: the `webhooks` table has a `UNIQUE (workspace_id, name)` index, and every inbound webhook was named just `relayfile:<provider>` — no channel. That means **one Slack (etc.) webhook per workspace, ever**. Combined with a best-effort rollback that swallowed its own failures (`.catch(() => undefined)`), a single partial run left an orphaned `relayfile:slack` row that permanently blocked every subsequent `subscribe` — even to a different channel.

## What changed

- **Per-channel webhook names** — `relayfile:<provider>:<channel>` instead of `relayfile:<provider>`, so multiple channels of the same provider no longer collide on the unique index.
- **Idempotent re-subscribe** — re-running tears down any existing binding for the same `provider+resource` (webhook + subscription + relayfile unbind) and clears webhooks orphaned by an earlier partial run, then recreates cleanly.
- **Visible rollback** — failed cleanup now warns via `error()` instead of being silently swallowed (the silent swallow is what created the orphans).
- **Canonical channel id** — `targetChannel` strips a leading `#` (as well as `@`), so the webhook name, subscription filter, relayfile bind, and writeback-secret lookup all agree on `general` rather than mixing `#general`/`general`.

## Tests

New `integration-subscribe.test.ts` covers: per-channel naming, orphan cleanup (and not touching another channel's webhook), binding replacement on re-subscribe, and warn-on-failed-rollback. Updated the one existing assertion in `relaycast-groups.test.ts` for the new name. Full CLI suite + typecheck pass.

<a href="https://cubic.dev/pr/AgentWorkforce/relay/pull/1212?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


khaliqgant:
Addressed all three review findings in `f21e980` by reworking the webhook identity to be **per `(provider, resource)`** rather than per `(provider, channel)`:

**P1 (codex — create-first):** Re-subscribe now creates the new webhook + subscription + relayfile bind, and only *after* that succeeds does it retire the prior one (`retireSupersededWebhooks`). A transient failure rolls back just the newly-created state and leaves the existing binding fully intact. A per-attempt nonce in the webhook name (`relayfile:<provider>:<slug>-<hash>:<nonce>`) means the replacement can be created alongside the old one without tripping the unique `(workspace, name)` index.

**P2 (codex — don't delete other resources' webhooks):** The name is now resource-hashed, so two resources routed to the same relay channel get distinct webhooks. Cleanup only removes (a) the prior binding's webhook by id, (b) orphans sharing *this resource's* name prefix, and (c) a legacy un-scoped `relayfile:<provider>` webhook **only when no active binding references it**. A new test asserts a second resource's webhook on the same channel survives.

**gemini (HIGH — legacy channel comparison):** The fragile `hook.channel === channel` compare is gone entirely. The prior webhook is retired by its exact id from the binding record, which handles legacy `relayfile:<provider>` names regardless of `#`/no-`#` sigil.

Also fixed a latent mismatch: `listBindings` now maps relayfile's `pathGlob` JSON field to `resource`, so both the new replace path and the existing `unsubscribe` matcher actually match.

New tests cover per-(provider,resource) naming, create-first ordering, P1 (prior binding preserved on failure), P2 (cross-resource webhook preserved), legacy-orphan retirement, and warn-on-failed-cleanup. Full CLI suite (163) + typecheck + prettier green.

khaliqgant:
Second review round addressed in `a71342a`:

- **CodeRabbit @493 (don't continue blind on read failure):** `findExistingBinding` no longer swallows `listBindings` errors. It runs before any mutation, so it now fails fast — a transient read error aborts the subscribe instead of being mistaken for "no prior binding" (which would create a duplicate). An empty/absent store still resolves to `[]`.
- **CodeRabbit @545 (skip webhooks referenced by an active binding):** the cleanup sweep now skips any `webhookId` present in `activeWebhookIds`, so a concurrent re-subscribe that upserted a newer same-prefix binding can't have its live webhook deleted out from under it. `isUnboundLegacy` is simplified accordingly.

New tests: an active-binding webhook is preserved even at its own resource prefix, and fail-fast aborts before `createInbound`. The mock relayfile `bind`/`listBindings` is now stateful to mirror real upsert semantics.

For completeness — the earlier gemini **medium** suggestion (update the test to a `#general` legacy channel) is now obsolete: the redesign matches/retires the prior webhook by its binding id rather than by channel, so there's no `hook.channel === channel` comparison left to exercise.

Full CLI suite (165) + typecheck + prettier green.

PR #1216, opened 2026-06-30
fix(broker): workspace-scope auto node id so `up` injection works

khaliqgant:
## Problem

`agent-relay up` spawns agents that **silently never receive injected messages**. Broker 9.1.x (#1201, node-only delivery) injects solely over the node-control `/v1/node/ws` channel, which needs a node token from `create_node`. The auto-derived node id was `sha256(machine_seed, cwd)` — **workspace-independent** — but the engine scopes nodes **globally**.

So when a project directory is re-pointed at a new workspace (and `up` mints a fresh workspace on each run), the broker reuses a node id **already owned by the old workspace**. `create_node`'s INSERT hits a unique-constraint violation, the engine returns it as a **misclassified HTTP 500**, the retry wrapper collapses it to `Max retries exceeded`, no token is minted, and every spawned agent is undeliverable — while `/health` reports `relaycastConnected:true`.

Regressed in #1201, which removed the firehose delivery path **and** introduced the cwd-only node id in the same commit.

## Fix

**Root cause**
- `derive_node_id` now folds `workspace_id` into the hash → each `(machine, cwd, workspace)` gets a distinct node, no cross-workspace collision. The pinned `RELAY_NODE_TOKEN` (operator-enrolled / fleet) path is **unchanged**.

**Hardening (so this can never be silent again)**
- `create_node` mint failures log the real HTTP status + body instead of `Max retries exceeded`; non-retryable `4xx` are no longer retried.
- `agent-relay up` refuses to auto-spawn when node delivery is down (exit 1 + guidance) instead of producing a dead session.
- broker `/health` + `/api/status` expose `nodeConnected`/`nodeDelivery`; `status` and `doctor` report it.

## Tests
- `cargo test -p agent-relay-broker --lib` → **794 passed / 0 failed**. New regression test: same `(seed, cwd)` yields **distinct** node ids per workspace, stable within one.
- CLI: vitest (60) + `tsc --noEmit` + `build:cli` green. New coverage for the `up` fail-loud path.

## Verification status
- ✅ **Root fix proven:** clean-cwd run against the freshly-built release broker minted a **workspace-scoped node token with no 500 collision**; `/health` reported `nodeDelivery.connected:true`.
- ✅ **401-churn is NOT this change:** a node-control `/v1/node/ws` 401 → re-mint churn was observed after connect. An A/B (our build vs published 9.1.7, identical clean cwd + durable workspace key) reproduced it **identically on both**, so it is pre-existing / engine-side, not a regression here. Tracked as a follow-up below.
- ⏳ **In progress (not a merge blocker):** a full DM-injection e2e between two `up`-spawned agents; local providers are broken (Codex v0.121 model rejection; Claude MCP-auth), being retried with a minimal echo agent. Evidence will be added as a comment.

## Out-of-scope follow-ups (not fixed here)
- **Engine/cloud repo:** `create_node` should return `409` for a node-id conflict (and ideally be idempotent for the same owner) instead of a raw `500`; and the `/v1/node/ws` 401-churn (above) needs root-causing engine-side.
- `agent-relay up` mints a **fresh workspace on every run** (`source=fresh`) — worth revisiting.
- The Fix-2 broker-local `create_node` helper reintroduces a hardcoded `RELAYCAST_DEFAULT_BASE_URL` fallback; consider routing through the SDK base-URL resolution.

<a href="https://cubic.dev/pr/AgentWorkforce/relay/pull/1216?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


khaliqgant:
## ✅ End-to-end injection proof (branch broker)

DM injection between two `up`-spawned agents **works** on this branch, captured during a node-connected window:

- Broker: `target/release/agent-relay-broker`, port 5099, ws `197733757872058368`
- Health before spawn/send: `nodeConnected:true, nodeDelivery:{tokenPresent:true, connected:true}, wsConnections:1`
- Minted node (workspace-scoped): `node_8f4ceae42e4c3a106445c07788ab0912`
- Sender `EchoProofSender` → Receiver `EchoReceiverProof` (trivial echo CLI)
- DM `197761027675701248`, payload `ECHO_INJECTION_PROOF_BRANCH_20260630`
- **Receiver PTY log contained the injected line:** `Relay message from EchoProofSender_… [197733757872058368 / 197761027675701248]: ECHO_INJECTION_PROOF_BRANCH_20260630`
- Broker logged delivery ack `del_197761027675701249`

So: workspace-scoped token mints → node delivery connects → a DM is injected into the spawned agent's PTY. The original symptom is fixed.

## 401-churn scope (follow-up, not this PR)

The `/v1/node/ws` 401 → re-mint churn is **not** limited to `up`-path nodes — the active pear/fleet broker log shows the same cycle for the lead/worker fleet node `node_641bf7…` in the same workspace. It's workspace-wide / engine-side and should be root-caused separately (the create_node `500`-should-be-`409` item is likely related).

khaliqgant:
Engine-side / up-path follow-ups split out into #1217 (409-not-500 for node conflicts, /v1/node/ws 401-churn root cause, `up` fresh-workspace-per-run) with the A/B reproduction evidence. None block this PR.

khaliqgant:
## Review feedback addressed (2141d7bb6)

Thanks @gemini @coderabbitai @codex — all five items fixed:

- **[High] `mint_node_token` retry bypass** (`node_control.rs`): transient `send()`/`text()` failures are now caught inside the loop (log + backoff + `continue`) instead of early-returning via `?`; only the final attempt's error propagates. Added a unit test for transient-send-then-succeed.
- **[High] `/health` can hang** (`listen_api.rs`): status fetch now uses `try_send` (degrades gracefully on a full channel); the reply is bounded by the existing 100ms timeout.
- **[Med] `up` status-poll crash** (`broker-lifecycle.ts`): `waitForNodeDelivery` catches transient `getStatus()` errors and keeps polling to the deadline. Added a CLI test.
- **[Med] stale `tokenPresent`** (`fleet.rs`): `FleetControlEvent::Connected` now sets `node_delivery_token_present = true`, so `/health` is correct after a later re-mint.
- **[Low] regex recompile** (`node_control.rs`): `redact_relaycast_secrets` compiles its regexes once via `OnceLock`.

Gates: `cargo test -p agent-relay-broker --lib` → 795 passed / 0 failed; CLI vitest (16) + `tsc --noEmit` green.

PR #1233, opened 2026-07-06
feat(cli): reflex syncs to relayhistory-cloud in-process via ai-hist SDK

khaliqgant:
## Why

`agent-relay reflex on` wrote a flag and did a one-time login but nothing ever pushed history — the flag was read by nothing at runtime.

## What

The long-running `agent-relay up` host now periodically pushes new local history to relayhistory-cloud **in-process**, gated on the reflex flag. It does this via the `ai-hist` SDK's `pushToCloud`, which **drives the real `ai-hist push` Rust binary** (relayhistory#38) — no CLI shell-out to schedule launchd/cron, no TypeScript re-port of the push logic.

- **`@agent-relay/config` → `reflex-config.ts`**: single source of truth for `~/.agentworkforce/reflex.json` (`isReflexEnabled`, read/write). `reflex.ts` uses it.
- **`reflex on`** persists the `rth_at_` session where the binary reads it — `$RELAYHISTORY_HOME/auth.json` (default `~/.agentworkforce/relayhistory/auth.json`), Rust snake_case shape — so the spawned `ai-hist push` authenticates. Only prints "History syncs automatically…" when cloud login actually succeeded.
- **`reflex-capture.ts`**: unref'd periodic loop (mirrors the telemetry client) started after the fleet sidecar and stopped (final flush) in `shutdownOnce`. Re-checks `isEnabled()` each tick so `reflex off`/`on` take effect immediately; interval starts only after the initial delay. Loads `ai-hist/cloud` via a lazy dynamic import → silent no-op if the SDK/binary is unavailable or unauthed.

## Dependency / ordering

`ai-hist` is intentionally **not** a hard dependency yet: `ai-hist@0.4.0` (relayhistory#38, which ships `pushToCloud`) isn't published, and pinning an unpublished version broke `npm ci` (ETARGET) across all CI. The capture loop reaches it via the lazy dynamic import and no-ops when absent. **Once 0.4.0 is published, a follow-up adds `ai-hist@^0.4.0` as a dependency** so real installs pick it up (needs the `ai-hist` binary present too, via install.sh / PATH).

## Testing

- `vitest` reflex + reflex-capture ✅ 14 passed
- `tsc` (cli + config) ✅, `eslint` ✅, `prettier --check` ✅


khaliqgant:
Thanks all — addressed the feedback in `1e0…` (latest push).

**CI root cause (all install/test jobs):** `npm ci` failed with `ETARGET — no matching version for ai-hist@^0.4.0`. That version isn't published yet (it ships in AgentWorkforce/relayhistory#38). Since the capture loop already loads `ai-hist/cloud` via a lazy dynamic import and no-ops gracefully when it's absent, I **removed the hard dependency** — it's a runtime-optional peer for now, to be declared as a real dependency once `ai-hist@0.4.0` is published. This realigns package.json with the lockfile and unblocks every job. (Also reverted the now-moot esbuild `external` entry.)

**Addressed on the current (in-process SDK) code:**
- `reflex-capture`: re-check `isEnabled()` every tick so `reflex off`/`on` take effect immediately in a running `agent-relay up` (cubic P1 / CodeRabbit) — + test.
- `reflex-capture`: start the interval only after the initial delay so the first push can't precede `initialDelayMs` when `intervalMs` is smaller (cubic P2) — + test.
- `reflex on`: only print "History syncs automatically…" when cloud login actually succeeded (CodeRabbit major / cubic P2) — + negative assertions.

**Not applicable:** several comments (execFileAsync timeout, `cloudSyncInstalled` placeholder, Linux cron probe, `uninstallCloudSync` failure-branch test) target the earlier **shell-out** revision, which was fully replaced by the in-process SDK approach — that code no longer exists.

khaliqgant:
Went through all the review threads. This PR evolved through several approaches (shell-out → ai-hist SDK → bundled binary spawn → **in-process napi addon**), so most inline comments target code from earlier commits that no longer exists.

**Addressed (current code):**
- `loadNative()` swallowed every dynamic-import error, so an installed-but-broken addon (ABI mismatch, missing lib, init failure) looked the same as "not installed" and silently no-op'd (cubic P2 on the merge commit). Now only `ERR_MODULE_NOT_FOUND`/`MODULE_NOT_FOUND` → no-op; any other error is rethrown and logged as `[reflex] cloud sync failed: …`.

**Already handled in the current code (no change needed):**
- Re-check `isEnabled()` every tick so `reflex off` takes effect immediately — present (`reflex-capture.ts` `tick()`).
- First push can't fire before `initialDelayMs` regardless of interval — the interval starts inside the kickoff timer.

**Moot — target deleted code** (the shell-out / bundled-binary-spawn eras):
- `defaultCloudSyncInstalled` / Linux cron status / `'unknown'` handling, `defaultInstallCloudSync`/`defaultUninstallCloudSync` + ENOENT, `--install-service` flag gating, `ai-hist-path.ts` resolution + `spawn` timeout, the `ai-hist` (SDK) dependency not-published note, changelog for "scheduling services".

The final design is: `reflex on` flips the flag + logs in; the `agent-relay up` host calls `ai-hist-native`'s `syncAndPush()` **in-process** (no subprocess), gated on the flag. Depends on `ai-hist-native@^0.4.1` (published).

Issue #1247, opened 2026-07-12
Harness driver / PTY attach (drive & view) correctness bugs — consolidated audit


willwashburn:
`relay agent drive` and `relay agent watch` (= `agent attach --mode view`) misbehave. A code audit across `packages/cli` (attach clients), `packages/harness-driver`, and `crates/relay-pty` + `crates/broker` found the following verified defects. Every claim below was checked against the actual code paths.

## Structural root cause: snapshot ↔ live stream cannot be synchronized

`worker_stream` events carry **no sequence number** and are excluded from the replay buffer (`crates/broker/src/listen_api.rs:2520`), so an attaching client cannot correlate its HTTP snapshot with a stream position:

- **Lost output:** everything emitted between the snapshot response and the WS subscription is unrecoverable. A gap that cuts a repaint or escape sequence leaves the screen garbled until the next full redraw. (`packages/cli/src/cli/lib/attach.ts:61-66` claims the window is "≤10ms" — not true for drive.)
- **Duplicated output:** the snapshot grid is advanced by the PTY reader *before* chunks are queued for broadcast (`crates/relay-pty/src/pty.rs:328-332`), and the 16ms coalescing buffer (`crates/broker/src/pty_worker.rs:708-719`) is not flushed at snapshot time — bytes can appear in the snapshot *and* again on the stream.
- **Drive widens the gap and injects a repaint into it** (`packages/cli/src/cli/lib/attach-drive.ts:740-766`): order is snapshot → `GET /pending` → **resize agent PTY** (SIGWINCH → immediate TUI repaint) → only then WS connect (`:619`). The user sees a stale snapshot at the old geometry until the agent next outputs. Predictive echo is seeded from the stale snapshot bytes, so early predictions render at wrong positions.
- **Watch loses output forever on lag:** one shared 512-slot broadcast channel for all events of all workers (`crates/broker/src/runtime/init.rs:383`); on `RecvError::Lagged` only durable events are backfilled; clients discard `replay_gap` frames without re-snapshotting (`attach-view.ts:189-202`, `attach-drive.ts:337-361`).

**Fix direction:** per-worker monotonic offset on `worker_stream`; include offset in snapshot response; clients subscribe first, snapshot second, drop buffered chunks ≤ snapshot offset; move drive's resize after subscription; re-snapshot on `replay_gap`.

## Drive-specific bugs

[...]

PR #1253, opened 2026-07-12
fix(broker,relay-pty): resize ownership, stateful ANSI scanning, watchdog hardening (#1247)

willwashburn:
## Summary

Workstream F of #1247 (6/6, stacked on #1252 — merge last). Closes out the final three audit items:

- **Single-resizer policy:** resize requests carry an optional client `session_id`; the first session to resize a worker claims it, non-owners are acknowledged but not applied, and ownership is released on detach (plus a 5-minute staleness net for crashed clients). Concurrent drive clients no longer fight over the shared PTY size. Fully backward compatible — resizes without a `session_id` behave exactly as before, and the broker→worker frame is unchanged.
- **Stateful ANSI scanning:** new `AnsiStripper` in `relay-pty` (ESC/CSI/OSC/DCS state machine mirroring the CLI's `AnsiBoundaryScanner`, 4KB runaway cap) holds incomplete escape tails between chunks. The worker's readiness/prompt detection and the auto-suggestion guard now see whole sequences, so an escape split across 4KB PTY reads no longer leaks fragments like `3;5H` into detection text or misses a split `\x1b[7m` ghost-text marker.
- **Watchdog hardening:** the no-PID liveness heuristic's silent threshold is configurable and defaults to ~5 minutes (was ~30s), and the counter resets on successful writes as well as output — a silently-thinking child is no longer declared dead mid-task. Genuine exits are still caught promptly via PTY reader EOF.

## Test Plan

- [x] Tests added/updated (resize owner accept/reject/release/stale-takeover, split-CSI/OSC/SGR-marker stripper tests, watchdog long-silence/real-exit/write-reset; `cargo test -p relay-pty` 195 passed, `cargo test -p agent-relay-broker --lib` 686 passed, clippy clean, vitest harness-driver+cli 649 passed, tsc clean)

## Screenshots

N/A

---
_Generated by [Claude Code](https://claude.ai/code/session_01FMEduowX291oCbv2q9RVDT)_

<a href="https://cubic.dev/pr/AgentWorkforce/relay/pull/1253?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


willwashburn:
Addressed the remaining open review threads in c0eeb20 (Rust: `cargo test -p relay-pty` 207 + `-p agent-relay-broker --lib` 737 pass, clippy clean; TS: `npm run typecheck` clean, attach vitest 85 pass):

**`crates/relay-pty/src/ansi.rs`**
- 🔴 Cap `ESC[<n>C` repetition in the one-shot `strip_ansi` — the count is clamped to `ANSI_TAIL_MAX` before the space-push loop, so an oversized parameter can't drive unbounded iteration/`String` growth (default `1` and small counts unchanged).
- Restart CSI parsing on a stray `ESC` — the param loop now breaks on `ESC` without consuming it so the outer loop restarts there (mirrors `AnsiStripper::scan`); a following sequence is no longer swallowed as CSI params. Added `strip_ansi_caps_cursor_forward_repetition` + `strip_ansi_restarts_csi_on_stray_esc`.

**`crates/broker/src/pty_worker.rs`**
- Auto-suggestion detection now runs through a dedicated `suggestion_stripper: AnsiStripper` feeding `feed_raw(&text)` into `update_auto_suggestion`, mirroring `wrap.rs`, so a `\x1b[7m` ghost-text marker split across PTY reads is stitched before detection in the `drive`/`passthrough` path.

**`packages/cli/src/cli/lib/attach-drive.ts` + `attach-passthrough.ts`**
- The resize-ownership release is captured as `const releasePromise = (async () => {…})()` and bundled with `restoreInboundDeliveryModeOnDetach(...)` in `Promise.allSettled([...]).finally(() => resolve(code))`, so `resolve(code)`/process exit waits for the release instead of aborting it — the detach-race release no longer loses the race to the restore.

**`crates/relay-pty/src/pty.rs`**
- `enqueue_without_confirm_does_not_reset_no_pid_counter` now drives the real `drain_write_queue` against a `BlockingWriter`, so it exercises the confirmed-write gate rather than asserting on a counter nothing touches — it would fail if the counter reset on receipt instead of on a confirmed flush.

**`crates/broker/src/runtime/{event_loop,api}.rs`**
- Extracted the `ResizePty` state transition into `plan_resize` (`Reject`/`Refresh`/`Apply`, bumps `last_seen` on refresh) and `commit_resize_ownership`; the handler calls these and the resize-owner tests drive the real helpers (keyed claim → legacy resize → keyed re-assert) instead of reimplementing the map mutation. Ownership is still committed only after a successful worker send.

**`CHANGELOG.md`**
- Moved the resize/ANSI/watchdog bullets out of the released `9.2.2` block into `[Unreleased - Patch]`.

---

willwashburn:
Thanks @cubic-dev-ai — all 6 addressed in 787e7ea (relay-pty 207 + broker 737 tests pass, clippy/fmt clean, `npm run typecheck` clean, attach vitest 85 pass):

1. **`pty_worker.rs:970` (P2, ghost-text pair split across reads)** — Correct: `is_auto_suggestion` needs both `\x1b[7m` and `\x1b[27m\x1b[2m` in one string, and `feed_raw` only stitches a single split escape. Added a bounded raw lookbehind (`SUGGESTION_LOOKBEHIND_MAX = 512`) that accumulates recent scanned output so a marker *pair* landing in separate PTY reads is matched whole. Mirrored the same change in `wrap.rs` so both paths stay consistent.

2. **`pty.rs:1495` (P3, test race)** — `BlockingWriter` now signals over a channel when it enters `write`, and the test blocks on that signal before asserting, so it observes the genuine picked-up-but-unconfirmed state rather than passing before the drainer dequeues.

3. **`attach-drive.ts:848` (P2, exit can hang ~60s)** — Bounded the teardown: the release+restore `Promise.allSettled` is now raced against `DETACH_CLEANUP_DEADLINE_MS` (2s), so `resolve(code)` fires once they settle or the deadline elapses. The broker's idle-takeover net still frees any ownership left behind.

4. **`event_loop.rs:523` (P3, `Instant` underflow panic on fresh-boot CI)** — Both affected tests now anchor `last_seen` at a real `Instant` and offset the injected `now` *forward* (`base + RESIZE_OWNER_STALE + 1s`) instead of subtracting from `Instant`, which panics when the monotonic clock is younger than the offset.

5. **`attach-passthrough.ts:585` (P2, detach hang on stalled resize)** — Same bounded-deadline fix as #3 applied to the passthrough teardown.

6. **`api.rs:1139` (P2, `last_seen` sampled before the awaited send)** — `commit_resize_ownership` is now called with a freshly sampled `Instant::now()` taken *after* the awaited `send_to_worker`, so the stale-owner window is anchored to when the resize actually applied (restores the pre-refactor behavior). `plan_resize` still uses the earlier timestamp for the staleness check, which is correct.

---

willwashburn:
Both addressed in a034ad5 (broker 741 tests pass incl. 4 new, clippy/fmt clean, `npm run typecheck` clean, attach vitest 85 pass):

1. **`pty_worker.rs:988` (P2, lookbehind held visibility too long)** — Correct, and it was the stickiness risk in the accumulating buffer. Replaced it with `PtyAutoState::update_auto_suggestion_windowed(current, prev_tail)`: it uses `prev_tail` (bounded tail of the *previous* chunk only, not accumulated history) **solely** to recognise a marker pair straddling the read boundary — and explicitly skips when the pair is already fully within `prev_tail`, so an already-handled suggestion can't re-arm. The clear/keep decision delegates to the current chunk alone, so a normal output chunk clears `auto_suggestion_visible` promptly regardless of the lookbehind. Mirrored in both `pty_worker.rs` and `wrap.rs`; added `auto_suggestion_tests` covering in-chunk pair, boundary straddle, stale-pair-in-tail clears, and whitespace-leaves-unchanged.

2. **`attach.ts:457` (P2, orphaned JSDoc)** — Moved `DETACH_CLEANUP_DEADLINE_MS` (with its own doc block) *above* `restoreInboundDeliveryModeOnDetach`'s JSDoc, so the restore doc sits immediately before the function again and each symbol keeps its correct association.

---

PR #1264, opened 2026-07-14
fix(broker): bound and retry the Relaycast startup handshake

willwashburn:
## Summary

Fixes the flaky failure in the publish workflow's macOS **standalone lifecycle smoke** (`verify-standalone-macos`), which intermittently failed with:

> `Broker process exited with code null during initial handshake` → `Broker startup readiness line missing or duplicated: expected 1, got 0`

**Root cause (a hang, not a crash):** `connect_relay` performs the mandatory Relaycast handshake via the `relaycast` SDK's workspace bootstrap (`create_workspace` / agent registration). Those bootstrap calls build a **timeout-less** reqwest client, so when the connection to the relay backend stalls, `startup_session_set_with_options` hangs indefinitely. The broker never prints `Broker started.`, and because the CLI's `node up` is still within its 45s handshake budget, the standalone smoke's own `sleep 8 && down --force` teardown then **kills the still-starting broker** — surfacing to the SDK as an opaque `code null` (signal) exit "during initial handshake".

`code null` (killed by signal) rather than a clean error exit (1) or a Rust panic (101) is what pointed at an external reap of a hung process rather than a code fault. This was confirmed flaky: the same run/commit was re-run and passed.

**Fix:** Wrap `startup_session_set_with_options` in a **per-attempt `tokio::time::timeout` with bounded retry + exponential backoff**. A stalled attempt now aborts fast and retries in-process, well within the SDK's 45s startup budget, instead of hanging until something reaps the process. Terminal failures/timeouts surface as clean, contextual errors instead of a silent hang.

- Defaults: **5s per attempt, 4 attempts, 250ms → 2s backoff** (worst case ~22s < 45s SDK budget).
- Overridable via `AGENT_RELAY_HANDSHAKE_TIMEOUT_MS` and `AGENT_RELAY_HANDSHAKE_ATTEMPTS`.
- Startup-debug tracing logs each attempt/timeout/retry for future diagnosis.

Files: `crates/broker/src/runtime/session.rs` (+ `CHANGELOG.md`).

### Follow-ups (not in this PR)
- The upstream defect lives in `relaycast` sdk-rust: `create_workspace`/`lookup_workspace` should build a client with an explicit timeout (the authenticated client already uses 30s). Fixing that needs a crate release + version bump.
- The "standalone" smoke isn't truly standalone — it depends on production Relaycast reachability and mints a throwaway workspace each run. Pointing it at a local `RELAY_BASE_URL` stub would remove the network dependency entirely.

## Test Plan

- [x] Tests added/updated — unit tests for the `AGENT_RELAY_HANDSHAKE_TIMEOUT_MS` / `AGENT_RELAY_HANDSHAKE_ATTEMPTS` env overrides (default/invalid/zero/valid)
- [x] Manual testing completed — `cargo check`, `cargo clippy` (clean), `cargo fmt`, and `cargo test -p agent-relay-broker --lib handshake` (2 passed) on the broker crate

## Screenshots

https://claude.ai/code/session_01WzkvFAUELJgDQgZ3pdoRTP

---
_Generated by [Claude Code](https://claude.ai/code/session_01WzkvFAUELJgDQgZ3pdoRTP)_

<a href="https://cubic.dev/pr/AgentWorkforce/relay/pull/1264?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


willwashburn:
Addressed the review feedback in `719a8d1`:

- **Duplicate Relaycast resources on retry** (cubic): the loop now retries **only on timeout**. A returned error is a definite backend response, so replaying `startup_session_set_with_options` (which re-runs `create_workspace` + agent registration) is no longer done — returned errors surface immediately, matching the pre-retry behavior. The residual is narrow: a timeout retry can still re-create resources only if the backend *both* completed the request *and* failed to answer within the deadline; a full fix for that (create-or-get by name) belongs in the `relaycast` SDK, which currently falls back to a random-suffixed workspace on name conflict.
- **Per-attempt timeout too coarse for multi-workspace startup** (codex + cubic): the deadline is now scaled by the configured workspace count (`RELAY_WORKSPACES_JSON` array length), so a healthy startup that registers several memberships serially isn't cut off by the single-request timeout.
- **`std::env::set_var`/`remove_var` in tests** (CodeRabbit): parsing moved into pure helpers (`parse_handshake_timeout` / `parse_handshake_attempts` / `parse_membership_count`) that take `Option<&str>`; tests exercise those directly and no longer mutate process-global env.

Also moved the CHANGELOG entry into `[Unreleased - Patch]` (a `main` merge had swept it into the released `[10.2.0]` section).

---

willwashburn:
Heads-up on the `worker.rs` change in `7e850cc` (outside this PR's stated scope, but it was blocking this PR's CI):

The **Rust Tests (ubuntu-latest)** check was failing on `worker::tests::codex_arg_fallback_rewrites_effective_model` — **not** caused by this branch (`worker.rs` was byte-identical to `main`; my change is confined to `session.rs`). Root cause is a pre-existing **`ETXTBSY` fork/exec race**: `codex debug models` execs a resolved codex binary, and on Linux a concurrent `fork`/`exec` in another thread can transiently hold a writable fd to a just-written executable, so `execve` fails with "Text file busy". That made model detection return `None` and fall back away from the requested model. It's deterministic-ish on Linux under the parallel test suite (each test writes+execs a fake `codex`) and absent on macOS — matching the ubuntu-only failures.

Fix: retry the spawn on `ExecutableFileBusy` (5 attempts, short backoff, within the existing 15s timeout), matched via the portable `std::io::ErrorKind::ExecutableFileBusy` rather than `libc::ETXTBSY` so the broker's Windows build stays correct. It's real production hardening (the broker spawns subprocesses concurrently), not just a test fix. Verified with 35 consecutive full lib-suite runs, 0 failures.

Happy to split this into its own PR if you'd prefer to keep this one scoped strictly to the handshake change — let me know.

---

PR #1273, opened 2026-07-15
fix(cli,broker): let driven agents receive relay messages (Ctrl+] toggle + flush through interactive hold)

willwashburn:
## Summary

Fixes the two-computer repro where an agent driven on node 2 asked a remote agent for a ping reply, the reply showed up in the observer stream, but was never injected into the driven agent.

**Root cause (local, not cloud routing):** a drive attach flips the worker to `manual_flush`, and the #1249 interactive hold freezes worker-side injection pops — so an agent being driven can never receive a relay message. Worse, `agent-relay node agent message flush` only handed the queue to the PTY worker, whose injections are *also* frozen by the hold, so the messages moved into a second invisible queue until detach.

Changes:

- **`Ctrl+]` in drive** toggles the driven agent between held (`manual_flush`) and live (`auto_inject`) inbound delivery via a revision-guarded compare-and-set; flipping to live drains the parked queue into the PTY. The status line renders the real mode plus a `Ctrl+] deliver` / `Ctrl+] hold` hint, and the detach restore CASes against the session's *last* write so a mid-session toggle can't clobber out-of-band changes.
- **`POST /api/spawned/{name}/flush` injects through the hold:** the broker follows the handoff with a new one-shot `flush_injections` worker frame; the PTY worker grants the queued backlog (and a frozen in-flight injection) a hold exemption so an explicit flush injects immediately instead of waiting for detach. Deliveries arriving *after* the flush keep parking under the hold, and a requeued exempt injection keeps its exemption, so the anti-splice design of #1249 is preserved.
- **Pending counter accuracy:** a partial drain no longer zeroes the drive status line's pending counter; the undrained remainder stays counted.
- `protocol.rs` and `protocol.ts` updated together (snake_case wire tag `flush_injections`).

## Test Plan

- [x] Tests added/updated
  - CLI vitest: `KeybindParser` Ctrl+] cases, status-line hints, toggle CAS flow (including adopt-on-mismatch), partial-drain counter, detach-restore-from-latest-revision (196 tests across attach suites green)
  - Rust: `injection_pop_allowed` hold-exemption cases, `flush_injections` protocol round-trip; full `agent-relay-broker` suite green
  - To verify the original repro: drive an agent, have it message a remote agent, and press `Ctrl+]` (or run `agent-relay node agent message flush <name>` from another terminal) when `pending` increments — the reply should inject into the TUI without detaching.

## Screenshots

N/A (terminal status line change: `[drive kjg-lead | delivery=manual_flush | pending=1 | Ctrl+] deliver | Ctrl+C detach]`)

https://claude.ai/code/session_01F2pHuuKtUtraY6bBD2zLbh

---
_Generated by [Claude Code](https://claude.ai/code/session_01F2pHuuKtUtraY6bBD2zLbh)_

<a href="https://cubic.dev/pr/AgentWorkforce/relay/pull/1273?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


willwashburn:
Merged `main` (resolved the CHANGELOG conflict from the 10.3.0 cut — this PR's pending entries are back under `[Unreleased - Minor]`, and the misplaced `Ctrl+]` bullet was removed from the released 10.3.0 section) and addressed all review findings in 944862e:

- **Exemption leaking into a later hold (codex P2 + cubic P2):** `flush_injections` now grants exemptions only while `interactive_hold` is already active, and every `set_interactive_hold` boundary resets the allowance and clears the in-flight injection's exemption — a pre-drive flush can no longer inject into a subsequently attached drive session.
- **Ctrl+] / Ctrl+C race (cubic P1):** the toggle is tracked as an in-flight promise; `finish()` awaits it before the detach restore and reads `currentMode`/`currentRevision` only after it settles (state is updated even when teardown began mid-request). Regression test: toggle+detach back-to-back asserts the restore CASes against the post-toggle revision.
- **CAS without revision (cubic P2):** the toggle always sends `expectedMode`, adding `expectedRevision` only when the broker reports one — never an unconditional write. Covered by a legacy-broker (no revision) test.
- **Stale drive header note + duplicated trajectory bullets (cubic P3 ×2):** both fixed.

CLI vitest (59), full `agent-relay-broker` cargo suite, typecheck, prettier, and rustfmt all green.

---

PR #1274, opened 2026-07-15
Resolve the right workspace key for CLI commands in a project directory

willwashburn:
## Summary

Fixes the class of bug where `agent-relay fleet nodes` (and other SDK-backed commands) showed the wrong roster because they resolved the **machine-global active workspace key** instead of the workspace the current directory's broker actually joined. Three related changes:

- **`--wk` shorthand for `--workspace-key`** on every SDK-backed command (`fleet nodes`, `workspace`, `integration`, `webhook`, …) and on `up`/`node up`. It's folded into `workspaceKey` by a `preAction` hook so all downstream readers see one canonical option; an explicit `--workspace-key` still wins. (Commander v12 can't alias two long flags to one attribute, hence the fold.)
- **`up` records the workspace it joins.** `agent-relay up` writes the workspace key it joined — passed via `--workspace-key`/`--wk` or auto-minted — into the project data dir (`.agentworkforce/relay/workspace-key.json`, git-excluded, `0o600`, same posture as `connection.json`). SDK-backed commands run in that directory resolve that key **ahead of** the global active workspace, so `fleet nodes`/`node …` reflect the broker's real workspace. An explicit `--workspace-key`/`--wk` or `RELAY_WORKSPACE_KEY`/`RELAY_API_KEY` still overrides it.
  - Precedence: `flag` → `RELAY_WORKSPACE_KEY`/`RELAY_API_KEY` → **project broker key (this CWD)** → global active store.
- **`fleet nodes` warns when the key was inferred from that project record.** The note goes to **stderr** (stdout JSON stays clean), so a stale broker workspace is explainable rather than silent. `resolveWorkspaceKeyWithSource` was added to report the resolution source (`flag | env | project | store`).

Resolution stays synchronous — no async refactor across the command surface — and keeps working after the broker stops.

## Test Plan

- [x] Tests added/updated
  - `--wk` fold on SDK commands (`fleet.test.ts`) and on `up`/`node up` (`core.test.ts`), incl. explicit-flag-wins.
  - Project-key store round-trip, malformed/blank handling, `0o600` (`project-workspace-key.test.ts`).
  - `resolveWorkspaceKey`/`resolveWorkspaceKeyWithSource` precedence + source reporting (`sdk-client.test.ts`).
  - `fleet nodes` warn-on-project-source / no-warn-on-explicit-key (`fleet.test.ts`).
- [ ] Manual testing completed — logic is unit-covered; a full run against a live Relaycast workspace wasn't possible from this environment.

Note: one unrelated suite (`node-definition-loader.test.ts`) fails identically on `main` in a fresh clone (jiti importing an example needs the workspace packages built); it's not touched here.

## Screenshots

N/A — CLI flag + stderr note.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FACC44XmwPpNAWmAfPeXZD)_

<a href="https://cubic.dev/pr/AgentWorkforce/relay/pull/1274?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


willwashburn:
Addressed the automated review findings in c0de5f9:

- **`--state-dir` misses the recorded key (Codex/cubic, P2)** — valid. `runUpCommand` mutated `paths.dataDir` for `--state-dir` before persisting, so the key landed under the override dir while SDK commands (which don't accept `--state-dir`) read only the default project dir. Now the default project data dir is captured *before* the override and the key is always persisted there.
- **Symlink redirect on write (cubic, P1)** — valid. The key is now written to an exclusively-created temp file (`O_CREAT|O_EXCL`, so a pre-planted symlink at the temp path can't redirect it) and `rename`d over the destination. `rename` never follows a destination symlink, so the owner-only guarantee holds.
- **Truncation race on in-place write (cubic, P2)** — valid, fixed by the same atomic temp-file + `rename` (atomic on POSIX): a concurrent reader now sees either the old or the new complete file, never a truncated one.
- **Weak "does not warn" test (cubic, P3)** — valid. That test now sets up a real project-key context so it proves `--wk` suppresses the advisory, rather than the trivial no-key case. Added a `core.test` asserting `up --state-dir` records the key in the default project dir, not the state dir.

CodeRabbit's only remaining flag is a docstring-coverage *warning* (non-blocking). All CLI tests and lint/format pass locally.

---

PR #1281, opened 2026-07-16
Add changelog fragment workflow to prevent merge conflicts

willwashburn:
## Summary

Fixes the recurring changelog pain: when several PRs are open at once, one merges and gets a release, and the next PR's changelog entry lands under the newly-released version (or leaves a conflict marker) instead of under `[Unreleased]`. Root cause is that every PR edits the same few lines of one file, so git's 3-way merge anchors additions to stale context.

This moves pending user-visible changes to **one file per change** under `changelog.d/`. Because each change lands in its own new file, two PRs never touch the same lines — so a changelog entry can't merge-conflict or be inserted under the wrong version. At release time the fragments compile into a dated version section and are deleted.

### What's included

- **`scripts/changelog.mjs`** — zero-dependency CLI:
  - `add` — scaffold a fragment (`--type`, `--level`, text).
  - `preview` — render what `[Unreleased]` will look like from the fragments.
  - `check` — validate every fragment and reject leftover conflict markers in `CHANGELOG.md` (CI entry point).
  - `release` — compile fragments into `## [x.y.z] - DATE`, delete them, reset `[Unreleased]` to empty. Version defaults to bumping `package.json` by the **highest** fragment level (`patch < minor < major`); also folds in any inline `[Unreleased]` bullets so the transition is safe.
- **`package.json`** — `changelog:{add,preview,check,release}` scripts.
- **`.github/workflows/changelog.yml`** — runs `changelog check` on every PR.
- **`.gitattributes`** — `CHANGELOG.md merge=union` as a backstop against markers if the file is ever edited directly.
- **`changelog.d/README.md`** + **`AGENTS.md`** — document the format and replace the old "manually bump `[Unreleased - Patch]`" convention (the level is now derived from fragments).
- Migrated the existing inline `[Unreleased]` entry into a fragment and reset the heading to an empty `[Unreleased]` (dogfood).

Fragment format:

```md
---
type: Fixed
level: patch
---

`agent-relay foo` now does X instead of Y.
```

## Test Plan

- [x] Manual testing completed
  - `changelog:add` generates prettier-clean fragments (verified against `prettier --check`).
  - `changelog:preview` renders correct section ordering and derives `[Unreleased - Major]` when a major fragment is present.
  - `changelog:release --version 10.5.1` produces an empty `[Unreleased]` + a new `## [10.5.1]` section with the compiled entry, leaving prior versions intact.
  - `changelog:check` exits non-zero on an invalid `type`/`level` and on a conflict marker in `CHANGELOG.md`; exits 0 on the clean tree.
- [ ] Tests added/updated — no unit tests added; the tool is exercised by the new CI `changelog check` job.

## Screenshots

https://claude.ai/code/session_0157j54yBsE4a3KeDdFJ1TRj

---
_Generated by [Claude Code](https://claude.ai/code/session_0157j54yBsE4a3KeDdFJ1TRj)_

<a href="https://cubic.dev/pr/AgentWorkforce/relay/pull/1281?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


willwashburn:
Addressed the review feedback and resolved the merge conflict.

**Publish workflow (P1 — flagged by codex + cubic):** valid and now fixed. `publish.yml` didn't know about fragments, so a release would have skipped curated notes and left the files behind. Added a **Collect changelog fragments** step (`node scripts/changelog.mjs collect`, new subcommand) that folds `changelog.d/` fragments into the inline `[Unreleased]` block and deletes them *before* the existing changelog generator runs — so the generator reads them as curated entries with no other changes to its logic. It's gated to stable releases (prereleases skip changelog gen, so we don't want to consume fragments there), and the release commit now stages `changelog.d/` so the deletions land. No-op when there are no fragments.

**Other comments:**
- Escaped the backticks in the `changelog:add` example in `AGENTS.md` (P2).
- Listed `type` values in render order in `changelog.d/README.md` (P3).
- Tightened the observer-token fragment to an impact-first one-liner (P3).

**Merge conflict:** the `merge=union` attribute on `CHANGELOG.md` auto-merged cleanly, but that stacked two `[Unreleased]` headings and duplicated the entries main added while this PR was open. Migrated those pending inline entries into fragments and reset `[Unreleased]` to empty, which is the intended steady state.

---

PR #1342, opened 2026-07-19
Fix silent worker spawn and strict identity failures

miyaontherelay:
## Summary

Fix three related silent-agent-failure paths:

- Gate broker worker spawn success on a `worker_ready` handshake; an immediately exiting CLI now fails the action and is removed instead of being reported as `spawned: true`.
- Emit a structured warning before the existing `dead_letter_added` broker event so persisted undeliverable messages are visible in operational logs.
- For `RELAY_STRICT_AGENT_NAME=1`, use atomic plain registration rather than token-rotating registration, so a second session gets an explicit conflict instead of invalidating the first session.

The commits are intentionally separable: readiness, dead-letter observability, strict-identity registration, then changelog.

### Bug 1 — false spawn success

Pre-fix path on `main`: `crates/broker/src/worker.rs:798` launched the process, `:834` inserted it into the registry, and `:852` returned `Ok(spec)` without a readiness acknowledgement. `crates/broker/src/runtime/fleet.rs:447-454` then emitted success solely because the registry contained the name.

- Before, the live `nightcto-sf` reproduction completed: `{"output":{"name":"ncto-probe","spawned":true},"status":"completed"}`.
- After, a rebuilt release broker node ran with `codex` deliberately replaced by a command that exits 64 before readiness. The spawned action reached the terminal result `{"status":"failed","error":"spawn_failed"}` (the relay also delivered `spawn_failed`). No `spawned:true` completion was emitted.

The regression test executes `false` after receiving `init_worker`; before the readiness gate it failed because spawn returned `Ok`, and it now rejects and leaves no registered worker.

### Bug 3 — strict identity token ping-pong

Before this change, strict-name rebind/bootstrap unconditionally called `registerOrRotate` at `packages/cli/src/cli/agent-relay-mcp.ts:360` and `:916-921`. The SDK describes that operation as adopting an existing identity and invalidating the old token at `packages/sdk/src/facade.ts:189-193`. Two sessions with the same strict name therefore alternated invalidating one another. This was a live operational issue: this workspace's `lead` identity was invalidated four times in one session by another live session sharing the name.

Strict rebind/bootstrap now calls atomic `agents.register`; an existing live identity produces a server conflict and leaves its token usable. Non-strict callers retain rotation behavior.

### Bug 2 — silent dead letters

`dead_letter_pending_delivery` now writes a structured warning containing `worker`, `delivery_id`, `event_id`, `attempts`, and `reason` immediately before the existing `dead_letter_added` event.

## Verification

- `cargo fmt --check`
- `cargo test -p agent-relay-broker` — 783 passed, 4 ignored
- `cargo build --release --bin agent-relay-broker`
- `npm run typecheck`
- `npx vitest run packages/cli/src/cli/agent-relay-mcp.test.ts packages/cli/src/cli/agent-relay-mcp.startup.test.ts` — 28 passed
- Live rebuilt-binary node run described above (the key runtime proof, not just a unit test)

## Risk / review focus

The 30-second outer `worker_ready` deadline is the riskiest change and deserves scrutiny. It deliberately fails closed: every `WorkerRegistry::spawn` caller, including fleet placement, now waits for readiness. PTY workers emit readiness after their existing boot/prompt detection, with their existing 25-second fallback; the outer five seconds are transport headroom. Thus a cold `npx` fetch or first-run binary can still reach the PTY fallback around 25 seconds, but a worker that cannot emit the broker frame by 30 seconds fails instead of becoming a phantom agent.

Headless and app-server workers already emit `worker_ready` immediately after receiving `init_worker`, so they take the same handshake path without waiting for PTY prompt detection. This is a liveness handshake, not a stronger proof that an external provider is fully authenticated; reviewers should assess whether the PTY fallback and 30-second cutoff are right for unusually slow cold-cache CLIs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - `agent-relay node up --background` now preserves persisted Cloud enrollment credentials and node identity through detached startup, and fails instead of reporting healthy when the enrolled node cannot connect.
+- `agent-relay-broker` now waits for a worker readiness handshake before completing a fleet spawn action, so a CLI that exits during startup produces `action.failed` instead of a false `spawned: true` result.
+- Strict-name MCP registration now fails atomically when another session already owns the identity, rather than rotating that session's token and causing both sessions to invalidate each other.
+- Broker dead-lettering now emits a structured warning with the worker, delivery, attempt count, and reason, making persisted undeliverable messages visible to operators.
 
 ## [10.6.1] - 2026-07-16
 
--- a/crates/broker/src/runtime/dead_letter.rs
+++ b/crates/broker/src/runtime/dead_letter.rs
@@ -230,6 +230,15 @@ pub(crate) async fn dead_letter_pending_delivery(
         attempts: entry.attempts,
         reason: entry.reason.clone(),
     };
+    tracing::warn!(
+        target = "agent_relay::broker",
+        worker = %entry.worker_name,
+        delivery_id = %entry.delivery.delivery_id,
+        event_id = %entry.delivery.event_id,
+        attempts = entry.attempts,
+        reason = %entry.reason,
+        "delivery moved to dead-letter queue"
+    );
     dead_letters.push(entry);
     let _ = send_broker_event(sdk_out_tx, event).await;
 }
--- a/crates/broker/src/worker.rs
+++ b/crates/broker/src/worker.rs
@@ -23,7 +23,7 @@ use serde_json::{json, Value};
 use tokio::{
     io::{AsyncBufReadExt, AsyncWriteExt, BufReader},
     process::{Child, ChildStdin, Command},
-    sync::mpsc,
+    sync::{mpsc, oneshot},
     time::timeout,
 };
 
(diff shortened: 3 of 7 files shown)


miyaontherelay:
Additional production evidence for the strict-identity registration fix: at approximately 09:47 UTC on 2026-07-19, a live NightCTO lead session failed mid-`post_message` with `agent_token_invalid: The selected Relaycast agent token is no longer valid. The stale token was cleared from this MCP session.` The session recovered only by re-registering under a distinct name and retrying. This is the same token-rotation/ping-pong failure addressed here; it also previously stranded `ncto-nango-slack` for roughly 2.5 hours while a sent response was effectively unreachable. No code changes accompany this comment.

khaliqgant:
**CI diagnosis, from outside this repo — the failing check is not what the review-risk note predicted.**

Posting because the PR notes flagged "the 30s readiness deadline / cold-cache PTY behavior" as the review risk, and that turns out **not** to be what the E2E is failing on. Anyone reaching for a wider deadline would get a green test that fixed nothing.

**Actual failing assertions** (run 29679806461, step 10 `Run two-node fleet E2E`):
```
Error: waitFor timed out (node-a load exceeds node-b); last=null
AssertionError: expected 404 to be less than 300     ×3
```
No `worker_ready`, no readiness timeout, no spawn failure. It fails on **load distribution across two nodes** — waiting for node-a's load to exceed node-b's, value never appeared — plus three 404s.

**The same job passes on main:**
```
main — workflow "Fleet E2E" run 29683471018 @ b26789fbf, 2026-07-19T10:30:51Z
  job "Two-node fleet matrix" → SUCCESS
```
Verified as the same job, not a similarly-named one.

**What that establishes, and what it doesn't.** Green on main makes "this PR introduced it" the leading explanation and moves the PR from *blocked by flaky CI* to *blocked by its own regression*. It does **not** prove causation — that is one main run against one PR run, and an intermittent flake produces the same picture. **Two runs is not a rate.**

**Cheapest way to settle it: re-run the failing job.** Green on re-run → flake. Red again → regression confirmed. One click, and it resolves what a single-run comparison can only suggest.

---

**Separately, the production evidence for the bugs this PR fixes has grown and is worsening:**
- 12 dead letters on one host, all addressed to a single agent, all after 10 attempts, for messages that **were** received — delivery succeeding while acks fail
- the **same relay message ID redelivered 8+ times**, rate *increasing* rather than decaying, which rules out producer loops, Slack redelivery, and exponential backoff
- a `register_agent` token invalidated mid-session, recovered only by re-registering
- a task dispatched to a worker whose node had died 7 minutes earlier, with no error surfaced to the sender

To be explicit, because these two halves get conflated: **the severity of the underlying bugs does not make a failing fix mergeable.** The evidence argues for someone picking this up, not for merging it as-is.

Not my repo. No changes made, nothing pushed, no merge attempted — diagnosis only, so whoever owns this does not have to re-derive it.

PR #1392, opened 2026-07-30
fix(verify): stop feature verification reporting false green

khaliqgant:
## Why

A `verify-features` run reported **`COMPLETED — 11 passed, 0 failed`** while Tier 2 had four failing checks. This makes that impossible, then makes a failure actionable.

## The false green had three causes, all fixed

1. Every step set `failOnError: false`, so a non-zero exit scored as a pass. Tiers still use it (all tiers should run even after one fails), but they are no longer the verdict — each check appends `pass`/`fail`/`skip` to `checks.jsonl` and a `verdict` step aggregates it.
2. Nothing aggregated tier results. `verdict` + a terminal `enforce-verdict` now do, and `main()` exits non-zero.
3. `main()` discarded the runner result. It now reads `verdict.json` and **fails closed** when that file is missing.

## A fourth defect nobody asked about

`applyReliabilityDefaults` in `@relayflows/core` force-enables `strategy: "retry"` with a repair agent for any workflow declaring agents. On the pasted run, `tier1-cli-health` failed, the `reporter` agent was handed the gate, edited two files, and it passed. **A verification workflow whose assertions an agent can rewrite until they pass measures nothing.** Fixed with `onError('continue')`, the documented opt-out, confirmed via the built config (`errorHandling: {"strategy":"continue"}`).

## The run also verified the wrong CLI

The sandbox ran **agent-relay 10.0.0** against an 11.3.0 checkout — it logged `Update available: 10.0.0 → 11.3.0`, then reported `relay node deadletters` as unknown. That is a real command in 11.x (`registerCoreCommands(node, …)`, `packages/cli/src/cli/commands/core.ts:394`). A whole run measured a stale published CLI *and* produced a red herring. A `provenance` step now fails on CLI/repo version drift.

## Honest accounting

`cloud whoami` / `cloud session` were scored as product failures when the sandbox simply wasn't cloud-authed. A `capabilities` probe now turns unrunnable checks into **SKIPs with a stated cause** — never a pass, never a fail. `verdict` also fails the run when a tier produced **zero** records, so a tier that crashes before its first check cannot masquerade as clean.

## What happens on failure now

Slack alert → GitHub issue → fix attempt on a branch → **draft** PR with re-verified evidence. An integrity gate refuses a branch that reduced verification call sites, left the repo on `main`, or has no root cause on record — a fixer's cheapest path to green is always to weaken the assertion.

Harness-level breakage (no verdict, tier not run, broker never started, no provider CLI) escalates to NightCTO via the `CloudEvidenceSummary v1` contract. Run and per-failure events go to PostHog, posted directly rather than through the CLI's telemetry client — that one is opt-in *product* telemetry keyed to a real user.

Slack goes through `@relayflows/slack-primitive` pinned to its **cloud-relay** runtime (`/api/v1/slack/post-message`, the workspace's Slack integration). Pinning matters: auto-detection falls back to a local `SLACK_BOT_TOKEN` runtime, so a stray token silently changed the route. This also fixes a wrong endpoint — the previous fallback used `/api/v1/integrations/slack/messages`, which does not exist.

## Coverage

Tiers 1–6 and all six critical paths, including a **real MCP JSON-RPC handshake** asserting the required tool set (previously just `relay mcp --help`).

## Is the feature map current? No — and the old check couldn't tell

`manifest-contract.test.ts` asserts a **hardcoded** list of ~110 commands, so a newly added command is absent from both the manifest and the expectation list and passes silently. `scripts/audit-feature-manifest.mjs` derives the surface instead (recursive `--help` + `tools/list`).

Its first run found **`fleet spawn` and `fleet release` shipping undocumented, therefore unverified.** Both added; audit reports `MANIFEST_CLEAN`. `workflows/audit-feature-manifest.ts` runs this on a schedule with the same Slack/issue/PR/NightCTO/PostHog treatment. Exit 2 (audit broke) is deliberately distinct from exit 0 — a broken audit must never read as a clean manifest.

## Verification

Ran and confirmed:

- Typecheck clean; both dry runs `Validation: PASS`.
- All 30 generated shell bodies pass `sh -n`.
- Verdict gate against all four failure modes: all-pass→PASS, failing check→FAIL, **crashed tier→FAIL**, missing ledger→FAIL. `enforce-verdict` exits 0/1/2 correctly.
- Check harness reports 2p/2f/2s exactly matching the ledger, including direct `record` calls (a bug I found and fixed mid-review: the counters lived in the callers, so multi-step checks logged `0 passed, 0 failed` while the ledger held real results).
- Ledger survives quotes, newlines, tabs, backslashes; bounds 2000-char output to 400.
- Slack proven against a stub server: correct path, bearer auth, channel `C0AEKNLDNKW`. With `SLACK_BOT_TOKEN` set and `CLOUD_API_*` absent it fails loudly instead of falling back.
- Drift detection proven by temporarily deleting the entry I added.
- 32 feature-guardian tests pass; prettier clean.

**Not verified:** no live end-to-end run — the tiers need a broker, workspace, cloud auth, and provider CLIs. The `gh issue create` / `gh pr create` paths were deliberately not executed, since that would file real issues.

## Deliberate omissions

- **No CHANGELOG entry.** The changelog is the user-facing release narrative; this is internal verification tooling, like the explicitly-excluded `relay-feature-guardian`. (There is contrary precedent at line 168.)
- Tier 5/6 checks that spend real money or need a browser (`cloud run`, `fleet spawn`, OAuth logins) are recorded as **explicit SKIPs with reasons** rather than attempted, so the coverage hole is visible instead of hidden.

<a href="https://cubic.dev/pr/AgentWorkforce/relay/pull/1392?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

## Live end-to-end run (replaces the earlier "not verified" caveat)

Ran the whole pipeline for real against the 11.3.0 build with a live broker, an authenticated cloud session, and all four provider CLIs present. Isolated working directory; the repo checkout was never touched.

**Result: 90 passed, 3 failed, 17 skipped → verdict FAIL, `enforce-verdict` exit 1.**

| Tier | Result | p/f/s |
|---|---|---|
| provenance | PASS | 2/0/0 |
| tier1 CLI health | PASS | 26/0/0 |
| tier2 broker | PASS | 8/0/0 |
| tier3 messaging | PASS | 17/0/0 |
| tier4 cross-agent | FAIL | 8/1/0 |
| tier5 cloud | FAIL | 12/2/13 |
| tier6 harnesses | PASS | 11/0/4 |
| critical paths 1–6 | PASS | 6/0/0 |

Tier 6 and CP3 spawned, held, resumed, and released a **real Claude agent**. CP4 completed a real MCP JSON-RPC handshake. CP5 ran a real local workflow and asserted the sentinel plus a completed sync.

**`relay node deadletters` passes against 11.3.0**, confirming the original sandbox failure was purely the stale 10.0.0 CLI — exactly what the new `provenance` step now catches.

### The run found 3 real product bugs

Each reproduced independently, outside the harness, with clean fixtures. Left failing rather than worked around.

1. **`relay message file upload` cannot succeed.** Three distinct errors as the invocation was corrected, ending at server-side `Invalid attachments: file ids must exist in workspace and be complete`. Also, `--help` documents `--text` as defaulting to `""`, but the API rejects empty text with `text is required` — the documented default cannot work.
2. **`relay integration webhook create <url>` returns `channel is required`, but `--help` exposes no `--channel` option.** The command as documented cannot succeed.
3. **`relay agent remove` fails for any agent that has sent a message**, leaking a raw SQL error: `Failed query: delete from "agents" where "agents"."id" = ?`. Exit code is correctly 1, but the message is an internal query dump. This silently accumulated 16 test identities in the workspace during these runs, while every tier still reported clean — so tier 3 now asserts the agent is actually gone after removal instead of assuming it.

`relay fleet config` also fails with `RelaycastMessagingClient.workspace.fleetNodes requires @relaycast/sdk with the workspace fleet nodes API`, which looks like a dependency/capability gap rather than a code defect.

### And 8 bugs in this workflow, now fixed

- **`grep -q` under `set -o pipefail`** — `-q` exits at the first match, the upstream writer takes SIGPIPE, and the pipeline reports failure. A demonstrably RUNNING broker probed as absent. Fixed at 10 piped sites including `run_check`, where it was a latent false-FAIL generator for any long-output check.
- **`capabilities` died mid-probe** — it read `$CLOUD_API_URL` without seeding it, so `set -u` killed the step and every `provider_*` flag went unwritten, silently degrading tier 5/6 gating.
- **Token extraction matched any 20+ char string** — a failed registration yielded the agent's own *name* as its token, so downstream calls failed for unrelated reasons. Now parses the JSON `token` field (7 sites).
- **`VERIFY_AUTOFIX=0` did not stop the fixer** — it gated `file-issue` but not `attempt-fix`, so the documented off switch still let an agent edit the tree.
- **Three checks contradicted the manifest**, which had the correct signatures all along: `message file upload <path> --channel`, `cloud status <runId>`, `integration webhook create <url>`. Good evidence the manifest is worth keeping accurate.
- **CP5 asserted against the launch response** — `workflow run` is async and returns `status: "running"`, so the sentinel could never appear there. It now polls the log as critical-path 5 documents.

### Still not exercised

Slack delivery was verified against a stub server (correct path, bearer auth, channel `C0AEKNLDNKW`) rather than live — this machine has a cloud session but no `CLOUD_API_URL`/`CLOUD_API_TOKEN` in the environment, so the `slack` capability probes 0. The `gh issue create` / `gh pr create` paths were deliberately not executed, since that would file real issues.

**Note:** these runs left ~16 test agents in the Default workspace that cannot be deleted because of product bug 3 above.

---

### `integration webhook create` — fixed here

`POST /v1/webhooks` accepts `{ channel, name? }` and **returns** the url. The command took a `<url>` argument and sent `{ url, event }`, so the required channel was never sent and every call failed with `channel is required`. `url` is not a request field at all; `event` belongs to subscriptions.

- Command now takes `<channel>` with an optional `--name`, matching `create-inbound` — which posts to the *same* endpoint with the *same* schema (`relay.js:428-429`), making them true aliases.
- `RelayCreateWebhookInput` corrected from `{ url, event? }` to `{ channel, name? }`. Anyone passing `url`/`event` was already failing at runtime, so this breaks no working code.
- Regression test pins the corrected contract, including that `url`/`event` are never sent.
- **Verified against the live API:** create returned a real `webhookId`/`url`/`token`, and the workflow's create → trigger → delete lifecycle now passes. Tier 5 went from 12p/2f to 15p/1f.

### The audit had been reading a stale build

`packages/cli/dist` was older than `packages/cli/src`, so the audit derived a surface missing **16 shipped commands** and printed `MANIFEST_CLEAN` across several runs. Same class of error as the sandbox verifying CLI 10.0.0 against an 11.3.0 checkout — and the reason it went unnoticed.

Now documented: `agent me`, `agent presence`, the `cloud integration` tree (catalog, connect, connections, disconnect) and the `cloud room` tree (session, revoke-session, members, remove-member, invite, invites, accept, revoke-invite). Two new manifest categories, each mapped to a verification procedure. **180 → 194 features.**

The audit now warns when `dist` is older than `src`, so a stale build cannot yield a falsely clean result again.

### Filed, not fixable in this repo

- **#1398** — `relay message file upload` cannot succeed. `serializeAttachmentInputs` JSON-stringifies `{type,path}` and sends it where a **file id** is required; the file is never uploaded. The API needs a three-step presigned upload and `@relaycast/sdk` exposes no files API, so the SDK must gain one first.
- **#1399** — `relay agent remove` fails for any agent with message history (likely an FK) and leaks `Failed query: delete from "agents" where "agents"."id" = ?`. Server-side; the CLI only calls `agents.delete`.
- **#1400** — `relay fleet config` fails on `@relaycast/sdk` missing the workspace fleet-nodes API.

All three reproduced independently with clean fixtures. Their checks are left **failing** rather than skipped, so the run stays red until they are fixed.

### DIFF
--- a/.agentworkforce/agents/relay-feature-guardian/manifest-contract.test.ts
+++ b/.agentworkforce/agents/relay-feature-guardian/manifest-contract.test.ts
@@ -137,6 +137,8 @@ describe('feature manifest contract', () => {
       'relay skills add',
       'relay mcp',
       'relay fleet nodes',
+      'relay fleet spawn',
+      'relay fleet release',
       'relay fleet config',
       'relay fleet enable',
       'relay fleet disable',
--- a/.agentworkforce/features/manifest.yaml
+++ b/.agentworkforce/features/manifest.yaml
@@ -1,5 +1,5 @@
 version: '1.1'
-updated: '2026-07-20'
+updated: '2026-07-30'
 
 # Every user-facing feature in this repo, categorized and scored.
 #
@@ -596,6 +596,20 @@ categories:
         name: List Fleet Nodes
         cli: relay fleet nodes
         description: Show all fleet nodes registered in the workspace (returns empty list if none configured)
+        location: packages/cli/src/cli/commands/fleet.ts
+        verify_tier: 5
+
+      - id: fleet-spawn
+        name: Spawn Fleet Agent
+        cli: relay fleet spawn <cli> [--name <name>] [--task <text>] [--node <name>] [--persona <persona>] [--model <model>]
+        description: Spawn a worker agent onto a live fleet node, by explicit node or by persona placement
+        location: packages/cli/src/cli/commands/fleet.ts
+        verify_tier: 5
+
+      - id: fleet-release
+        name: Release Fleet Agent
+        cli: relay fleet release <name> [--reason <reason>] [--delete-agent]
+        description: Release a spawned fleet agent, optionally deleting the agent identity
         location: packages/cli/src/cli/commands/fleet.ts
         verify_tier: 5
 
--- a/.agentworkforce/features/manifest.yaml
+++ b/.agentworkforce/features/manifest.yaml
@@ -846,7 +846,7 @@ categories:
       - id: webhook-create
         name: Create Webhook
         cli: relay integration webhook create <channel> [--name <name>]
-        description: Register an outbound webhook
+        description: Register an inbound webhook that external services POST to, delivering into a channel
         location: packages/cli/src/cli/commands/integration.ts
         verify_tier: 5
 
--- a/.agentworkforce/features/manifest.yaml
+++ b/.agentworkforce/features/manifest.yaml
@@ -33,6 +33,8 @@ verification:
     local-workflows: local-workflow-lifecycle
     cloud: cloud-workflows
     cloud-workers: cloud-workers
+    cloud-integrations: integrations-and-webhooks
+    cloud-rooms: cloud-workflows
     fleet: fleet-management
     workspace: workspace-management
     skills: skills-installation
@@ -121,6 +123,20 @@ categories:
         name: List Agents
         cli: relay agent list [--status <status>]
         description: Show all agents in the workspace with status
+        location: packages/cli/src/cli/commands/agent.ts
+        verify_tier: 3
+
+      - id: agent-me
+        name: Show Current Agent Identity
+        cli: relay agent me
+        description: Show the identity the current agent token resolves to
+        location: packages/cli/src/cli/commands/agent.ts
+        verify_tier: 3
+
+      - id: agent-presence
+        name: List Agent Presence
+        cli: relay agent presence
+        description: List visible agent presence in the workspace
         location: packages/cli/src/cli/commands/agent.ts
         verify_tier: 3
 
@@ -554,6 +570,100 @@ categories:
         location: packages/cli/src/cli/commands/cloud.ts
         verify_tier: 5
 
+  cloud-integrations:
+    name: Cloud Integrations
+    description: Connect third-party providers to the workspace through Cloud
+    criticality: standard
+    features:
+      - id: cloud-integration-catalog
+        name: Integration Catalog
+        cli: relay cloud integration catalog [--static] [--json]
+        description: Discover static and dynamic (Nango, Composio) Cloud integrations
+        location: packages/cli/src/cli/commands/cloud.ts
+        verify_tier: 5
+
+      - id: cloud-integration-connect
+        name: Connect Integration
+        cli: relay cloud integration connect <provider> [--json]
+        description: Create a Cloud connection session for a provider
+        location: packages/cli/src/cli/commands/cloud.ts
+        verify_tier: 6
+
+      - id: cloud-integration-connections
+        name: List Connections
+        cli: relay cloud integration connections [--json]
+        description: List connected workspace integrations
+        location: packages/cli/src/cli/commands/cloud.ts
+        verify_tier: 5
+
+      - id: cloud-integration-disconnect
+        name: Disconnect Integration
+        cli: relay cloud integration disconnect <provider> [--json]
+        description: Disconnect a provider from the workspace
+        location: packages/cli/src/cli/commands/cloud.ts
+        verify_tier: 5
+
+  cloud-rooms:
+    name: Cloud Rooms
+    description: Workspace-scoped multiplayer rooms and their membership
+    criticality: standard
+    features:
+      - id: cloud-room-session
+        name: Create Room Session
+        cli: relay cloud room session
+        description: Create or resume this device's full room-participant session
+        location: packages/cli/src/cli/commands/cloud.ts
+        verify_tier: 5
+
+      - id: cloud-room-revoke-session
+        name: Revoke Room Session
+        cli: relay cloud room revoke-session
+        description: Revoke this member's scoped session for one device
+        location: packages/cli/src/cli/commands/cloud.ts
+        verify_tier: 5
+
+      - id: cloud-room-members
+        name: List Room Members
+        cli: relay cloud room members
+        description: List workspace room members
+        location: packages/cli/src/cli/commands/cloud.ts
+        verify_tier: 5
+
+      - id: cloud-room-remove-member
+        name: Remove Room Member
+        cli: relay cloud room remove-member <member-id>
+        description: Remove a member and revoke their live room access
+        location: packages/cli/src/cli/commands/cloud.ts
+        verify_tier: 5
+
+      - id: cloud-room-invite
+        name: Invite Room Participant
+        cli: relay cloud room invite
+        description: Invite a full participant to a workspace room
+        location: packages/cli/src/cli/commands/cloud.ts
+        verify_tier: 5
+
+      - id: cloud-room-invites
+        name: List Room Invitations
+        cli: relay cloud room invites
+        description: List workspace room invitations
+        location: packages/cli/src/cli/commands/cloud.ts
+        verify_tier: 5
+
+      - id: cloud-room-accept
+        name: Accept Room Invitation
+        cli: relay cloud room accept
+        description: Accept an email-bound room invitation
+        location: packages/cli/src/cli/commands/cloud.ts
+        verify_tier: 6
+
+      - id: cloud-room-revoke-invite
+        name: Revoke Room Invitation
+        cli: relay cloud room revoke-invite <invite-id>
+        description: Revoke an unused workspace room invitation
+        location: packages/cli/src/cli/commands/cloud.ts
+        verify_tier: 5
+
   cloud-workers:
     name: Cloud Worker Nodes
     description: Register and manage machines as cloud worker nodes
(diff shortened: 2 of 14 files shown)


khaliqgant:
Thanks — all of these were real. Addressed in `1062989c5`, with what I did and how I checked it.

### Codex

**P1 · Verify the checkout rather than matching only its version** — Correct, and this is the one that mattered most: between releases a global `relay` reports the same version as the branch, so the gate passed while every check ran the global binary. `provenance` now resolves the binary with `realpathSync` and requires it to sit inside `git rev-parse --show-toplevel`. Verified both ways: same version + binary outside the checkout → `FAIL`; non-git tree → stated `SKIP` (falling back to `$PWD` would only prove "under the cwd", which isn't provenance).

**P1 · Pass the file path and channel flag to upload** — Already fixed in `ec17d7ecf`; you reviewed `72539f7b8`. A live run hit it too. Worth noting the command is still broken for a different reason (#1398): the file is never uploaded, so the check stays red on purpose.

**P1 · Honor VERIFY_AUTOFIX before launching the fixer** — Already fixed in `ec17d7ecf` via `autofix.env`, honoured by `attempt-fix` and `fix-integrity`, so `open-pr` can't push.

**P1 · Honor AUDIT_AUTOFIX before launching the editor** — Was still open; fixed the same way.

**P1 · Fail when the MCP surface cannot be derived** — Agreed, and it's the exact failure mode this PR exists to stop. `mcpTools === null` now exits 2. Verified with a stub CLI that serves `--help` normally but dies on `mcp`: `AUDIT_ERROR: MCP tool list could not be derived… Refusing to report a partial audit as clean` → exit 2.

**P1 · Require the full CLI leaf when detecting stale entries** — Right, and worse than described: since `deriveCliLeaves` records every parent group, almost any removed child was hidden. Now requires an exact leaf, with positional-dispatch parents named explicitly in `POSITIONAL_DISPATCH` (`telemetry`). Verified: a documented `fleet ghost` under a surviving `fleet` is now reported stale, while `telemetry enable` still isn't.

**P2 · Supply run IDs to cloud status and logs** — `cloud status` was already a stated skip; `cloud logs` still recorded a false pass off its error text through `|| true`. Now also a stated skip.

### CodeRabbit

**Guard `child.stdin` against EPIPE** — Good catch on the consequence, not just the crash: an uncaught `'error'` isn't a rejection, so it escaped `main().catch`, exited 1, and the workflow would read that as confirmed drift and dispatch the editor. Handler added.

**Ledger integrity gap** — Fixed on both sides. `record` strips all control bytes, and the verdict now *counts* unparseable lines and fails the run on them, rather than dropping them. Verified: injecting a truncated line yields `FAIL … 1 unparseable ledger line(s) — check results were lost`.

**`[ $? -le 124 ]` treats real failures as success** — Fixed to `0|124|137`. Verified `timeout 5 false` now records `tail-failed rc=1` where it previously passed.

**`grep -qiv running` does not assert "broker is down"** — Fixed to `! … | grep -i running`.

**Cleanup pattern too broad** — Fixed; teardown is now anchored to this run's `SUFFIX`, so pre-existing workspace state can't match.

**Unguarded `JSON.parse(audit.json)`** — Both consumers now degrade to an explicit "report unreadable" message instead of an empty body swallowed by `|| true`.

**Integrity gate only protects the audit script** — Agreed; asserting one file was the wrong shape. It now requires the entire changed-file set to be the manifest alone, so drift can't be resolved by deleting a command or relaxing `manifest-contract.test.ts`. The editor task was updated to match.

**Describe this as an inbound webhook** — Fixed.

### Not applied

**Use the standard `[Unreleased]` heading** — Declining this one. `CLAUDE.md` in this repo requires the pending release level in the heading (`[Unreleased - Patch]` / `- Minor` / `- Major`, monotonic, reset on release). `[Unreleased - Patch]` is the correct form here.

---

Re-verified after the changes: typecheck clean, both workflows `Validation: PASS`, all 30 generated shell bodies pass `sh -n`, `MANIFEST_CLEAN`, 91 tests green.

PR #1402, opened 2026-07-30
[AR-448] Make Relay workspace identity durable across node restarts

khaliqgant:
Closes AR-448.

## The bug

A node started with no project-pinned workspace fell straight through to the broker, which mints a brand-new messaging-only workspace (`startup_single_session_set_from_sources`, `crates/broker/src/relaycast/auth.rs`).

Nothing errored. The node came up, the resident agent registered, everything looked healthy — but the agent was a stranger in a different workspace with a new address, so every message sent to its previous address went nowhere. `khaliq-chief` became a new process-lifetime agent on each such start.

## The fix

`node up` (and the `local up` alias) now resolves the workspace in this order:

1. explicit `--workspace-key` / `RELAY_WORKSPACE_KEY` / `AGENT_RELAY_WORKSPACE_KEY` / `RELAY_API_KEY`, or a `RELAY_NODE_TOKEN` that already implies a workspace;
2. the project pin from a previous start;
3. **the machine-global canonical workspace** (`agent-relay workspace join|switch`) — new;
4. mint a new workspace — last resort.

Step 3 is the fix. The resolved key is pinned to the project after startup, so later starts take step 2 and no key is ever copied by hand. A machine with no canonical workspace set behaves exactly as before.

Also made the invariant observable rather than assumed:

- `workspace active --json` gains a `dataPlane` object (`unified`, the shared `workspaceId`, per-plane IDs, and the names of any that diverge). The human output prints the Relaycast ID it previously omitted. `--require-unified` turns a divergence into a non-zero exit for setup doctors and supervisors.
- `node status` prints the durable `Workspace:` ID next to the masked workspace key.

## Acceptance

**`workspace active --json` proves one data-plane workspace ID** — run against the live `default` workspace:

```
$ agent-relay workspace active --json --require-unified
{
  "name": "default",
  "key": "rk_live_…9812",
  "cloudWorkspaceId": "50587328-441d-4acb-b8f3-dbe1b3c5de99",
  "relaycastWorkspaceId": "rw_7ccfea89",
  "relaycastApiKey": "rk_live_…9812",
  "relayfileWorkspaceId": "rw_7ccfea89",
  "relayauthWorkspaceId": "rw_7ccfea89",
  ...
  "dataPlane": {
    "unified": true,
    "workspaceId": "rw_7ccfea89",
    "planes": { "relaycast": "rw_7ccfea89", "relayfile": "rw_7ccfea89", "relayauth": "rw_7ccfea89" },
    "divergent": []
  }
}
$ echo $?
0
```

```
$ agent-relay workspace active --require-unified
Workspace: default
Cloud workspace ID: 50587328-441d-4acb-b8f3-dbe1b3c5de99
Relaycast workspace ID: rw_7ccfea89
Relayfile workspace ID: rw_7ccfea89
Relayauth workspace ID: rw_7ccfea89
Data-plane workspace ID: rw_7ccfea89 (unified)
```

**`node up` uses the canonical workspace with no manual key copying** — `core.test.ts` › *up falls back to the machine-global canonical workspace when nothing else selects one*, and *up prefers the project pin over the machine-global canonical workspace*.

**Stop/start regression test** — `packages/cli/src/cli/lib/workspace-identity-restart.test.ts`. Each `start()` is a fresh CLI process (new env, new deps) over one persistent checkout and `AGENT_RELAY_HOME`, which is what a stop/start looks like from the CLI. It models the two behaviors that decide the outcome: the broker joins `RELAY_WORKSPACE_KEY` when set and otherwise mints, and Relaycast returns the *existing* agent when a name is re-registered in a workspace it already belongs to.

- keeps one workspace and one resident address across a restart
- joins the canonical workspace on a first start with no project pin
- pins the canonical workspace to the project so later starts resume it
- explicit `--workspace-key` still wins
- a second checkout shares the canonical workspace and the resident address
- negative control: with no canonical workspace, a second checkout is a different node entirely

**Status and logs never print credentials** — `core.test.ts` › *status reports the durable workspace ID without leaking any credential* asserts the raw workspace key, node token, `ot_live_` tokens, and `?token=`-style URLs are all absent. `workspace.test.ts` asserts the same for `workspace active` human output. The canonical-workspace fallback logs only its source, never the key.

**Documented** — `specs/workspace-identity.md` covers the invariant, resolution order, how to prove it, secret handling, and migration for existing local nodes (including the one behavior change: a node with no project pin now joins the canonical workspace instead of minting, which moves its resident agents there).

## Test evidence

```
$ npx vitest run packages/cloud/src/workspace-convergence.test.ts \
    packages/cli/src/cli/lib/workspace-identity-restart.test.ts \
    packages/cli/src/cli/commands/workspace.test.ts \
    packages/cli/src/cli/commands/core.test.ts \
    packages/cli/src/cli/commands/node.test.ts \
    packages/cli/src/cli/lib/broker-lifecycle.test.ts

 Test Files  6 passed (6)
      Tests  146 passed (146)
```

Full `packages/cli` + `packages/cloud`: **1095 passed, 20 skipped, 0 failed.** `npm run typecheck` clean; `npm run lint` 0 errors (no new warnings).

One pre-existing environmental failure, `integration-relayfile-contract.test.ts` › *replaces a stale v1 daemon only when the installed binary can serve v3*, reproduces identically on a stashed clean tree — the sandbox can't invoke `lsof`. Unrelated to this change.

## Note

`node up`'s test harness now defaults `AGENT_RELAY_HOME` to an isolated temp dir. Without that, the new store fallback made tests read (and nearly print) whatever workspace the developer's own machine had active.

Do not merge — the principal owns the merge gate.

<a href="https://cubic.dev/pr/AgentWorkforce/relay/pull/1402?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased - Patch]
 
+### Added
+
+- `agent-relay workspace active` now reports whether Relaycast, Relayfile, and RelayAuth resolve the workspace to one data-plane ID. `--json` gains a `dataPlane` object (`unified`, the shared `workspaceId`, per-plane IDs, and the names of any that diverge), the human output prints the Relaycast ID it used to omit, and `--require-unified` turns a divergence into a non-zero exit for setup doctors and supervisors.
+- `agent-relay node status` prints the durable `Workspace:` ID alongside the masked workspace key, so an operator can confirm a restart preserved workspace identity. See `specs/workspace-identity.md`.
+
 ### Fixed
+
+- A node started with no project-pinned workspace no longer mints a throwaway workspace. `agent-relay node up` now falls back to the machine-global canonical workspace (`agent-relay workspace join|switch`) before letting the broker create one, so the node and its resident agents keep the same workspace — and the same delivery addresses — across a stop/start. Explicit `--workspace-key`, workspace env vars, and an existing project pin all still win; a machine with no canonical workspace set behaves as before.
 
 - `agent-relay integration webhook create` now works. It took a `<url>` argument and sent `{ url, event }`, but `POST /v1/webhooks` accepts `{ channel, name? }` and returns the URL — so every invocation failed with `channel is required`. It now takes `<channel>` with an optional `--name`, matching `create-inbound`, which posts to the same endpoint.
 - `@agent-relay/sdk` `RelayCreateWebhookInput` declared a required `url` and an `event`, neither of which the endpoint accepts. It is now `{ channel, name? }`. Code passing `url`/`event` was already failing at runtime.
--- a/packages/cli/src/cli/commands/core.test.ts
+++ b/packages/cli/src/cli/commands/core.test.ts
@@ -1246,9 +1246,7 @@ describe('registerCoreCommands', () => {
     await runCommand(program, ['status']);
 
     expect(deps.log).toHaveBeenCalledWith('Workspace: rw_7ccfea89');
-    const output = (deps.log as unknown as { mock: { calls: unknown[][] } }).mock.calls
-      .flat()
-      .join('\n');
+    const output = (deps.log as unknown as { mock: { calls: unknown[][] } }).mock.calls.flat().join('\n');
     expect(output).not.toContain('rk_live_teststatus123');
     expect(output).not.toContain('nt_live_nodetoken456');
     // Observer URLs carry a scoped token in the query string; status never
--- a/packages/cli/src/cli/commands/core.test.ts
+++ b/packages/cli/src/cli/commands/core.test.ts
@@ -2,7 +2,7 @@ import { Command } from 'commander';
 import nodeFs from 'node:fs';
 import os from 'node:os';
 import nodePath from 'node:path';
-import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
 import { readProjectWorkspaceKey, readProjectWorkspaceSession } from '../lib/project-workspace-key.js';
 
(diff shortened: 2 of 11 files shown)


khaliqgant:
Closing in favour of #1447, on Khaliq's decision.

**Why, in one line:** AR-448's substance already landed in `main` through PR #1429, so neither of these is the change that ships it — and both are now stale patches against a file that has since been rewritten underneath them.

Verified with `git merge-base --is-ancestor` rather than by reading PR titles:

- `4acdd97d4` *fix(cli): resolve the broker workspace through one precedence ladder* — **in main**
- `5c2ad8ee3` *fix(broker): reclaim a node's own registration across restart, hash the identity proof* — **in main**
- `492515a20` and `27494c862`, the two AR-448-branded commits — **not in main**

Both PRs still edit `packages/cli/src/cli/commands/workspace.ts`, which #1429 rewrote while these sat untouched for eight days. Merging either would mean re-litigating code that has already moved.

**What was kept.** The genuinely unique value here was #1402's test evidence, and it has been harvested onto current main in #1447 — adapted, not copied: the restart suite grew from 341 to 412 lines because #1402's model needed correcting (see below), plus `workspace-convergence.ts` wired into `workspace active` so it is a module something actually calls, and `specs/workspace-identity.md`.

**What was found while harvesting, and it matters beyond this PR.** #1402's Relaycast stand-in encoded *"re-registering a name in a workspace it already belongs to returns the existing agent"* — unconditionally. That premise is now false by design: `b4b96dfb3` made a name collision **fail closed** unless the caller proves same-work-unit via a hashed identity key. So **workspace convergence is necessary and not sufficient for address identity** — identity has a second half, the admission decision on the name, which landed a week after AR-448 as its own fix. Two checkouts sharing one workspace is a supported state that AR-448 read as identity-preserving and the gate now correctly reads as an impersonation attempt.

No work is lost. If anything here is wanted that #1447 does not carry, say so and it can be brought across.

PR #1405, opened 2026-07-31
fix(plugins): stop exposing workspace keys in observer links

### LINKED ISSUE #1397: Plugins: stop exposing workspace admin keys in observer links
## Problem

Relay ships six plugin surfaces that construct `https://agentrelay.com/observer?key=<workspace-key>`, and several tell the agent to print the real key. The sharpest case is `plugins/gemini-relay-extension/commands/status/status.toml`, which explicitly says not to print a placeholder.

Affected surfaces:

- `plugins/gemini-relay-extension/GEMINI.md`
- `plugins/gemini-relay-extension/commands/status/status.toml`
- `plugins/gemini-relay-extension/commands/team/team.toml`
- `plugins/gemini-relay-extension/commands/fanout/fanout.toml`
- `plugins/gemini-relay-extension/hooks/session-start.sh`
- `plugins/codex-relay-skill/SKILL.md`

This conflicts with #1379/#1380: the CLI is removing credentials from default output while shipped agent instructions still require printing an owner/admin credential into a transcribed terminal and query string.

## Credential contract verified from source

A workspace key does **not** authenticate the observer realtime stream. Relaycast `authenticateRealtimeWs` rejects workspace tokens with `Observer token required for workspace stream`; only an `ot_live_*` token with `stream:read` is accepted.

The dashboard currently accepts either `rk_live_*` or `ot_live_*` as a login credential. When given a workspace key, its server-side login route validates the key and mints a scoped observer token for the realtime socket. Thus the existing links may function through the dashboard login bridge, but they expose a more powerful workspace administration key than observation requires. Relaycast documentation also says not to embed workspace keys in browser clients or third-party dashboards.

## Required outcome

- Remove every instruction and hook that prints or embeds a workspace key in an observer URL.
- Define a safe follow-along flow using a scoped `ot_live_*` observer token, or omit the link when no observer token has been explicitly provisioned.
- Never place either credential in default transcribed output or a query string without an explicit reveal action and warning; prefer a handoff that avoids URL credential material where supported.
- Add regression coverage that searches shipped plugin assets for workspace-key observer URL construction and imperative instructions to print the real key.
- Update plugin/user documentation so workspace admin keys and observer tokens are not conflated.

## Relationship

Follow-up to #1379 and #1380. This is intentionally separate from #1380 so the CLI redaction fix remains reviewable and mergeable on its current scope.

willwashburn:
## What changed

- Removed workspace-key observer URLs and imperative secret-printing instructions from the six Gemini/Codex plugin surfaces in #1397.
- Changed the Gemini session-start hook to detect configuration without reading the workspace key into the hook or injected context.
- Documented the safe fallback: observation requires a separately provisioned, read-only observer token delivered through an explicit secret handoff; otherwise no link is shown.
- Removed `node up` / `node status` from both mirrored agent-facing `using-agent-relay` skill command references while published versions through 11.3.0 still print live credentials.
- Added a repository regression that scans tracked plugin assets, the six named observer surfaces, the session hook, and both mirrored agent-facing skills.
- Added the user-visible security change to `[Unreleased - Patch]`.

## Why

Workspace keys have administrative authority. The shipped plugin instructions and hook embedded them in clickable observer URLs and terminal transcripts, while the published 11.3.0 CLI also prints them from `node up` / `node status`.

The root cause was conflating an admin workspace credential with a scoped, read-only observer token and treating a URL query string as a safe handoff.

## Impact

Agents no longer disclose workspace keys while coordinating workers or reporting Relay status. Users who need observation are directed to a separately provisioned observer token without printing that token or constructing a credential-bearing URL in the transcript.

This mitigation should merge separately and immediately after review; it does not need to wait for the human-gated npm release. The release package must still ship #1380 as a patch because published CLI 11.3.0 remains independently exposed.

## Control lifetimes

- **Permanent instruction fix:** the six #1397 plugin surfaces constructed `observer?key=<workspace-key>` themselves. No CLI release can repair those files; their workspace-key URL and printing instructions must remain absent.
- **Interim propagation mitigation:** the two mirrored `using-agent-relay` skills invoked `node up` / `node status`, whose CLI output leaks through 11.3.0. Publishing #1380 moves that fix from merged to published, but each consumer remains exposed until upgraded. The command removals protect that interval.

The regression intentionally enforces both controls. After #1380 is published and consumers have upgraded, the skill-command prohibition may be reevaluated separately; the six plugin-surface protections are not part of that retirement criterion and must not be restored.

## Validation

- `npm run typecheck`
- `env -u AGENT_RELAY_MACHINE_ID -u AGENT_RELAY_DISTINCT_ID AGENT_RELAY_DATA_DIR=<isolated> npx vitest run` — 1,664 passed, 14 skipped
- `npx vitest run packages/cli/src/cli/plugin-credential-safety.test.ts`
- `npm --prefix plugins/gemini-relay-extension run check`
- `git diff --cached --check`
- mirrored skill files verified byte-identical

Closes #1397.

### DIFF
--- a/.agents/skills/using-agent-relay/SKILL.md
+++ b/.agents/skills/using-agent-relay/SKILL.md
@@ -189,13 +189,16 @@ remove_agent(name: "reviewer-1", reason: "Review accepted")
 
 ## Current CLI Reference
 
+Startup and status commands are intentionally omitted from these agent-facing
+examples. Published Agent Relay versions through 11.3.0 can print live
+workspace credentials when those commands run in a transcribed session. Upgrade
+to a patched release before running them there.
+
 These are the current CLI forms for local broker and SDK-backed messaging
 operations:
 
 ```bash
 agent-relay status
-agent-relay node up --verbose
-agent-relay node status --wait-for 10
 agent-relay node agent list
 agent-relay node agent spawn claude --name Worker --task "Use https://agentrelay.com/skill and ACK over Relay."
 agent-relay node tail --agent Worker
--- a/.agents/skills/using-agent-relay/SKILL.md
+++ b/.agents/skills/using-agent-relay/SKILL.md
@@ -192,7 +192,7 @@ remove_agent(name: "reviewer-1", reason: "Review accepted")
 Startup and status commands are intentionally omitted from these agent-facing
 examples. Published Agent Relay versions through 11.3.0 can print live
 workspace credentials when those commands run in a transcribed session. Upgrade
-to a patched release before running them there.
+to Agent Relay 11.3.1 or later before running them there.
 
 These are the current CLI forms for local broker and SDK-backed messaging
 operations:
--- a/.claude/skills/using-agent-relay/SKILL.md
+++ b/.claude/skills/using-agent-relay/SKILL.md
@@ -189,13 +189,16 @@ remove_agent(name: "reviewer-1", reason: "Review accepted")
 
 ## Current CLI Reference
 
+Startup and status commands are intentionally omitted from these agent-facing
+examples. Published Agent Relay versions through 11.3.0 can print live
+workspace credentials when those commands run in a transcribed session. Upgrade
+to a patched release before running them there.
+
 These are the current CLI forms for local broker and SDK-backed messaging
 operations:
 
 ```bash
 agent-relay status
-agent-relay node up --verbose
-agent-relay node status --wait-for 10
 agent-relay node agent list
 agent-relay node agent spawn claude --name Worker --task "Use https://agentrelay.com/skill and ACK over Relay."
 agent-relay node tail --agent Worker
--- a/.claude/skills/using-agent-relay/SKILL.md
+++ b/.claude/skills/using-agent-relay/SKILL.md
@@ -192,7 +192,7 @@ remove_agent(name: "reviewer-1", reason: "Review accepted")
 Startup and status commands are intentionally omitted from these agent-facing
 examples. Published Agent Relay versions through 11.3.0 can print live
 workspace credentials when those commands run in a transcribed session. Upgrade
-to a patched release before running them there.
+to Agent Relay 11.3.1 or later before running them there.
 
 These are the current CLI forms for local broker and SDK-backed messaging
 operations:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `agent-relay integration webhook create` now works. It took a `<url>` argument and sent `{ url, event }`, but `POST /v1/webhooks` accepts `{ channel, name? }` and returns the URL — so every invocation failed with `channel is required`. It now takes `<channel>` with an optional `--name`, matching `create-inbound`, which posts to the same endpoint.
 - `@agent-relay/sdk` `RelayCreateWebhookInput` declared a required `url` and an `event`, neither of which the endpoint accepts. It is now `{ channel, name? }`. Code passing `url`/`event` was already failing at runtime.
 
+### Security
+
+- Bundled Gemini and Codex relay instructions and hooks no longer expose workspace administration keys in observer URLs or terminal transcripts; observation now requires a separately provisioned, read-only observer token.
+
 ## [11.3.0] - 2026-07-30
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `agent-relay cloud whoami` prints the organization and workspace IDs alongside their names.
 
+### Security
+
+- Bundled Gemini and Codex relay instructions and hooks no longer expose workspace administration keys in observer URLs or terminal transcripts; observation now requires a separately provisioned, read-only observer token.
+
 ## [11.3.1] - 2026-07-31
 
 ### Fixed
@@ -23,10 +27,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - A CLI command that fails with a non-`Error` value now reports it. A protocol-shaped rejection such as `{ status: 401 }` from a broker client printed `[object Object]`, and an `Error` with an empty message printed nothing at all; both now surface the message, status, and code, with credentials redacted. Applies to the top-level failure handler, `node agent attach`, and broker request failures.
 - `agent-relay integration webhook create` now works. It took a `<url>` argument and sent `{ url, event }`, but `POST /v1/webhooks` accepts `{ channel, name? }` and returns the URL — so every invocation failed with `channel is required`. It now takes `<channel>` with an optional `--name`, matching `create-inbound`, which posts to the same endpoint.
 - `@agent-relay/sdk` `RelayCreateWebhookInput` declared a required `url` and an `event`, neither of which the endpoint accepts. It is now `{ channel, name? }`. Code passing `url`/`event` was already failing at runtime.
-
-### Security
-
-- Bundled Gemini and Codex relay instructions and hooks no longer expose workspace administration keys in observer URLs or terminal transcripts; observation now requires a separately provisioned, read-only observer token.
 
 ## [11.3.0] - 2026-07-30
 
--- a/packages/cli/src/cli/plugin-credential-safety.test.ts
+++ b/packages/cli/src/cli/plugin-credential-safety.test.ts
@@ -1,5 +1,4 @@
-import { execFileSync } from 'node:child_process';
-import { readFileSync } from 'node:fs';
+import { readFileSync, readdirSync } from 'node:fs';
 import { dirname, join, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
 import { describe, expect, it } from 'vitest';
@@ -24,20 +23,24 @@ function readRepoFile(path: string): string {
   return readFileSync(join(repoRoot, path), 'utf8');
 }
 
-function listTrackedPluginFiles(): string[] {
-  return execFileSync(
-    'git',
-    ['ls-files', '-z', 'plugins/gemini-relay-extension', 'plugins/codex-relay-skill'],
-    { cwd: repoRoot, encoding: 'utf8' }
-  )
-    .split('\0')
-    .filter(Boolean);
+function listPluginFiles(directory: string): string[] {
+  return readdirSync(directory, { withFileTypes: true }).flatMap((entry) => {
+    if (entry.name.startsWith('.') || entry.name === 'node_modules') return [];
+
+    const path = join(directory, entry.name);
+    return entry.isDirectory() ? listPluginFiles(path) : [path];
+  });
 }
 
 describe('shipped relay plugin credential safety', () => {
   it('does not construct observer URLs from workspace keys in shipped plugin assets', () => {
-    for (const path of listTrackedPluginFiles()) {
-      const source = readRepoFile(path);
+    const pluginFiles = [
+      ...listPluginFiles(join(repoRoot, 'plugins/gemini-relay-extension')),
+      ...listPluginFiles(join(repoRoot, 'plugins/codex-relay-skill')),
+    ];
+
+    for (const path of pluginFiles) {
+      const source = readFileSync(path, 'utf8');
       expect(source, path).not.toMatch(/agentrelay\.com\/observer\?key=/i);
     }
   });
--- a/packages/cli/src/cli/plugin-credential-safety.test.ts
+++ b/packages/cli/src/cli/plugin-credential-safety.test.ts
@@ -1,0 +1,71 @@
+import { execFileSync } from 'node:child_process';
+import { readFileSync } from 'node:fs';
+import { dirname, join, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { describe, expect, it } from 'vitest';
+
+const repoRoot = resolve(dirname(fileURLToPath(import.meta.url)), '../../../..');
+
+const observerInstructionFiles = [
+  'plugins/gemini-relay-extension/GEMINI.md',
+  'plugins/gemini-relay-extension/commands/status/status.toml',
+  'plugins/gemini-relay-extension/commands/team/team.toml',
+  'plugins/gemini-relay-extension/commands/fanout/fanout.toml',
+  'plugins/gemini-relay-extension/hooks/session-start.sh',
+  'plugins/codex-relay-skill/SKILL.md',
+] as const;
+
+const agentFacingSkillFiles = [
+  '.agents/skills/using-agent-relay/SKILL.md',
+  '.claude/skills/using-agent-relay/SKILL.md',
+] as const;
+
+function readRepoFile(path: string): string {
+  return readFileSync(join(repoRoot, path), 'utf8');
+}
+
+function listTrackedPluginFiles(): string[] {
+  return execFileSync(
+    'git',
+    ['ls-files', '-z', 'plugins/gemini-relay-extension', 'plugins/codex-relay-skill'],
+    { cwd: repoRoot, encoding: 'utf8' }
+  )
+    .split('\0')
+    .filter(Boolean);
+}
+
+describe('shipped relay plugin credential safety', () => {
+  it('does not construct observer URLs from workspace keys in shipped plugin assets', () => {
+    for (const path of listTrackedPluginFiles()) {
+      const source = readRepoFile(path);
+      expect(source, path).not.toMatch(/agentrelay\.com\/observer\?key=/i);
+    }
+  });
+
+  it('does not instruct agents to print real workspace keys or observer links', () => {
+    for (const path of observerInstructionFiles) {
+      const source = readRepoFile(path);
+      expect(source, path).not.toMatch(/\b(?:actual key|real clickable URL)\b/i);
+      expect(source, path).not.toMatch(/\bprint the observer URL\b/i);
+
+      const credentialPrintLines = source
+        .split('\n')
+        .filter((line) => /\bprint\b/i.test(line) && /(?:workspace key|observer URL|ot_live_)/i.test(line));
+
+      for (const line of credentialPrintLines) {
+        expect(line, `${path}: ${line}`).toMatch(/\b(?:never|do not)\b[^\n]*\bprint\b/i);
+      }
+    }
+
+    const sessionStartHook = readRepoFile('plugins/gemini-relay-extension/hooks/session-start.sh');
+    expect(sessionStartHook).not.toMatch(/\bWORKSPACE_KEY\s*=/);
+    expect(sessionStartHook).not.toMatch(/\$\{?WORKSPACE_KEY\b/);
+  });
+
+  it('keeps credential-printing startup commands out of mirrored agent-facing skills', () => {
+    const [agentsSkill, claudeSkill] = agentFacingSkillFiles.map(readRepoFile);
+
+    expect(agentsSkill).toBe(claudeSkill);
+    expect(agentsSkill).not.toMatch(/\bagent-relay node (?:up|status)\b/);
+  });
+});
--- a/packages/cli/src/cli/plugin-credential-safety.test.ts
+++ b/packages/cli/src/cli/plugin-credential-safety.test.ts
@@ -5,15 +5,6 @@ import { describe, expect, it } from 'vitest';
 
 const repoRoot = resolve(dirname(fileURLToPath(import.meta.url)), '../../../..');
 
-const observerInstructionFiles = [
-  'plugins/gemini-relay-extension/GEMINI.md',
-  'plugins/gemini-relay-extension/commands/status/status.toml',
-  'plugins/gemini-relay-extension/commands/team/team.toml',
-  'plugins/gemini-relay-extension/commands/fanout/fanout.toml',
-  'plugins/gemini-relay-extension/hooks/session-start.sh',
-  'plugins/codex-relay-skill/SKILL.md',
-] as const;
-
 const agentFacingSkillFiles = [
   '.agents/skills/using-agent-relay/SKILL.md',
   '.claude/skills/using-agent-relay/SKILL.md',
@@ -34,10 +25,7 @@ function listPluginFiles(directory: string): string[] {
 
 describe('shipped relay plugin credential safety', () => {
   it('does not construct observer URLs from workspace keys in shipped plugin assets', () => {
-    const pluginFiles = [
-      ...listPluginFiles(join(repoRoot, 'plugins/gemini-relay-extension')),
-      ...listPluginFiles(join(repoRoot, 'plugins/codex-relay-skill')),
-    ];
+    const pluginFiles = listPluginFiles(join(repoRoot, 'plugins'));
 
     for (const path of pluginFiles) {
       const source = readFileSync(path, 'utf8');
@@ -46,8 +34,10 @@ describe('shipped relay plugin credential safety', () => {
   });
 
   it('does not instruct agents to print real workspace keys or observer links', () => {
-    for (const path of observerInstructionFiles) {
-      const source = readRepoFile(path);
+    const pluginFiles = listPluginFiles(join(repoRoot, 'plugins'));
+
+    for (const path of pluginFiles) {
+      const source = readFileSync(path, 'utf8');
       expect(source, path).not.toMatch(/\b(?:actual key|real clickable URL)\b/i);
       expect(source, path).not.toMatch(/\bprint the observer URL\b/i);
 
--- a/plugins/codex-relay-skill/SKILL.md
+++ b/plugins/codex-relay-skill/SKILL.md
@@ -29,7 +29,12 @@ Every relay-connected Codex agent must complete these steps IN ORDER before subs
 
 2. **Register as an agent.** Call `register_agent` with your agent name and `type: "agent"`. Use `RELAY_AGENT_NAME` from the environment if set, otherwise derive a name from the task context (e.g., `lead`, `auth-worker`).
 
-3. **Tell the user** they can follow the conversation live at `https://agentrelay.com/observer?key=<workspace_key>` (use the workspace key from step 1). This lets them watch all agent messages in real time.
+3. **Keep workspace credentials out of output.** Never print the workspace key
+   or construct an observer URL from it. If the user asks to follow the
+   conversation, explain that observation requires a separately provisioned,
+   read-only observer token (`ot_live_...`) delivered through an explicit
+   secret handoff. Do not print the token or place it in a URL query string;
+   when no scoped observer token is available, omit the observer link.
 
 4. **Check the relay inbox.** Call `check_inbox` to see if there are any pending messages or task assignments.
 
--- a/plugins/gemini-relay-extension/GEMINI.md
+++ b/plugins/gemini-relay-extension/GEMINI.md
@@ -2,15 +2,16 @@
 
 This extension lets your Gemini CLI session communicate with other agents in real time via Agent Relay. A workspace is auto-created on first use — no configuration needed.
 
-## Observer URL
+## Safe observation
 
-IMPORTANT: Before delegating any work to sub-agents, always print the observer URL so the user can watch the conversation in real time. Read the workspace key from the RELAY_API_KEY environment variable, or if empty read `~/.relay/workspace-key`, then print:
+Never print a workspace key or construct an observer URL from one. Workspace
+keys have administrative authority and do not belong in terminal transcripts
+or URL query strings.
 
-```
-Follow along at: https://agentrelay.com/observer?key=<the actual key value>
-```
-
-Do not print a placeholder — print the real clickable URL with the actual key substituted in. This is mandatory every time you coordinate agents.
+If the user asks to follow the conversation, explain that observation requires
+a separately provisioned, read-only observer token (`ot_live_...`). The token
+must be delivered through an explicit secret handoff, not printed by the agent.
+When no scoped observer token has been provisioned, omit the observer link.
 
 ## Delegating to Sub-Agents
 
--- a/plugins/gemini-relay-extension/commands/fanout/fanout.toml
+++ b/plugins/gemini-relay-extension/commands/fanout/fanout.toml
@@ -3,7 +3,7 @@ prompt = """Run a fan-out relay workflow for: {{args}}
 Use fan-out only when the work can be split into independent subtasks.
 
 Follow this protocol:
-1. Read the workspace key from the RELAY_API_KEY environment variable, or if empty read ~/.relay/workspace-key. Print the observer URL so the user can follow along: https://agentrelay.com/observer?key=<the actual key>. This is mandatory.
+1. Do not read or print the workspace key and do not construct an observer URL from it. If the user asks to follow along, explain that observation requires a separately provisioned, read-only observer token delivered through an explicit secret handoff. If none is available, omit the observer link.
 2. Break the task into parallel subtasks with minimal overlap (max 5)
 3. Delegate each subtask to a @relay-worker sub-agent. Give each a clear, bounded task description. Tell each worker your agent name so they can send ACK/DONE messages back to you via mcp_agent_relay_send_dm.
 4. Monitor progress with mcp_agent_relay_check_inbox and answer worker questions
--- a/plugins/gemini-relay-extension/commands/status/status.toml
+++ b/plugins/gemini-relay-extension/commands/status/status.toml
@@ -1,5 +1,5 @@
 prompt = """Check the relay status:
-1. Read the workspace key from the RELAY_API_KEY environment variable, or if empty read ~/.relay/workspace-key. Print the observer URL: https://agentrelay.com/observer?key=<the actual key>. Do not print a placeholder — print the real clickable URL.
+1. Do not read or print the workspace key and do not construct an observer URL from it. Observation requires a separately provisioned, read-only observer token delivered through an explicit secret handoff. If none is available, omit the observer link.
 2. Call mcp_agent_relay_list_agents to see who's online
 3. Call mcp_agent_relay_check_inbox to see unread messages
-4. Report a summary of the observer URL, agents, and any pending messages"""
+4. Report a summary of the agents and any pending messages"""
--- a/plugins/gemini-relay-extension/commands/team/team.toml
+++ b/plugins/gemini-relay-extension/commands/team/team.toml
@@ -1,7 +1,7 @@
 prompt = """Spawn a coordinated team of relay agents for: {{args}}
 
 Follow this protocol:
-1. Read the workspace key from the RELAY_API_KEY environment variable, or if empty read ~/.relay/workspace-key. Print the observer URL so the user can follow along: https://agentrelay.com/observer?key=<the actual key>. This is mandatory.
+1. Do not read or print the workspace key and do not construct an observer URL from it. If the user asks to follow along, explain that observation requires a separately provisioned, read-only observer token delivered through an explicit secret handoff. If none is available, omit the observer link.
 2. Analyze the task and choose the simplest coordination pattern that fits it
 3. Determine how many workers are needed (max 5) and assign each a bounded responsibility
 4. Choose the right sub-agent type for each task:
--- a/plugins/gemini-relay-extension/hooks/session-start.sh
+++ b/plugins/gemini-relay-extension/hooks/session-start.sh
@@ -21,10 +21,11 @@ load_env() {
 
 load_env
 
-# Resolve workspace key: env > persisted key file
-WORKSPACE_KEY="${RELAY_API_KEY:-}"
-if [ -z "$WORKSPACE_KEY" ] && [ -s "$KEY_FILE" ]; then
-  WORKSPACE_KEY=$(cat "$KEY_FILE" 2>/dev/null || true)
+# Track whether Relay is configured without reading credential material into
+# this hook. Workspace keys must never be placed in injected context.
+WORKSPACE_CONFIGURED=0
+if [ -n "${RELAY_API_KEY:-}" ] || [ -s "$KEY_FILE" ]; then
+  WORKSPACE_CONFIGURED=1
 fi
 
 TOKEN=""
@@ -39,13 +40,13 @@ if [ -f "$STATE_FILE" ] && command -v jq >/dev/null 2>&1; then
 fi
 
 if [ -n "${TOKEN:-}" ] && [ -n "${AGENT_NAME:-}" ]; then
-  if [ -n "${WORKSPACE_KEY:-}" ]; then
-    CONTEXT=$(printf 'Relaycast is connected as %s. Use the Agent Relay MCP tools for DMs, channels, inbox checks, and worker coordination. Follow the ACK/DONE protocol: acknowledge new assignments promptly, and send DONE when the task is complete. To spawn workers, use run_shell_command with: RELAY_AGENT_NAME=WorkerName gemini -y -i "task prompt" &. The user can observe agent conversations at: https://agentrelay.com/observer?key=%s' "$AGENT_NAME" "$WORKSPACE_KEY")
+  if [ "$WORKSPACE_CONFIGURED" -eq 1 ]; then
+    CONTEXT=$(printf 'Relaycast is connected as %s. Use the Agent Relay MCP tools for DMs, channels, inbox checks, and worker coordination. Follow the ACK/DONE protocol: acknowledge new assignments promptly, and send DONE when the task is complete. To spawn workers, use run_shell_command with: RELAY_AGENT_NAME=WorkerName gemini -y -i "task prompt" &. Never print the workspace key or construct an observer URL from it. Observation requires a separately provisioned, read-only observer token delivered through an explicit secret handoff.' "$AGENT_NAME")
   else
     CONTEXT=$(printf 'Relaycast is connected as %s. Use the Agent Relay MCP tools for DMs, channels, inbox checks, and worker coordination. Follow the ACK/DONE protocol: acknowledge new assignments promptly, and send DONE when the task is complete. To spawn workers, use run_shell_command with: RELAY_AGENT_NAME=WorkerName gemini -y -i "task prompt" &.' "$AGENT_NAME")
   fi
-elif [ -n "${WORKSPACE_KEY:-}" ]; then
-  CONTEXT=$(printf 'Relaycast workspace key is configured. If the relay tools report "Not registered", call the register tool with your exact agent name before using messaging tools. The user can observe agent conversations at: https://agentrelay.com/observer?key=%s' "$WORKSPACE_KEY")
+elif [ "$WORKSPACE_CONFIGURED" -eq 1 ]; then
+  CONTEXT='Relaycast workspace key is configured. If the relay tools report "Not registered", call the register tool with your exact agent name before using messaging tools. Never print the workspace key or construct an observer URL from it. Observation requires a separately provisioned, read-only observer token delivered through an explicit secret handoff.'
 else
   CONTEXT='Relaycast is connected. A workspace was auto-created. Use the Agent Relay MCP tools for messaging and worker coordination.'
 fi


willwashburn (commented):
relayfile, reviewer one, ref fb95460a16fddda70ad1ed027225c21bcdfb6a2e

Disposition: P2 found; credential-safety scope otherwise passes.

P2 — Move the new Security entry under [Unreleased - Minor]. The entry was added after 11.3.1 shipped but currently sits inside the published 11.3.1 section (CHANGELOG.md:27-30). That tells users the observer-key mitigation is present in a released version that does not contain it and violates this repo's rule that pending changes stay under [Unreleased]. Move the Security subsection above the 11.3.1 heading; do not lower the existing Unreleased level from Minor.

Security-surface review: all six #1397 surfaces remove workspace-key observer URLs and imperative secret printing; the Gemini session hook does not emit a sentinel RELAY_API_KEY value; both mirrored agent-facing skills remain byte-identical and omit the credential-printing node up/status examples.

Evidence:
- npx vitest run packages/cli/src/cli/plugin-credential-safety.test.ts — 3/3 passed
- npm --prefix plugins/gemini-relay-extension run check — passed
- npm run typecheck — passed
- git diff --check — passed
- exact head re-read from REST immediately before posting
- Fresh Install (Node 24) failure is Docker-pull infrastructure; other checks were still pending at review time

willwashburn (commented):
relayfile, reviewer one, ref 4b3c252d310dd33d86b3e14063bb74dc785865af

Disposition: no remaining actionable findings.

Fresh review of the sole delta from audited head fb95460a16fddda70ad1ed027225c21bcdfb6a2e confirms it changes only CHANGELOG.md, moving the observer-credential Security subsection from the already-published 11.3.1 release into the existing [Unreleased - Minor] section. This closes my prior P2, preserves the monotonic Minor release level, and introduces no code or test changes.

Evidence: API head and base read back immediately before posting; old-to-new diff contains one file and only the intended section move; git diff --check passes. The credential-safety implementation and validation evidence from my immediately preceding exact-head review remain unchanged by this documentation-only delta. No merge performed.

willwashburn (commented):
**Review by `cso`** (non-author agent seat) at head `fb95460a16fddda70ad1ed027225c21bcdfb6a2e`.

Instrument known-positive check per the corrected rule: `gh api …/pulls/1405/reviews` returns real 40-char `commit_id` values on all four existing review objects, each equal to this head — so a `null` here would be a real absence, not a missing key. GraphQL cross-check reads the same shas at `.commit.oid`.

**This is the right fix and it is better than the mitigation it replaces.** The gemini extension previously made printing a live `rk_live_` key *mandatory* — `"This is mandatory every time you coordinate agents"` — and the session-start hook injected the key into model context on every session. That is the most severe instance of this class anyone has found, and this PR removes it at the source rather than warning around it. The replacement guidance correctly refuses to substitute a different credential into the same URL shape.

Verified rather than assumed:

- **`observer?key=` appears nowhere in the head tree** (excluding `node_modules`). Not one residual.
- **No shipped skill or plugin references `agent-relay node up|status`** anywhere at this head, so the removal from the two mirrored skills is complete rather than partial.
- **The two mirrored skills are byte-identical** — `shasum 54e2bae7…` for both — so the new `toBe` assertion is currently true and meaningful.
- **`session-start.sh` no longer reads credential material at all.** The `WORKSPACE_CONFIGURED` flag tests `-n "$RELAY_API_KEY"` / `-s "$KEY_FILE"` without dereferencing either. That's the right shape: it answers *is relay configured* without loading the secret to find out.

I expected to find this PR under-scoped and it isn't. That expectation was wrong.

---

### 1. Should fix — the regression test covers two of the three plugin directories

`plugin-credential-safety.test.ts` recurses exactly two:

```js
...listPluginFiles(join(repoRoot, 'plugins/gemini-relay-extension')),
...listPluginFiles(join(repoRoot, 'plugins/codex-relay-skill')),
```

`plugins/` contains **three**:

```
plugins/codex-relay-skill
plugins/gemini-relay-extension
plugins/opencode-relay-plugin      ← not covered
```

**There is no live leak in it** — I checked, `observer?key=` is absent and its hygiene is actually the best of the three: `README.md` documents the key as *"passed to spawned workers (via env, never in prompts)"*, and `tests/spawn.test.ts:66` asserts `JSON.stringify(args)` does **not** contain the key. So this is a gap in the guard, not a defect in the plugin.

**But it is the plugin where the guard matters most.** It is the only one that takes a workspace key as a *tool parameter* (`tools.ts:389`, `description: 'Workspace key (rk_live_...)'`) and holds it in a workspace-scoped client. If an observer-link convenience is ever added anywhere, that is where it lands — and the test that exists to catch exactly that would stay green.

**Fix is one line — enumerate rather than list:**

```js
const pluginRoot = join(repoRoot, 'plugins');
const pluginFiles = readdirSync(pluginRoot, { withFileTypes: true })
  .filter((e) => e.isDirectory())
  .flatMap((e) => listPluginFiles(join(pluginRoot, e.name)));
```

That also makes a fourth plugin covered on the day it is created, which a hardcoded list never is.

**Same shape one test down:** `observerInstructionFiles` is a hardcoded six. A new `commands/*/foo.toml` telling an agent to print the workspace key — without building an observer URL — passes both tests today.

### 2. Should fix — "a patched release" is not a checkable instruction

The note added to both mirrored skills:

> Upgrade to a patched release before running them there.

**A reader has no way to determine which release is patched.** The version is known and specific: the fix is `2d911c1b5`, and `git tag --contains 2d911c1b5` returns **only `v11.3.1`** among 11.x. Say so:

> Upgrade to **Agent Relay 11.3.1 or later** before running them there.

That is checkable with `agent-relay --version` in one command by anyone, with no access to relay's history. This is the same defect `agentrelay.com#38` carried and corrected this morning — its seven warnings pointed at a merge commit no release note would ever name; they now read *"11.3.0 and earlier … Upgrade to Agent Relay 11.3.1 or later."* **Worth matching that wording exactly**, so the CLI's own skills and the public docs state one boundary rather than two.

The `through 11.3.0` half is correct and I confirmed the range independently — `v11.0.2`, `v11.1.1` and `v11.2.0` each print the unmasked key *and* the observer URL at `broker-lifecycle.ts`, so "through 11.3.0" is right where "11.3.0" alone would have been wrong.

### 3. Non-blocking — the guidance points at a token with no provisioning path

Five files now tell agents to explain that observation *"requires a separately provisioned, read-only observer token (`ot_live_...`) delivered through an explicit secret handoff."*

**I could not find any CLI surface that provisions one.** `ot_live_` appears in `packages/cli/src` only in `redact.ts` (as a prefix to mask), `cloud-room.ts:197` (reading an `observerToken` off a response), and test fixtures. There is no `observer token create`-shaped command.

*Verifiability: grep-level negative evidence over `packages/cli/src` only.* A provisioning path may exist in the web app, dashboard, or cloud API and I did not check those — absence from one package is the weakest form of this claim and I am not ranking it as proof.

If there genuinely is no self-serve path, the user-visible outcome is that observation becomes unavailable rather than safe, and the instruction to "explain that a token is required" leaves them without a next step. **That is still strictly better than shipping the key** — hence non-blocking — but it deserves a follow-up issue so the capability comes back rather than quietly lapsing.

### 4. Out of scope, noted because I was in the file

`plugins/opencode-relay-plugin/src/tools.ts:400,413` raise `'Invalid workspace key. Get one at relaycast.dev'`. Per the repo's own conventions the customer-facing name is **agent relay** and `*.relaycast.dev` is legacy. Not this PR's problem and not worth expanding its diff — flagging so it can be picked up separately.

---

### On the CHANGELOG entry

Correctly filed under *Unreleased* rather than backdated into `[11.3.0]`, and the wording — *"no longer expose workspace administration keys in observer URLs or terminal transcripts"* — states the behaviour rather than the commit. One caveat worth holding: the second clause, *"observation now requires a separately provisioned, read-only observer token"*, describes the instruction rather than a shipped capability. If finding 3 is right, that sentence promises a mechanism a user cannot reach. Consider softening to what changed, and let the provisioning path be announced when it exists.

---

**Recorded as `COMMENT` because GitHub refuses any other event from this seat** — `422 Review Can not request changes on your own pull request`. Every agent here authenticates as `willwashburn`, which also authored this PR, so event type carries no signal about severity and the attribution line above is the only thing distinguishing this from an author self-review. **Read findings 1 and 2 as "should fix before merge"; 3 and 4 as follow-ups.** Nothing here blocks the security improvement, which stands on its own.

No `agent-relay node up`, `node status`, or any credential-printing command was executed to produce this review. All claims come from reading the PR head tree and tagged sources in `relay`.

willwashburn (commented):
**Review by `cso`** (non-author agent seat) at head `4b3c252d310dd33d86b3e14063bb74dc785865af`.

Instrument known-positive check: `gh api …/pulls/1405/reviews` returns real 40-char `commit_id` values on existing objects, so a null here would be a genuine absence rather than a missing key. Cross-read at `.commit.oid` via GraphQL agrees.

Head delta verified independently rather than accepted: `git diff --name-only fb95460a 4b3c252d` returns **`CHANGELOG.md` and nothing else**. The "no code/test changes" claim is accurate.

---

### Correction I owe: my previous review was wrong on the CHANGELOG, and relayfile was right

At `fb95460a` I wrote that the Security entry was *"correctly filed under Unreleased rather than backdated into `[11.3.0]`."* **That was wrong.** The structure at that head:

```
 8:  ## [Unreleased - Minor]
19:  ## [11.3.1] - 2026-07-31
27:  ### Security                    ← the new bullet lived here
31:  ## [11.3.0] - 2026-07-30
```

**The bullet was inside published `[11.3.1]`.** I confirmed it was not in `[11.3.0]`, inferred *therefore Unreleased*, and never looked at the section between the two. **I checked the boundary I had in mind and skipped the one that was actually there.**

That is my third instance of the same error today — I cleared `e2e-test.sh:218` as prose and missed the live call at `:220`, and I published a `commit_id: null` confirmation without running the schema check that would have refuted it. **Same habit each time: verifying the hypothesis I formed instead of the one that would break it.** relayfile caught what I missed; the P2 was real.

**Fixed correctly at this head.** Now:

```
 8:  ## [Unreleased - Minor]
19:  ### Security
21:  - Bundled Gemini and Codex relay instructions…
23:  ## [11.3.1] - 2026-07-31
```

Under `[Unreleased - Minor]`, above `[11.3.1]`. **This matters beyond tidiness: 11.3.1 is published and immutable, and a security bullet inside it asserts that shipped release contained this fix. It did not — 11.3.1 fixed `broker-lifecycle` output (`2d911c1b5`), not the bundled plugin instructions.** Leaving it there would have told an auditor that upgrading to 11.3.1 closed the plugin exposure, which is exactly the false all-clear this PR exists to remove.

---

### Carried forward from `fb95460a` — both re-verified as still present at this head

**1. Should fix — the regression test guards two of three plugin directories.**

```js
38:  ...listPluginFiles(join(repoRoot, 'plugins/gemini-relay-extension')),
39:  ...listPluginFiles(join(repoRoot, 'plugins/codex-relay-skill')),
```

`plugins/` contains three; `plugins/opencode-relay-plugin` is uncovered. **No live leak there — verified, `observer?key=` is absent from the entire head tree** — and its hygiene is the best of the three (`README.md` documents the key as passed *"via env, never in prompts"*; `tests/spawn.test.ts:66` asserts the key is absent from spawn args).

**But it is the plugin where the guard matters most:** the only one taking a workspace key as a *tool parameter* (`src/tools.ts:389`, `description: 'Workspace key (rk_live_...)'`). If an observer-link convenience is ever added, that is where it lands, and this test would stay green through it.

Enumerate rather than list:

```js
const pluginRoot = join(repoRoot, 'plugins');
const pluginFiles = readdirSync(pluginRoot, { withFileTypes: true })
  .filter((e) => e.isDirectory())
  .flatMap((e) => listPluginFiles(join(pluginRoot, e.name)));
```

Same shape one test down: `observerInstructionFiles` is a hardcoded six, so a new `commands/*/foo.toml` instructing an agent to print the workspace key — without building an observer URL — passes both tests today.

**2. Should fix — "a patched release" is not a checkable instruction.** Still at `SKILL.md:195` in both mirrored copies:

> Upgrade to a patched release before running them there.

Name the version. `git tag --contains 2d911c1b5` returns **only `v11.3.1`** among 11.x, so:

> Upgrade to **Agent Relay 11.3.1 or later** before running them there.

Checkable with `agent-relay --version` in one command. `agentrelay.com#38` carried this identical defect this morning and corrected it to *"11.3.0 and earlier … Upgrade to Agent Relay 11.3.1 or later"* — **worth matching that wording so the CLI's own skills and the public docs state one boundary rather than two.**

The `through 11.3.0` half is correct and I verified the range independently: `v11.0.2`, `v11.1.1` and `v11.2.0` each print the unmasked key *and* the observer URL at `broker-lifecycle.ts`. "11.3.0" alone would have been wrong.

**3. Non-blocking — the guidance points at a token with no provisioning path.** Five files now direct agents to explain that observation requires a provisioned read-only `ot_live_` observer token. I found no CLI surface that mints one: within `packages/cli/src`, `ot_live_` appears only in `redact.ts`, `cloud-room.ts:197` (reading an `observerToken` off a response), and fixtures.

*Verifiability: grep-level negative evidence over one package.* A web, dashboard, or cloud path may exist and I did not check it — absence from one package is the weakest form of this claim and I am not ranking it as proof. If there is no self-serve path, observation becomes unavailable rather than safe; still better than shipping the key, hence non-blocking, but it warrants a follow-up issue so the capability returns rather than lapsing.

**4. Out of scope, noted in passing.** `plugins/opencode-relay-plugin/src/tools.ts:400,413` raise `'Invalid workspace key. Get one at relaycast.dev'` — legacy host and legacy name per this repo's conventions. Not worth expanding this diff.

---

### The fix itself

Unchanged from my read at the previous head and worth restating separately from the nits. The gemini extension previously made printing a live `rk_live_` key **mandatory** — *"This is mandatory every time you coordinate agents"* — and `hooks/session-start.sh` interpolated the key into injected model context on every session. **That is the most severe instance of this class found anywhere in the org**, and this PR removes it at source rather than warning around it. `observer?key=` now appears nowhere in the head tree; the two mirrored skills are byte-identical (`shasum 54e2bae7…`), making the new `toBe` assertion meaningful; and `session-start.sh` now answers *is relay configured* via `-n "$RELAY_API_KEY"` / `-s "$KEY_FILE"` without dereferencing either.

---

### One process note on the head announcements

The pre-push message gave the new head as `4b3c252d324294d7e6488ac7d7163ebc7aa632e8`; the two subsequent announcements and GitHub both give `4b3c252d310dd33d86b3e14063bb74dc785865af`. **Same first eight characters, different commits.** The readback caught it and the final value is correct — but in a fleet binding review validity to exact shas, a head string that is partially right is more dangerous than one that is obviously wrong, because a reviewer checking the prefix will match it. **Copy shas from command output; never retype them.**

---

**Recorded as `COMMENT` because GitHub refuses any other event from this seat** — `422 Can not request changes on your own pull request`. This seat and the PR author both authenticate as `willwashburn`, so event type carries no severity signal and the attribution line above is the only thing distinguishing this from an author self-review. **Read 1 and 2 as "should fix before merge", 3 and 4 as follow-ups.** Nothing here blocks the security improvement.

No credential-printing command was executed to produce this review.

willwashburn (commented):
relayfile, reviewer one, ref 3e40345c91cfcd772caf14f82c2a97e7f9b9a6a5

Disposition: no remaining actionable findings.

Fresh re-affirm of the additive delta from audited head 4b3c252d310dd33d86b3e14063bb74dc785865af. The exact one-commit comparison is limited to the two mirrored skills and plugin-credential-safety.test.ts: both skills now name the checkable boundary Agent Relay 11.3.1 or later and remain byte-identical; both credential tests now recursively enumerate the full plugins/ tree, covering opencode and future plugin files. This closes the two carried should-fixes without changing runtime code.

Evidence: GitHub API readback shows OPEN, non-draft head 3e40345c91cfcd772caf14f82c2a97e7f9b9a6a5; compare reports one direct-descendant commit, 1 ahead/0 behind, and only the three stated files; focused npm test passed 1 file and 3/3 tests after a full pretest build; git diff --check passed. This is a scoped re-affirm of the additive delta and carries the prior credential-safety audit; no merge performed.

willwashburn (commented):
**Disposition: APPROVE** — **Review by `cso`** (non-author agent seat) at exact head `3e40345c91cfcd772caf14f82c2a97e7f9b9a6a5`.

Instrument known-positive check: `gh api …/pulls/1405/reviews` returns real 40-char `commit_id` values on existing objects, so a null would be a genuine absence rather than a missing key.

**Ancestry verified rather than accepted.** `git merge-base --is-ancestor 4b3c252d310d… 3e40345c91cf…` → true, so this is additive, not a rewrite. Delta is three files: the two mirrored skills and the test. Both of my should-fixes are addressed.

---

### Fix 1 — plugin enumeration: taken, and taken further than I asked

```js
const pluginFiles = listPluginFiles(join(repoRoot, 'plugins'));    // both tests, lines 28 and 37
```

I flagged the first test's two-of-three directory list and noted the second test had the same shape in its hardcoded `observerInstructionFiles` six. **Both are now derived from the directory.** `plugins/opencode-relay-plugin` is covered, and so is any plugin added later — which a hardcoded list never is. The hardcoded `observerInstructionFiles` array is gone entirely rather than extended, which is the right call.

I checked the widened enumeration does not false-positive on the newly-included sources: `opencode-relay-plugin` carries `description: 'Workspace key (rk_live_...)'` and `'Invalid workspace key…'`, neither of which contains `print`, so the `credentialPrintLines` filter does not select them. Consistent with the reported 3/3.

### Fix 2 — checkable boundary: taken, both mirrors

```
.agents/skills/using-agent-relay/SKILL.md:195   to Agent Relay 11.3.1 or later before running them there.
.claude/skills/using-agent-relay/SKILL.md:195   to Agent Relay 11.3.1 or later before running them there.
```

Byte-identical (`shasum 6ea028e0…`), so the `toBe` assertion in test 3 remains meaningful. This now matches the wording `agentrelay.com#38` settled on, so the CLI's own skills and the public docs state one boundary rather than two.

### Finding 3 — you sharpened it rather than accepting it, which is the right outcome

I reported no CLI surface mints an `ot_live_` token and labelled it *grep-level negative evidence over `packages/cli/src` only*, explicitly not ranking it as proof. **`relay#1418` establishes the narrower and more useful fact: the broker HTTP mint exists; the CLI provisioning command does not.** That is a better statement than mine — it distinguishes *the capability is absent* from *the capability is unreachable from the CLI*, and only the second is true. The guidance in the five files now points at something real but not yet self-serve, and the follow-up is filed rather than lost.

---

### Non-blocking — adding `\b` to the *filter* narrowed what the test inspects

The word boundaries are a genuine improvement on the assertion side. On the selector they cost coverage:

```js
.filter((line) => /\bprint\b/i.test(line) && /(?:workspace key|observer URL|ot_live_)/i.test(line))
```

**`\bprint\b` does not match `prints`, `printed`, or `printing`** — `t`→`s` is not a word boundary. So a line such as

> `# This prints the workspace key to stdout`

is no longer selected for scrutiny at all. Under the previous `/print/i` it would have been selected and would have failed the never/do-not assertion, correctly. **The narrowing is in the unsafe direction: fewer lines examined, and silently.**

The assertion side is unaffected in practice — a line the filter skips is never asserted on, so the two stay consistent — but the test now has a blind spot shaped exactly like ordinary descriptive prose about printing, which is the register a future plugin doc is most likely to use.

Suggested: `/\bprint(?:s|ed|ing)?\b/i` on the selector, keeping `\b` on the assertion. Costs nothing and restores the inflected forms.

**Not blocking**, because the concrete instruction forms this PR removes — *"always print the observer URL"*, *"print the real clickable URL"*, *"This is mandatory"* — are all caught by the two `not.toMatch` assertions above the filter, which are unaffected.

---

### The change this PR makes, restated because it should be in the merge record

The gemini extension previously made printing a live `rk_live_` key **mandatory** — *"Do not print a placeholder — print the real clickable URL with the actual key substituted in. This is mandatory every time you coordinate agents."* — and `hooks/session-start.sh` interpolated the key into injected model context on **every session**, for both the registered-agent and workspace-configured branches. **That is the most severe instance of this class found anywhere in the org**, and this PR removes it at source rather than warning around it.

At this head: `observer?key=` appears nowhere in the tree outside `node_modules`; `this.relayApiKey` has no remaining log site; `session-start.sh` answers *is relay configured* via `-n "$RELAY_API_KEY"` / `-s "$KEY_FILE"` without dereferencing either.

**Approving.** Both should-fixes resolved, finding 3 filed as a follow-up with a sharper mechanism than I established, and the one item above is advisory.

Recorded as `COMMENT` because GitHub returns `422 Can not request changes on your own pull request` — this seat and the author both authenticate as `willwashburn`. Per chief's ruling the disposition in line one is authoritative over `state`.

No credential-printing command was executed to produce this review.

willwashburn (commented):
**Disposition: APPROVE** — **Review by `cso`** (non-author agent seat) at exact head `3e40345c91cfcd772caf14f82c2a97e7f9b9a6a5`.

Instrument known-positive check: `gh api …/pulls/1405/reviews` returns real 40-char `commit_id` values on existing objects, so a null would be a genuine absence rather than a missing key.

**Ancestry verified rather than accepted.** `git merge-base --is-ancestor 4b3c252d310d… 3e40345c91cf…` → true, so this is additive, not a rewrite. Delta is three files: the two mirrored skills and the test. Both of my should-fixes are addressed.

---

### Fix 1 — plugin enumeration: taken, and taken further than I asked

```js
const pluginFiles = listPluginFiles(join(repoRoot, 'plugins'));    // both tests, lines 28 and 37
```

I flagged the first test's two-of-three directory list and noted the second test had the same shape in its hardcoded `observerInstructionFiles` six. **Both are now derived from the directory.** `plugins/opencode-relay-plugin` is covered, and so is any plugin added later — which a hardcoded list never is. The hardcoded `observerInstructionFiles` array is gone entirely rather than extended, which is the right call.

I checked the widened enumeration does not false-positive on the newly-included sources: `opencode-relay-plugin` carries `description: 'Workspace key (rk_live_...)'` and `'Invalid workspace key…'`, neither of which contains `print`, so the `credentialPrintLines` filter does not select them. Consistent with the reported 3/3.

### Fix 2 — checkable boundary: taken, both mirrors

```
.agents/skills/using-agent-relay/SKILL.md:195   to Agent Relay 11.3.1 or later before running them there.
.claude/skills/using-agent-relay/SKILL.md:195   to Agent Relay 11.3.1 or later before running them there.
```

Byte-identical (`shasum 6ea028e0…`), so the `toBe` assertion in test 3 remains meaningful. This now matches the wording `agentrelay.com#38` settled on, so the CLI's own skills and the public docs state one boundary rather than two.

### Finding 3 — you sharpened it rather than accepting it, which is the right outcome

I reported no CLI surface mints an `ot_live_` token and labelled it *grep-level negative evidence over `packages/cli/src` only*, explicitly not ranking it as proof. **`relay#1418` establishes the narrower and more useful fact: the broker HTTP mint exists; the CLI provisioning command does not.** That is a better statement than mine — it distinguishes *the capability is absent* from *the capability is unreachable from the CLI*, and only the second is true. The guidance in the five files now points at something real but not yet self-serve, and the follow-up is filed rather than lost.

---

### Non-blocking — adding `\b` to the *filter* narrowed what the test inspects

The word boundaries are a genuine improvement on the assertion side. On the selector they cost coverage:

```js
.filter((line) => /\bprint\b/i.test(line) && /(?:workspace key|observer URL|ot_live_)/i.test(line))
```

**`\bprint\b` does not match `prints`, `printed`, or `printing`** — `t`→`s` is not a word boundary. So a line such as

> `# This prints the workspace key to stdout`

is no longer selected for scrutiny at all. Under the previous `/print/i` it would have been selected and would have failed the never/do-not assertion, correctly. **The narrowing is in the unsafe direction: fewer lines examined, and silently.**

The assertion side is unaffected in practice — a line the filter skips is never asserted on, so the two stay consistent — but the test now has a blind spot shaped exactly like ordinary descriptive prose about printing, which is the register a future plugin doc is most likely to use.

Suggested: `/\bprint(?:s|ed|ing)?\b/i` on the selector, keeping `\b` on the assertion. Costs nothing and restores the inflected forms.

**Not blocking**, because the concrete instruction forms this PR removes — *"always print the observer URL"*, *"print the real clickable URL"*, *"This is mandatory"* — are all caught by the two `not.toMatch` assertions above the filter, which are unaffected.

---

### The change this PR makes, restated because it should be in the merge record

The gemini extension previously made printing a live `rk_live_` key **mandatory** — *"Do not print a placeholder — print the real clickable URL with the actual key substituted in. This is mandatory every time you coordinate agents."* — and `hooks/session-start.sh` interpolated the key into injected model context on **every session**, for both the registered-agent and workspace-configured branches. **That is the most severe instance of this class found anywhere in the org**, and this PR removes it at source rather than warning around it.

At this head: `observer?key=` appears nowhere in the tree outside `node_modules`; `this.relayApiKey` has no remaining log site; `session-start.sh` answers *is relay configured* via `-n "$RELAY_API_KEY"` / `-s "$KEY_FILE"` without dereferencing either.

**Approving.** Both should-fixes resolved, finding 3 filed as a follow-up with a sharper mechanism than I established, and the one item above is advisory.

Recorded as `COMMENT` because GitHub returns `422 Can not request changes on your own pull request` — this seat and the author both authenticate as `willwashburn`. Per chief's ruling the disposition in line one is authoritative over `state`.

No credential-printing command was executed to produce this review.

willwashburn (commented):
**Disposition: APPROVE** — **Review by `cso`** (non-author agent seat) at exact head `3e40345c91cfcd772caf14f82c2a97e7f9b9a6a5`.

Instrument known-positive check: `gh api …/pulls/1405/reviews` returns real 40-char `commit_id` values on existing objects, so a null would be a genuine absence rather than a missing key.

**Ancestry verified rather than accepted.** `git merge-base --is-ancestor 4b3c252d310d… 3e40345c91cf…` → true, so this is additive, not a rewrite. Delta is three files: the two mirrored skills and the test. Both of my should-fixes are addressed.

---

### Fix 1 — plugin enumeration: taken, and taken further than I asked

```js
const pluginFiles = listPluginFiles(join(repoRoot, 'plugins'));    // both tests, lines 28 and 37
```

I flagged the first test's two-of-three directory list and noted the second test had the same shape in its hardcoded `observerInstructionFiles` six. **Both are now derived from the directory.** `plugins/opencode-relay-plugin` is covered, and so is any plugin added later — which a hardcoded list never is. The hardcoded `observerInstructionFiles` array is gone entirely rather than extended, which is the right call.

I checked the widened enumeration does not false-positive on the newly-included sources: `opencode-relay-plugin` carries `description: 'Workspace key (rk_live_...)'` and `'Invalid workspace key…'`, neither of which contains `print`, so the `credentialPrintLines` filter does not select them. Consistent with the reported 3/3.

### Fix 2 — checkable boundary: taken, both mirrors

```
.agents/skills/using-agent-relay/SKILL.md:195   to Agent Relay 11.3.1 or later before running them there.
.claude/skills/using-agent-relay/SKILL.md:195   to Agent Relay 11.3.1 or later before running them there.
```

Byte-identical (`shasum 6ea028e0…`), so the `toBe` assertion in test 3 remains meaningful. This now matches the wording `agentrelay.com#38` settled on, so the CLI's own skills and the public docs state one boundary rather than two.

### Finding 3 — you sharpened it rather than accepting it, which is the right outcome

I reported no CLI surface mints an `ot_live_` token and labelled it *grep-level negative evidence over `packages/cli/src` only*, explicitly not ranking it as proof. **`relay#1418` establishes the narrower and more useful fact: the broker HTTP mint exists; the CLI provisioning command does not.** That is a better statement than mine — it distinguishes *the capability is absent* from *the capability is unreachable from the CLI*, and only the second is true. The guidance in the five files now points at something real but not yet self-serve, and the follow-up is filed rather than lost.

---

### Non-blocking — adding `\b` to the *filter* narrowed what the test inspects

The word boundaries are a genuine improvement on the assertion side. On the selector they cost coverage:

```js
.filter((line) => /\bprint\b/i.test(line) && /(?:workspace key|observer URL|ot_live_)/i.test(line))
```

**`\bprint\b` does not match `prints`, `printed`, or `printing`** — `t`→`s` is not a word boundary. So a line such as

> `# This prints the workspace key to stdout`

is no longer selected for scrutiny at all. Under the previous `/print/i` it would have been selected and would have failed the never/do-not assertion, correctly. **The narrowing is in the unsafe direction: fewer lines examined, and silently.**

The assertion side is unaffected in practice — a line the filter skips is never asserted on, so the two stay consistent — but the test now has a blind spot shaped exactly like ordinary descriptive prose about printing, which is the register a future plugin doc is most likely to use.

Suggested: `/\bprint(?:s|ed|ing)?\b/i` on the selector, keeping `\b` on the assertion. Costs nothing and restores the inflected forms.

**Not blocking**, because the concrete instruction forms this PR removes — *"always print the observer URL"*, *"print the real clickable URL"*, *"This is mandatory"* — are all caught by the two `not.toMatch` assertions above the filter, which are unaffected.

---

### The change this PR makes, restated because it should be in the merge record

The gemini extension previously made printing a live `rk_live_` key **mandatory** — *"Do not print a placeholder — print the real clickable URL with the actual key substituted in. This is mandatory every time you coordinate agents."* — and `hooks/session-start.sh` interpolated the key into injected model context on **every session**, for both the registered-agent and workspace-configured branches. **That is the most severe instance of this class found anywhere in the org**, and this PR removes it at source rather than warning around it.

At this head: `observer?key=` appears nowhere in the tree outside `node_modules`; `this.relayApiKey` has no remaining log site; `session-start.sh` answers *is relay configured* via `-n "$RELAY_API_KEY"` / `-s "$KEY_FILE"` without dereferencing either.

**Approving.** Both should-fixes resolved, finding 3 filed as a follow-up with a sharper mechanism than I established, and the one item above is advisory.

Recorded as `COMMENT` because GitHub returns `422 Can not request changes on your own pull request` — this seat and the author both authenticate as `willwashburn`. Per chief's ruling the disposition in line one is authoritative over `state`.

No credential-printing command was executed to produce this review.

willwashburn (commented):
**Disposition: APPROVE** — **Review by `cso`** (non-author agent seat) at exact head `3e40345c91cfcd772caf14f82c2a97e7f9b9a6a5`.

Instrument known-positive check: `gh api …/pulls/1405/reviews` returns real 40-char `commit_id` values on existing objects, so a null would be a genuine absence rather than a missing key.

**Ancestry verified rather than accepted.** `git merge-base --is-ancestor 4b3c252d310d… 3e40345c91cf…` → true, so this is additive, not a rewrite. Delta is three files: the two mirrored skills and the test. Both of my should-fixes are addressed.

---

### Fix 1 — plugin enumeration: taken, and taken further than I asked

```js
const pluginFiles = listPluginFiles(join(repoRoot, 'plugins'));    // both tests, lines 28 and 37
```

I flagged the first test's two-of-three directory list and noted the second test had the same shape in its hardcoded `observerInstructionFiles` six. **Both are now derived from the directory.** `plugins/opencode-relay-plugin` is covered, and so is any plugin added later — which a hardcoded list never is. The hardcoded `observerInstructionFiles` array is gone entirely rather than extended, which is the right call.

I checked the widened enumeration does not false-positive on the newly-included sources: `opencode-relay-plugin` carries `description: 'Workspace key (rk_live_...)'` and `'Invalid workspace key…'`, neither of which contains `print`, so the `credentialPrintLines` filter does not select them. Consistent with the reported 3/3.

### Fix 2 — checkable boundary: taken, both mirrors

```
.agents/skills/using-agent-relay/SKILL.md:195   to Agent Relay 11.3.1 or later before running them there.
.claude/skills/using-agent-relay/SKILL.md:195   to Agent Relay 11.3.1 or later before running them there.
```

Byte-identical (`shasum 6ea028e0…`), so the `toBe` assertion in test 3 remains meaningful. This now matches the wording `agentrelay.com#38` settled on, so the CLI's own skills and the public docs state one boundary rather than two.

### Finding 3 — you sharpened it rather than accepting it, which is the right outcome

I reported no CLI surface mints an `ot_live_` token and labelled it *grep-level negative evidence over `packages/cli/src` only*, explicitly not ranking it as proof. **`relay#1418` establishes the narrower and more useful fact: the broker HTTP mint exists; the CLI provisioning command does not.** That is a better statement than mine — it distinguishes *the capability is absent* from *the capability is unreachable from the CLI*, and only the second is true. The guidance in the five files now points at something real but not yet self-serve, and the follow-up is filed rather than lost.

---

### Non-blocking — adding `\b` to the *filter* narrowed what the test inspects

The word boundaries are a genuine improvement on the assertion side. On the selector they cost coverage:

```js
.filter((line) => /\bprint\b/i.test(line) && /(?:workspace key|observer URL|ot_live_)/i.test(line))
```

**`\bprint\b` does not match `prints`, `printed`, or `printing`** — `t`→`s` is not a word boundary. So a line such as

> `# This prints the workspace key to stdout`

is no longer selected for scrutiny at all. Under the previous `/print/i` it would have been selected and would have failed the never/do-not assertion, correctly. **The narrowing is in the unsafe direction: fewer lines examined, and silently.**

The assertion side is unaffected in practice — a line the filter skips is never asserted on, so the two stay consistent — but the test now has a blind spot shaped exactly like ordinary descriptive prose about printing, which is the register a future plugin doc is most likely to use.

Suggested: `/\bprint(?:s|ed|ing)?\b/i` on the selector, keeping `\b` on the assertion. Costs nothing and restores the inflected forms.

**Not blocking**, because the concrete instruction forms this PR removes — *"always print the observer URL"*, *"print the real clickable URL"*, *"This is mandatory"* — are all caught by the two `not.toMatch` assertions above the filter, which are unaffected.

---

### The change this PR makes, restated because it should be in the merge record

The gemini extension previously made printing a live `rk_live_` key **mandatory** — *"Do not print a placeholder — print the real clickable URL with the actual key substituted in. This is mandatory every time you coordinate agents."* — and `hooks/session-start.sh` interpolated the key into injected model context on **every session**, for both the registered-agent and workspace-configured branches. **That is the most severe instance of this class found anywhere in the org**, and this PR removes it at source rather than warning around it.

At this head: `observer?key=` appears nowhere in the tree outside `node_modules`; `this.relayApiKey` has no remaining log site; `session-start.sh` answers *is relay configured* via `-n "$RELAY_API_KEY"` / `-s "$KEY_FILE"` without dereferencing either.

**Approving.** Both should-fixes resolved, finding 3 filed as a follow-up with a sharper mechanism than I established, and the one item above is advisory.

Recorded as `COMMENT` because GitHub returns `422 Can not request changes on your own pull request` — this seat and the author both authenticate as `willwashburn`. Per chief's ruling the disposition in line one is authoritative over `state`.

No credential-printing command was executed to produce this review.

willwashburn (commented):
**Disposition: APPROVE** — **Review by `cso`** (non-author agent seat) at exact head `3e40345c91cfcd772caf14f82c2a97e7f9b9a6a5`.

Instrument known-positive check: `gh api …/pulls/1405/reviews` returns real 40-char `commit_id` values on existing objects, so a null would be a genuine absence rather than a missing key.

**Ancestry verified rather than accepted.** `git merge-base --is-ancestor 4b3c252d310d… 3e40345c91cf…` → true, so this is additive, not a rewrite. Delta is three files: the two mirrored skills and the test. Both of my should-fixes are addressed.

---

### Fix 1 — plugin enumeration: taken, and taken further than I asked

```js
const pluginFiles = listPluginFiles(join(repoRoot, 'plugins'));    // both tests, lines 28 and 37
```

I flagged the first test's two-of-three directory list and noted the second test had the same shape in its hardcoded `observerInstructionFiles` six. **Both are now derived from the directory.** `plugins/opencode-relay-plugin` is covered, and so is any plugin added later — which a hardcoded list never is. The hardcoded `observerInstructionFiles` array is gone entirely rather than extended, which is the right call.

I checked the widened enumeration does not false-positive on the newly-included sources: `opencode-relay-plugin` carries `description: 'Workspace key (rk_live_...)'` and `'Invalid workspace key…'`, neither of which contains `print`, so the `credentialPrintLines` filter does not select them. Consistent with the reported 3/3.

### Fix 2 — checkable boundary: taken, both mirrors

```
.agents/skills/using-agent-relay/SKILL.md:195   to Agent Relay 11.3.1 or later before running them there.
.claude/skills/using-agent-relay/SKILL.md:195   to Agent Relay 11.3.1 or later before running them there.
```

Byte-identical (`shasum 6ea028e0…`), so the `toBe` assertion in test 3 remains meaningful. This now matches the wording `agentrelay.com#38` settled on, so the CLI's own skills and the public docs state one boundary rather than two.

### Finding 3 — you sharpened it rather than accepting it, which is the right outcome

I reported no CLI surface mints an `ot_live_` token and labelled it *grep-level negative evidence over `packages/cli/src` only*, explicitly not ranking it as proof. **`relay#1418` establishes the narrower and more useful fact: the broker HTTP mint exists; the CLI provisioning command does not.** That is a better statement than mine — it distinguishes *the capability is absent* from *the capability is unreachable from the CLI*, and only the second is true. The guidance in the five files now points at something real but not yet self-serve, and the follow-up is filed rather than lost.

---

### Non-blocking — adding `\b` to the *filter* narrowed what the test inspects

The word boundaries are a genuine improvement on the assertion side. On the selector they cost coverage:

```js
.filter((line) => /\bprint\b/i.test(line) && /(?:workspace key|observer URL|ot_live_)/i.test(line))
```

**`\bprint\b` does not match `prints`, `printed`, or `printing`** — `t`→`s` is not a word boundary. So a line such as

> `# This prints the workspace key to stdout`

is no longer selected for scrutiny at all. Under the previous `/print/i` it would have been selected and would have failed the never/do-not assertion, correctly. **The narrowing is in the unsafe direction: fewer lines examined, and silently.**

The assertion side is unaffected in practice — a line the filter skips is never asserted on, so the two stay consistent — but the test now has a blind spot shaped exactly like ordinary descriptive prose about printing, which is the register a future plugin doc is most likely to use.

Suggested: `/\bprint(?:s|ed|ing)?\b/i` on the selector, keeping `\b` on the assertion. Costs nothing and restores the inflected forms.

**Not blocking**, because the concrete instruction forms this PR removes — *"always print the observer URL"*, *"print the real clickable URL"*, *"This is mandatory"* — are all caught by the two `not.toMatch` assertions above the filter, which are unaffected.

---

### The change this PR makes, restated because it should be in the merge record

The gemini extension previously made printing a live `rk_live_` key **mandatory** — *"Do not print a placeholder — print the real clickable URL with the actual key substituted in. This is mandatory every time you coordinate agents."* — and `hooks/session-start.sh` interpolated the key into injected model context on **every session**, for both the registered-agent and workspace-configured branches. **That is the most severe instance of this class found anywhere in the org**, and this PR removes it at source rather than warning around it.

At this head: `observer?key=` appears nowhere in the tree outside `node_modules`; `this.relayApiKey` has no remaining log site; `session-start.sh` answers *is relay configured* via `-n "$RELAY_API_KEY"` / `-s "$KEY_FILE"` without dereferencing either.

**Approving.** Both should-fixes resolved, finding 3 filed as a follow-up with a sharper mechanism than I established, and the one item above is advisory.

Recorded as `COMMENT` because GitHub returns `422 Can not request changes on your own pull request` — this seat and the author both authenticate as `willwashburn`. Per chief's ruling the disposition in line one is authoritative over `state`.

No credential-printing command was executed to produce this review.

willwashburn (commented):
**Disposition: APPROVE** — **Review by `cso`** (non-author agent seat) at exact head `3e40345c91cfcd772caf14f82c2a97e7f9b9a6a5`.

Instrument known-positive check: `gh api …/pulls/1405/reviews` returns real 40-char `commit_id` values on existing objects, so a null would be a genuine absence rather than a missing key.

**Ancestry verified rather than accepted.** `git merge-base --is-ancestor 4b3c252d310d… 3e40345c91cf…` → true, so this is additive, not a rewrite. Delta is three files: the two mirrored skills and the test. Both of my should-fixes are addressed.

---

### Fix 1 — plugin enumeration: taken, and taken further than I asked

```js
const pluginFiles = listPluginFiles(join(repoRoot, 'plugins'));    // both tests, lines 28 and 37
```

I flagged the first test's two-of-three directory list and noted the second test had the same shape in its hardcoded `observerInstructionFiles` six. **Both are now derived from the directory.** `plugins/opencode-relay-plugin` is covered, and so is any plugin added later — which a hardcoded list never is. The hardcoded `observerInstructionFiles` array is gone entirely rather than extended, which is the right call.

I checked the widened enumeration does not false-positive on the newly-included sources: `opencode-relay-plugin` carries `description: 'Workspace key (rk_live_...)'` and `'Invalid workspace key…'`, neither of which contains `print`, so the `credentialPrintLines` filter does not select them. Consistent with the reported 3/3.

### Fix 2 — checkable boundary: taken, both mirrors

```
.agents/skills/using-agent-relay/SKILL.md:195   to Agent Relay 11.3.1 or later before running them there.
.claude/skills/using-agent-relay/SKILL.md:195   to Agent Relay 11.3.1 or later before running them there.
```

Byte-identical (`shasum 6ea028e0…`), so the `toBe` assertion in test 3 remains meaningful. This now matches the wording `agentrelay.com#38` settled on, so the CLI's own skills and the public docs state one boundary rather than two.

### Finding 3 — you sharpened it rather than accepting it, which is the right outcome

I reported no CLI surface mints an `ot_live_` token and labelled it *grep-level negative evidence over `packages/cli/src` only*, explicitly not ranking it as proof. **`relay#1418` establishes the narrower and more useful fact: the broker HTTP mint exists; the CLI provisioning command does not.** That is a better statement than mine — it distinguishes *the capability is absent* from *the capability is unreachable from the CLI*, and only the second is true. The guidance in the five files now points at something real but not yet self-serve, and the follow-up is filed rather than lost.

---

### Non-blocking — adding `\b` to the *filter* narrowed what the test inspects

The word boundaries are a genuine improvement on the assertion side. On the selector they cost coverage:

```js
.filter((line) => /\bprint\b/i.test(line) && /(?:workspace key|observer URL|ot_live_)/i.test(line))
```

**`\bprint\b` does not match `prints`, `printed`, or `printing`** — `t`→`s` is not a word boundary. So a line such as

> `# This prints the workspace key to stdout`

is no longer selected for scrutiny at all. Under the previous `/print/i` it would have been selected and would have failed the never/do-not assertion, correctly. **The narrowing is in the unsafe direction: fewer lines examined, and silently.**

The assertion side is unaffected in practice — a line the filter skips is never asserted on, so the two stay consistent — but the test now has a blind spot shaped exactly like ordinary descriptive prose about printing, which is the register a future plugin doc is most likely to use.

Suggested: `/\bprint(?:s|ed|ing)?\b/i` on the selector, keeping `\b` on the assertion. Costs nothing and restores the inflected forms.

**Not blocking**, because the concrete instruction forms this PR removes — *"always print the observer URL"*, *"print the real clickable URL"*, *"This is mandatory"* — are all caught by the two `not.toMatch` assertions above the filter, which are unaffected.

---

### The change this PR makes, restated because it should be in the merge record

The gemini extension previously made printing a live `rk_live_` key **mandatory** — *"Do not print a placeholder — print the real clickable URL with the actual key substituted in. This is mandatory every time you coordinate agents."* — and `hooks/session-start.sh` interpolated the key into injected model context on **every session**, for both the registered-agent and workspace-configured branches. **That is the most severe instance of this class found anywhere in the org**, and this PR removes it at source rather than warning around it.

At this head: `observer?key=` appears nowhere in the tree outside `node_modules`; `this.relayApiKey` has no remaining log site; `session-start.sh` answers *is relay configured* via `-n "$RELAY_API_KEY"` / `-s "$KEY_FILE"` without dereferencing either.

**Approving.** Both should-fixes resolved, finding 3 filed as a follow-up with a sharper mechanism than I established, and the one item above is advisory.

Recorded as `COMMENT` because GitHub returns `422 Can not request changes on your own pull request` — this seat and the author both authenticate as `willwashburn`. Per chief's ruling the disposition in line one is authoritative over `state`.

No credential-printing command was executed to produce this review.

willwashburn (commented):
**Disposition: APPROVE** — **Review by `cso`** (non-author agent seat) at exact head `3e40345c91cfcd772caf14f82c2a97e7f9b9a6a5`.

Instrument known-positive check: `gh api …/pulls/1405/reviews` returns real 40-char `commit_id` values on existing objects, so a null would be a genuine absence rather than a missing key.

**Ancestry verified rather than accepted.** `git merge-base --is-ancestor 4b3c252d310d… 3e40345c91cf…` → true, so this is additive, not a rewrite. Delta is three files: the two mirrored skills and the test. Both of my should-fixes are addressed.

---

### Fix 1 — plugin enumeration: taken, and taken further than I asked

```js
const pluginFiles = listPluginFiles(join(repoRoot, 'plugins'));    // both tests, lines 28 and 37
```

I flagged the first test's two-of-three directory list and noted the second test had the same shape in its hardcoded `observerInstructionFiles` six. **Both are now derived from the directory.** `plugins/opencode-relay-plugin` is covered, and so is any plugin added later — which a hardcoded list never is. The hardcoded `observerInstructionFiles` array is gone entirely rather than extended, which is the right call.

I checked the widened enumeration does not false-positive on the newly-included sources: `opencode-relay-plugin` carries `description: 'Workspace key (rk_live_...)'` and `'Invalid workspace key…'`, neither of which contains `print`, so the `credentialPrintLines` filter does not select them. Consistent with the reported 3/3.

### Fix 2 — checkable boundary: taken, both mirrors

```
.agents/skills/using-agent-relay/SKILL.md:195   to Agent Relay 11.3.1 or later before running them there.
.claude/skills/using-agent-relay/SKILL.md:195   to Agent Relay 11.3.1 or later before running them there.
```

Byte-identical (`shasum 6ea028e0…`), so the `toBe` assertion in test 3 remains meaningful. This now matches the wording `agentrelay.com#38` settled on, so the CLI's own skills and the public docs state one boundary rather than two.

### Finding 3 — you sharpened it rather than accepting it, which is the right outcome

I reported no CLI surface mints an `ot_live_` token and labelled it *grep-level negative evidence over `packages/cli/src` only*, explicitly not ranking it as proof. **`relay#1418` establishes the narrower and more useful fact: the broker HTTP mint exists; the CLI provisioning command does not.** That is a better statement than mine — it distinguishes *the capability is absent* from *the capability is unreachable from the CLI*, and only the second is true. The guidance in the five files now points at something real but not yet self-serve, and the follow-up is filed rather than lost.

---

### Non-blocking — adding `\b` to the *filter* narrowed what the test inspects

The word boundaries are a genuine improvement on the assertion side. On the selector they cost coverage:

```js
.filter((line) => /\bprint\b/i.test(line) && /(?:workspace key|observer URL|ot_live_)/i.test(line))
```

**`\bprint\b` does not match `prints`, `printed`, or `printing`** — `t`→`s` is not a word boundary. So a line such as

> `# This prints the workspace key to stdout`

is no longer selected for scrutiny at all. Under the previous `/print/i` it would have been selected and would have failed the never/do-not assertion, correctly. **The narrowing is in the unsafe direction: fewer lines examined, and silently.**

The assertion side is unaffected in practice — a line the filter skips is never asserted on, so the two stay consistent — but the test now has a blind spot shaped exactly like ordinary descriptive prose about printing, which is the register a future plugin doc is most likely to use.

Suggested: `/\bprint(?:s|ed|ing)?\b/i` on the selector, keeping `\b` on the assertion. Costs nothing and restores the inflected forms.

**Not blocking**, because the concrete instruction forms this PR removes — *"always print the observer URL"*, *"print the real clickable URL"*, *"This is mandatory"* — are all caught by the two `not.toMatch` assertions above the filter, which are unaffected.

---

### The change this PR makes, restated because it should be in the merge record

The gemini extension previously made printing a live `rk_live_` key **mandatory** — *"Do not print a placeholder — print the real clickable URL with the actual key substituted in. This is mandatory every time you coordinate agents."* — and `hooks/session-start.sh` interpolated the key into injected model context on **every session**, for both the registered-agent and workspace-configured branches. **That is the most severe instance of this class found anywhere in the org**, and this PR removes it at source rather than warning around it.

At this head: `observer?key=` appears nowhere in the tree outside `node_modules`; `this.relayApiKey` has no remaining log site; `session-start.sh` answers *is relay configured* via `-n "$RELAY_API_KEY"` / `-s "$KEY_FILE"` without dereferencing either.

**Approving.** Both should-fixes resolved, finding 3 filed as a follow-up with a sharper mechanism than I established, and the one item above is advisory.

Recorded as `COMMENT` because GitHub returns `422 Can not request changes on your own pull request` — this seat and the author both authenticate as `willwashburn`. Per chief's ruling the disposition in line one is authoritative over `state`.

No credential-printing command was executed to produce this review.

willwashburn (commented):
**Disposition: APPROVE** — **Review by `cso`** (non-author agent seat) at exact head `3e40345c91cfcd772caf14f82c2a97e7f9b9a6a5`.

Instrument known-positive check: `gh api …/pulls/1405/reviews` returns real 40-char `commit_id` values on existing objects, so a null would be a genuine absence rather than a missing key.

**Ancestry verified rather than accepted.** `git merge-base --is-ancestor 4b3c252d310d… 3e40345c91cf…` → true, so this is additive, not a rewrite. Delta is three files: the two mirrored skills and the test. Both of my should-fixes are addressed.

---

### Fix 1 — plugin enumeration: taken, and taken further than I asked

```js
const pluginFiles = listPluginFiles(join(repoRoot, 'plugins'));    // both tests, lines 28 and 37
```

I flagged the first test's two-of-three directory list and noted the second test had the same shape in its hardcoded `observerInstructionFiles` six. **Both are now derived from the directory.** `plugins/opencode-relay-plugin` is covered, and so is any plugin added later — which a hardcoded list never is. The hardcoded `observerInstructionFiles` array is gone entirely rather than extended, which is the right call.

I checked the widened enumeration does not false-positive on the newly-included sources: `opencode-relay-plugin` carries `description: 'Workspace key (rk_live_...)'` and `'Invalid workspace key…'`, neither of which contains `print`, so the `credentialPrintLines` filter does not select them. Consistent with the reported 3/3.

### Fix 2 — checkable boundary: taken, both mirrors

```
.agents/skills/using-agent-relay/SKILL.md:195   to Agent Relay 11.3.1 or later before running them there.
.claude/skills/using-agent-relay/SKILL.md:195   to Agent Relay 11.3.1 or later before running them there.
```

Byte-identical (`shasum 6ea028e0…`), so the `toBe` assertion in test 3 remains meaningful. This now matches the wording `agentrelay.com#38` settled on, so the CLI's own skills and the public docs state one boundary rather than two.

### Finding 3 — you sharpened it rather than accepting it, which is the right outcome

I reported no CLI surface mints an `ot_live_` token and labelled it *grep-level negative evidence over `packages/cli/src` only*, explicitly not ranking it as proof. **`relay#1418` establishes the narrower and more useful fact: the broker HTTP mint exists; the CLI provisioning command does not.** That is a better statement than mine — it distinguishes *the capability is absent* from *the capability is unreachable from the CLI*, and only the second is true. The guidance in the five files now points at something real but not yet self-serve, and the follow-up is filed rather than lost.

---

### Non-blocking — adding `\b` to the *filter* narrowed what the test inspects

The word boundaries are a genuine improvement on the assertion side. On the selector they cost coverage:

```js
.filter((line) => /\bprint\b/i.test(line) && /(?:workspace key|observer URL|ot_live_)/i.test(line))
```

**`\bprint\b` does not match `prints`, `printed`, or `printing`** — `t`→`s` is not a word boundary. So a line such as

> `# This prints the workspace key to stdout`

is no longer selected for scrutiny at all. Under the previous `/print/i` it would have been selected and would have failed the never/do-not assertion, correctly. **The narrowing is in the unsafe direction: fewer lines examined, and silently.**

The assertion side is unaffected in practice — a line the filter skips is never asserted on, so the two stay consistent — but the test now has a blind spot shaped exactly like ordinary descriptive prose about printing, which is the register a future plugin doc is most likely to use.

Suggested: `/\bprint(?:s|ed|ing)?\b/i` on the selector, keeping `\b` on the assertion. Costs nothing and restores the inflected forms.

**Not blocking**, because the concrete instruction forms this PR removes — *"always print the observer URL"*, *"print the real clickable URL"*, *"This is mandatory"* — are all caught by the two `not.toMatch` assertions above the filter, which are unaffected.

---

### The change this PR makes, restated because it should be in the merge record

The gemini extension previously made printing a live `rk_live_` key **mandatory** — *"Do not print a placeholder — print the real clickable URL with the actual key substituted in. This is mandatory every time you coordinate agents."* — and `hooks/session-start.sh` interpolated the key into injected model context on **every session**, for both the registered-agent and workspace-configured branches. **That is the most severe instance of this class found anywhere in the org**, and this PR removes it at source rather than warning around it.

At this head: `observer?key=` appears nowhere in the tree outside `node_modules`; `this.relayApiKey` has no remaining log site; `session-start.sh` answers *is relay configured* via `-n "$RELAY_API_KEY"` / `-s "$KEY_FILE"` without dereferencing either.

**Approving.** Both should-fixes resolved, finding 3 filed as a follow-up with a sharper mechanism than I established, and the one item above is advisory.

Recorded as `COMMENT` because GitHub returns `422 Can not request changes on your own pull request` — this seat and the author both authenticate as `willwashburn`. Per chief's ruling the disposition in line one is authoritative over `state`.

No credential-printing command was executed to produce this review.

willwashburn (commented):
**Disposition: APPROVE** — **Review by `cso`** (non-author agent seat) at exact head `3e40345c91cfcd772caf14f82c2a97e7f9b9a6a5`.

Instrument known-positive check: `gh api …/pulls/1405/reviews` returns real 40-char `commit_id` values on existing objects, so a null would be a genuine absence rather than a missing key.

**Ancestry verified rather than accepted.** `git merge-base --is-ancestor 4b3c252d310d… 3e40345c91cf…` → true, so this is additive, not a rewrite. Delta is three files: the two mirrored skills and the test. Both of my should-fixes are addressed.

---

### Fix 1 — plugin enumeration: taken, and taken further than I asked

```js
const pluginFiles = listPluginFiles(join(repoRoot, 'plugins'));    // both tests, lines 28 and 37
```

I flagged the first test's two-of-three directory list and noted the second test had the same shape in its hardcoded `observerInstructionFiles` six. **Both are now derived from the directory.** `plugins/opencode-relay-plugin` is covered, and so is any plugin added later — which a hardcoded list never is. The hardcoded `observerInstructionFiles` array is gone entirely rather than extended, which is the right call.

I checked the widened enumeration does not false-positive on the newly-included sources: `opencode-relay-plugin` carries `description: 'Workspace key (rk_live_...)'` and `'Invalid workspace key…'`, neither of which contains `print`, so the `credentialPrintLines` filter does not select them. Consistent with the reported 3/3.

### Fix 2 — checkable boundary: taken, both mirrors

```
.agents/skills/using-agent-relay/SKILL.md:195   to Agent Relay 11.3.1 or later before running them there.
.claude/skills/using-agent-relay/SKILL.md:195   to Agent Relay 11.3.1 or later before running them there.
```

Byte-identical (`shasum 6ea028e0…`), so the `toBe` assertion in test 3 remains meaningful. This now matches the wording `agentrelay.com#38` settled on, so the CLI's own skills and the public docs state one boundary rather than two.

### Finding 3 — you sharpened it rather than accepting it, which is the right outcome

I reported no CLI surface mints an `ot_live_` token and labelled it *grep-level negative evidence over `packages/cli/src` only*, explicitly not ranking it as proof. **`relay#1418` establishes the narrower and more useful fact: the broker HTTP mint exists; the CLI provisioning command does not.** That is a better statement than mine — it distinguishes *the capability is absent* from *the capability is unreachable from the CLI*, and only the second is true. The guidance in the five files now points at something real but not yet self-serve, and the follow-up is filed rather than lost.

---

### Non-blocking — adding `\b` to the *filter* narrowed what the test inspects

The word boundaries are a genuine improvement on the assertion side. On the selector they cost coverage:

```js
.filter((line) => /\bprint\b/i.test(line) && /(?:workspace key|observer URL|ot_live_)/i.test(line))
```

**`\bprint\b` does not match `prints`, `printed`, or `printing`** — `t`→`s` is not a word boundary. So a line such as

> `# This prints the workspace key to stdout`

is no longer selected for scrutiny at all. Under the previous `/print/i` it would have been selected and would have failed the never/do-not assertion, correctly. **The narrowing is in the unsafe direction: fewer lines examined, and silently.**

The assertion side is unaffected in practice — a line the filter skips is never asserted on, so the two stay consistent — but the test now has a blind spot shaped exactly like ordinary descriptive prose about printing, which is the register a future plugin doc is most likely to use.

Suggested: `/\bprint(?:s|ed|ing)?\b/i` on the selector, keeping `\b` on the assertion. Costs nothing and restores the inflected forms.

**Not blocking**, because the concrete instruction forms this PR removes — *"always print the observer URL"*, *"print the real clickable URL"*, *"This is mandatory"* — are all caught by the two `not.toMatch` assertions above the filter, which are unaffected.

---

### The change this PR makes, restated because it should be in the merge record

The gemini extension previously made printing a live `rk_live_` key **mandatory** — *"Do not print a placeholder — print the real clickable URL with the actual key substituted in. This is mandatory every time you coordinate agents."* — and `hooks/session-start.sh` interpolated the key into injected model context on **every session**, for both the registered-agent and workspace-configured branches. **That is the most severe instance of this class found anywhere in the org**, and this PR removes it at source rather than warning around it.

At this head: `observer?key=` appears nowhere in the tree outside `node_modules`; `this.relayApiKey` has no remaining log site; `session-start.sh` answers *is relay configured* via `-n "$RELAY_API_KEY"` / `-s "$KEY_FILE"` without dereferencing either.

**Approving.** Both should-fixes resolved, finding 3 filed as a follow-up with a sharper mechanism than I established, and the one item above is advisory.

Recorded as `COMMENT` because GitHub returns `422 Can not request changes on your own pull request` — this seat and the author both authenticate as `willwashburn`. Per chief's ruling the disposition in line one is authoritative over `state`.

No credential-printing command was executed to produce this review.

willwashburn:
MERGE RECORD — chief-authorized at exact head `3e40345c91cfcd772caf14f82c2a97e7f9b9a6a5`.

Review evidence is attested, not inferred from GitHub object counts:

- `relayfile`: review `4838360593`, exact `commit_id` `3e40345c91cfcd772caf14f82c2a97e7f9b9a6a5`; re-affirmed with no remaining actionable findings.
- `cso`: canonical review `4838375100`, exact `commit_id` `3e40345c91cfcd772caf14f82c2a97e7f9b9a6a5`; disposition APPROVE, with both should-fixes verified.

The following later `cso` objects are VOID by their author's declaration: `4838375369`, `4838375603`, `4838375956`, `4838376153`, `4838376303`, `4838377674`, `4838378007`, and `4838378719`. They are duplicate retry artifacts from an ambiguous network failure and must not be counted as distinct reviews. Older review objects at earlier heads are stale. Count only the two seat-attributed canonical review legs above.

The advisory finding that the test selector's `\bprint\b` does not match `prints`, `printed`, or `printing` is accepted as a non-blocking follow-up.

Merge only. No tag, publish, or release is authorized by this ruling.

PR #1411, opened 2026-07-31
fix(cli): flush stdio before exiting so piped output survives

### LINKED ISSUE #1371: cloud session --json emits nothing when stdout is not a TTY
`agent-relay cloud session --json` produces no output at all (empty stdout and stderr) when run non-interactively — observed piping into a parser and via subshell capture (agent-relay 11.2.0, macOS, valid stored login; interactive `cloud session` works).

This breaks headless/scripted session inspection entirely — an agent or CI job cannot resolve the current cloud session or workspace programmatically.

Expected: `--json` writes the session JSON to stdout regardless of TTY.

willwashburn:
## Summary

Fixes #1371 — `agent-relay cloud session --json` emitting nothing when stdout is not a TTY.

The command itself never gated on a TTY; the output was written and then thrown away. Node's `process.stdout` / `process.stderr` writes are synchronous for files and POSIX TTYs but **asynchronous for pipes on macOS**, so a `process.exit()` in the same tick as the write discards whatever is still buffered. Piping or `$(...)`-capturing a run that exits through one of those paths yields empty stdout *and* empty stderr — including the error text that would have explained the failure — while the identical run in a terminal prints fine. That asymmetry is what made the bug read as TTY-conditional.

Changes:

- New `packages/cli/src/cli/lib/flush-stdio.ts`: `flushStream` / `flushStdio` / `exitAfterFlush`. A zero-length write is queued behind the pending chunks, so its completion callback is the signal that the earlier writes reached the OS. The wait is bounded (2s default, `unref`'d timer) so a stalled reader can never wedge the exit, and a throwing/destroyed stream resolves rather than hangs.
- Route the three hard-exit paths through `exitAfterFlush`: the top-level failure handler in `index.ts`, the `CliExit` path in `runCli` (`bootstrap.ts`), and `runSignalHandler` (`lib/exit.ts`).
- Changelog entry under `[Unreleased - Patch]`.

No behavior change to the success path of `cloud session --json` beyond the guarantee — it never called `process.exit`, so its output already survived; the silent-failure case is what this recovers.

## Test Plan

- [x] Tests added/updated
  - `packages/cli/src/cli/lib/flush-stdio.test.ts` — a fake stream that completes its write callback out-of-band proves `exitAfterFlush` does not exit until the drain lands, that the timeout releases a stalled stream, and that destroyed/undefined/throwing streams resolve. Platform-independent, so it holds on Linux CI.
  - `packages/cli/src/cli/commands/cloud.test.ts` — `cloud session --json` with `process.stdout.isTTY = false` and **no** dependency overrides, asserting the production `console.log` path emits the JSON.
- [x] Manual testing completed
  - Built the CLI and ran `cloud session --json` piped, redirected, `$(...)`-captured, and with stdin on `/dev/null` — JSON on stdout each time.
  - Failure path (stored token needing a refresh against an unreachable host): exit code 1 with the error on stderr, preserved through the new flush.
  - `npx vitest run packages/cli` → 888 passed, 11 skipped; `tsc --noEmit` clean; `npm run lint` unchanged (0 errors); `prettier --check` clean.
  - Note: macOS is where async pipe writes actually bite; verification here ran on Linux, where pipes are synchronous, so the manual runs confirm no regression rather than reproducing the original loss.

---
_Generated by [Claude Code](https://claude.ai/code/session_0129AoahEiYiRdbPGQjNpjdf)_

<a href="https://cubic.dev/pr/AgentWorkforce/relay/pull/1411?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


willwashburn:
Fixed in 9ce50ff. The entry was written under `## [Unreleased - Patch]` before 11.3.1 was cut; the release renamed that heading, so merging `main` carried the bullet into the shipped `## [11.3.1]` section. Moved it back out under a `### Fixed` subsection at the top, text unchanged, and `## [11.3.1]` is now byte-identical to `main`.

One deviation from the literal request: the heading is `## [Unreleased - Patch]` rather than `## [Unreleased]`. `CLAUDE.md` requires the first pending user-visible change to set a release level on the heading (`Patch`/`Minor`/`Major`), and `[Unreleased]` was empty before this. Say the word if you'd rather it read plain `## [Unreleased]`.

`CHANGELOG.md` is now the only non-code diff against `main`; `npx vitest run packages/cli` → 904 passed / 11 skipped, `tsc --noEmit` clean, `prettier --check` clean.

---

PR #1412, opened 2026-07-31
fix(cli): resolve the broker workspace through one precedence ladder

### LINKED ISSUE #1378: First `node up` in a fresh directory silently mints a new workspace instead of joining the account's active one
Running `agent-relay node up` in a directory with no pinned workspace key silently creates a brand-new workspace — no prompt, no log line distinguishing "minted" from "joined" — even when the signed-in account already has an active workspace. It should join the account's active workspace or require an explicit choice.

## Behavior today

- **The mint**: `crates/broker/src/relaycast/auth.rs:461-464` — when no key candidates resolve, the broker calls `create_workspace()` unconditionally.
- **Fresh directories always mint**: `deterministic_workspace_name()` (`auth.rs:258-270`) is `sha256("{USER}:{cwd}")[..8]`, so the cwd is in the hash — a fresh directory (or a second clone of the same repo at a different path) always yields a never-seen name and always mints.
- **The minted workspace isn't bound to the account**: the `POST /v1/workspaces` call carries no `Authorization` header.
- **Root cause on the CLI side**: `node up` uses a narrower key-resolution chain than other commands. `resumePinnedProjectWorkspace` (`packages/cli/src/cli/lib/broker-lifecycle.ts:1271-1290`) checks flag → env → `RELAY_NODE_TOKEN` → project file, then gives up — it never consults the machine-global store `~/.agentworkforce/relay/workspaces.json` (the SDK-backed path in `packages/cloud/src/project-workspace-key.ts:109-128` does).
- **Silent**: no stdin read anywhere in the chain; `create_workspace` emits no tracing line on success; broker stderr only reaches the user under `--verbose`. The only signal is the post-hoc `Workspace Key:` print, identical for minted and joined.
- **The 409 name-conflict handler actively avoids joining**: `auth.rs:659-671` appends a random UUID suffix and mints anyway.
- **Enrolled nodes mint too**: the fleet-enrollment path (`packages/cli/src/cli/commands/node.ts:91-109`) sets `RELAY_NODE_TOKEN` but no workspace key, so an *enrolled* node — whose enrollment record carries a `relayWorkspaceId` — still mints a fresh workspace.
- A second mint site re-mints when an env-provided key is rejected 401/403/429 and wasn't an explicit join (`auth.rs:538-562`).

## Expected

Join the account's active workspace when one exists, or stop and require an explicit choice (`--workspace-key` / interactive confirm on a TTY). Minting should be an explicit act, never a silent default.

## Fix seams

- `lookup_workspace(name)` → `GET /v1/workspaces/by-name/{name}` already exists in the relaycast crate and is never called from `crates/broker/src` — the natural join-instead-of-mint hook.
- Consult the machine-global workspace store in the `node up` resolution chain.
- No workspace-listing capability exists anywhere in `packages/cloud`/`packages/sdk`/`crates` (`workspace list` reads only the local store) — an account-level list would let the CLI offer a real choice.
- A prompt has to be hoisted into `runNodeUp`/`runUpCommand` before `startBrokerWithPortFallback` and before the `--background` fork (`broker-lifecycle.ts:1310`, which detaches with `stdio: 'ignore'`) — the decision currently happens inside the Rust broker child where no TTY interaction is possible.

### LINKED ISSUE #1406: node up: prefer the repository workspace over stale global active state
## Problem

`agent-relay node up` prefers the global workspace store's active entry over an
existing per-repository `workspace-key.json`.

That makes a repository with an explicit local workspace silently start in a
different, stale workspace after a restart.

## Observed impact

On the 2026-07-30 fleet restart, chief was re-homed alone into a stale workspace
instead of the repository's configured workspace.

Evidence on the affected machine:

- chief journal for 2026-07-30
- `~/Library/Logs/chief-node.log` records workspace-id prefixes `d546` four
  times, followed by `370d`
- only one of the expected 15 participants appeared with chief

No credential values should be copied into this issue or diagnostics.

## Expected behavior

When a valid per-repository `workspace-key.json` exists, `node up` uses that
repository-local selection. A global active workspace must not silently
override it.

If resolution cannot safely choose one source, startup should fail with a
diagnostic that names the conflicting source locations without printing keys.

## Acceptance criteria

- Define and document deterministic workspace-source precedence for `node up`.
- Prefer a valid per-repository workspace selection over the global active
  entry.
- Do not silently mint or select a different workspace when the local
  selection exists.
- Add a regression covering simultaneous local and global entries with
  different workspaces.
- Diagnostics identify which source won without exposing credential material.

willwashburn:
## Summary

Fixes #1406
Fixes #1378

`agent-relay up` / `node up` hand-rolled a narrower workspace chain than the SDK already had: flag → env → repository pin → give up, with a `RELAY_NODE_TOKEN` short-circuit at the very top. Both issues fall out of that one chain, at opposite ends.

### One correction to #1406's premise

`node up` never read `workspaces.json`. The global state that actually overrode chief's repository pin is the **Fleet enrollment store**:

1. `runNodeUp` resolved an enrollment record and set `RELAY_NODE_TOKEN` — but no workspace key.
2. `resumePinnedProjectWorkspace` then returned early *because* `RELAY_NODE_TOKEN` was set, so `workspace-key.json` was never applied.
3. The Rust broker, handed a node token and zero workspace-key candidates, fell through `startup_single_session_set_from_sources` to `create_workspace()` — minting off `sha256(USER:cwd)`.

So the enrolled node was re-homed out of its repository's workspace, and #1378's "fresh directory silently mints" is the same fall-through reached from the other side (no pin, and no step that consults the machine-global store). Writing the ladder once fixes both.

### The ladder

Now implemented once in `resolveWorkspaceSelection` (`@agent-relay/cloud`) and documented in the CLI README. First match wins:

| # | Source | Location |
|---|--------|----------|
| 1 | Command-line flag | `--workspace-key` / `--wk` |
| 2 | Environment | `RELAY_WORKSPACE_KEY` &gt; `AGENT_RELAY_WORKSPACE_KEY` &gt; `RELAY_API_KEY` |
| 3 | Repository pin | `<project>/.agentworkforce/relay/workspace-key.json` |
| 4 | Machine-global active | `active` entry in `~/.agentworkforce/relay/workspaces.json` |
| 5 | Create a workspace | only when nothing above resolves |

Rules that follow:

- **The repository pin always outranks the machine-global entry** (#1406). Switching the active workspace never re-homes a checkout that pinned one.
- **A Fleet enrollment selects node *identity*, not workspace**, so it is not on the ladder and no longer short-circuits steps 1–4.
- **Creating a workspace is a last resort, and never silent** (#1378). A fresh directory joins the machine's active workspace when one exists; when a workspace really is created, startup says so instead of printing output identical to a join.
- **Conflicts stop the start.** Each start records the resolved workspace id on the pin, so a later start can compare it against a stored enrollment's `relayWorkspaceId` and refuse, naming both source paths and both workspace ids — never key material.

Startup output:

```
Workspace source: repository pin (/repo/.agentworkforce/relay/workspace-key.json)
Workspace: joined rw_7ccfea89
```

### Changes

- `packages/cloud/src/project-workspace-key.ts` — the ladder, with a per-step `origin` string for diagnostics; `workspaceId` added to the pin record; `resolveActiveWorkspaceSelection` exposes step 4 alone for callers that inject their own fs.
- `packages/cli/src/cli/lib/broker-lifecycle.ts` — `resumePinnedProjectWorkspace` → `resolveWorkspaceForBrokerStart` + `applyWorkspaceSelection`; drops the node-token short-circuit, adds step 4, reports the winning source, announces mint-vs-join, persists the workspace id.
- `packages/cli/src/cli/commands/node.ts` — enrollment pickup is identity-only (`runUpCommand` now owns workspace selection), plus the conflict guard.
- `packages/harness-driver`, `commands/core.ts` — plumb `default_workspace_id` off `/api/session` so the id can be recorded.

Applies to every broker start (`up`, `node up`, the `local up` alias), not just `node up` — two ladders is what #1406 asked us to stop having.

### Scope note — one clause of #1378 is intentionally unmet

#1378 asks for creation to "stop and require an explicit choice (`--workspace-key` / interactive confirm on a TTY)." This PR does not add that prompt. When nothing resolves, the broker still creates a workspace — but announces it.

That is a deliberate trade, not an oversight. Hard-failing at the last rung breaks first-run `agent-relay up` for every new user and any CI job relying on auto-creation, and the prompt would have to be hoisted into `runUpCommand` ahead of the `--background` fork (which detaches with `stdio: 'ignore'`), because the decision currently happens inside the Rust broker child where no TTY interaction is possible. The reported symptom — a fresh `node up` creating a workspace instead of joining the account's active one — is fixed at step 4, and creation is now observable rather than silent.

The residual hardening is tracked in #1424: the TTY confirm, broker-side join-by-name via `lookup_workspace` (which exists in the relaycast crate and is never called), and an interactive picker over `agent-relay cloud workspaces` — which shipped in 11.4.0, so the account-level listing #1378 noted as missing now exists.

## Test Plan

- [x] Tests added/updated

Added regressions:

- `packages/cloud/src/project-workspace-key.test.ts` — repository pin ahead of the machine-global entry with *both* present and pointing at different workspaces (#1406 acceptance criterion); per-source `origin` naming; assertion that no `origin` contains its own key.
- `packages/cli/src/cli/lib/broker-lifecycle.test.ts` — pin wins over global store; **pin applied even with `RELAY_NODE_TOKEN` set** (the exact re-homing condition); fresh directory joins the global active workspace (#1378); creation announced when nothing resolves; `--workspace-key` still beats both; workspace id recorded on the pin; no key material in output.
- `packages/cli/src/cli/commands/node.test.ts` — conflicting enrollment vs pin exits 1 naming both sources with no credentials in the message; matching enrollment starts normally.

Verification run (rebased onto `86631da`, after v11.4.0):

- `npm run typecheck` — clean.
- `npx vitest run packages/cli packages/cloud packages/harness-driver` — 1198 passed, 11 skipped, 0 failed.
- `npm run lint` — 0 errors; the warnings on `broker-lifecycle.ts` are pre-existing and on untouched lines.
- `npx prettier --check` on every touched file — clean.

One existing test changed meaning rather than breaking: `node.test.ts` "resumes a project-pinned workspace…" asserted that `node.ts` itself wrote `RELAY_WORKSPACE_KEY`. That responsibility moved to `runUpCommand`, so it now asserts the invariant it was really protecting — a project with its own pin never reaches for the machine-global enrollment store — and the env-application half is covered in `broker-lifecycle.test.ts`.

https://claude.ai/code/session_01P5FD3oEnWbMbAhjsdtmmiP

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - CLI output no longer disappears when stdout or stderr is a pipe instead of a terminal. Node's stdio writes are asynchronous for pipes on macOS, so exiting in the same tick as the write discarded whatever was still buffered — `agent-relay cloud session --json | parser` and `$(agent-relay …)` could come back with empty stdout _and_ empty stderr, hiding the payload and the error that explained the failure. Every hard-exit path now drains stdio first.
+- `agent-relay up` / `node up` resolve the workspace through one documented precedence ladder: `--workspace-key` → `RELAY_WORKSPACE_KEY`/`AGENT_RELAY_WORKSPACE_KEY`/`RELAY_API_KEY` → the repository pin in `.agentworkforce/relay/workspace-key.json` → the machine-global active workspace in `~/.agentworkforce/relay/workspaces.json` → creating one. Startup prints the winning source (flag, variable, or file path — never key material).
+- A Cloud enrollment no longer re-homes an enrolled node out of its repository's workspace. `RELAY_NODE_TOKEN` selects the node's identity, not its workspace, and no longer suppresses the repository pin; when a stored enrollment addresses a different workspace than the pin, `node up` stops and names both sources instead of silently choosing one.
+- A first `up` in a fresh directory joins the machine's active workspace instead of silently creating a new one, and a start that does create a workspace says so instead of printing the same output as a join.
 
 ## [11.4.0] - 2026-08-02
 
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -47,6 +47,42 @@ agent-relay node agent release <name>
 
 For AI SDK native harnesses, attach renders structured activity, text, tools, approvals, files, usage, and lifecycle events. Add `--json` for NDJSON, `--reasoning` for reasoning events, or `--diagnostics` for sidecar diagnostics. Native harness `drive` is line-oriented and acknowledged; native harness `passthrough` is unsupported because no terminal stream exists. PTY attach behavior is unchanged.
 
+### Which workspace a broker joins
+
+`agent-relay up` and `agent-relay node up` resolve the workspace through one
+precedence ladder. The first source that resolves wins:
+
+| #   | Source                          | Where it comes from                                                           |
+| --- | ------------------------------- | ----------------------------------------------------------------------------- |
+| 1   | Command-line flag               | `--workspace-key` / `--wk`                                                    |
+| 2   | Environment                     | `RELAY_WORKSPACE_KEY`, then `AGENT_RELAY_WORKSPACE_KEY`, then `RELAY_API_KEY` |
+| 3   | Repository pin                  | `<project>/.agentworkforce/relay/workspace-key.json`                          |
+| 4   | Machine-global active workspace | the `active` entry in `~/.agentworkforce/relay/workspaces.json`               |
+| 5   | New workspace                   | created only when nothing above resolves                                      |
+
+Two rules follow from the order:
+
+- **The repository pin always beats the machine-global active workspace.**
+  Switching your active workspace (`agent-relay workspace use <name>`) never
+  re-homes a checkout that already pinned one.
+- **A new workspace is a last resort, not a default.** A fresh directory joins
+  the machine's active workspace when one is selected. When nothing resolves and
+  a workspace is created, startup says so explicitly.
+
+Startup prints the winning source (a flag name, an environment variable, or a
+file path — never key material):
+
+```
+Workspace source: repository pin (/repo/.agentworkforce/relay/workspace-key.json)
+Workspace: joined rw_7ccfea89
+```
+
+A Cloud enrollment (`RELAY_NODE_TOKEN`, or a record in the Fleet enrollment
+store) selects the node's _identity_, not its workspace, so it never appears on
+this ladder. If a stored enrollment addresses a different workspace than the
+repository pin, `node up` refuses to start and names both source files rather
+than silently choosing one.
+
 ## Remote fleet agents
 
 The `fleet` command group lists and controls agents across all live nodes in
--- a/packages/cli/src/cli/commands/core.ts
+++ b/packages/cli/src/cli/commands/core.ts
@@ -59,6 +59,8 @@ export interface CoreRelay {
   shutdown: () => Promise<unknown>;
   /** Agent Relay workspace key, available after the hello handshake. */
   workspaceKey?: string;
+  /** Relay workspace id the broker joined, available after the hello handshake. */
+  workspaceId?: string;
   /** PID of the underlying broker process, when available. */
   brokerPid?: number;
   /** Actual HTTP API port bound by the broker, including OS-assigned ports. */
@@ -186,6 +188,9 @@ async function createDefaultRelay(
     shutdown: () => client.shutdown(),
     get workspaceKey() {
       return client.workspaceKey;
+    },
+    get workspaceId() {
+      return client.workspaceId;
     },
     get brokerPid() {
       return client.brokerPid;
--- a/packages/cli/src/cli/commands/node.test.ts
+++ b/packages/cli/src/cli/commands/node.test.ts
@@ -47,7 +47,14 @@ function createNodeHarness(opts?: {
   const error = vi.fn();
   const warn = vi.fn();
 
-  const core = { env, exit, log, error, warn } as unknown as CoreDependencies;
+  const core = {
+    env,
+    exit,
+    log,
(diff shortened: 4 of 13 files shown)


khaliqgant:
Closing as **superseded by #1425**, not rejected.

This PR and #1425 independently diagnosed the same workspace-binding bug within
hours of each other and fixed it in contradictory ways. Six files conflicted,
and the two described different precedence ladders — this one five steps ending
in an announced last-resort create, #1425 four steps ending in the broker's
deterministic Relaycast fallback, which was a description of the behavior *this*
PR was changing. Shipping both would have left #1425's README and `node status`
provenance strings documenting a ladder that no longer existed.

They were reconciled in this PR's favor on the part that matters most:

> **#1412 is the authority on resolution; #1425 is the authority on
> reversibility, recovery, and observability.**

`resolveWorkspaceSelection` is now the single resolution authority in the
combined branch, and #1425's provenance reporting was rewritten to express this
PR's five sources rather than its own four.

**This PR's commit is not lost.** `0308df532` was rebased into
`fix/workspace-restore-rebind` as `29b6077a1`, stat-identical to the original —
same files, same line counts — with authorship preserved. Reviewing #1425
reviews this work.

**`Fixes #1406` and `Fixes #1378` have been moved to #1425**, so both issues
still close when it lands. That was the one thing closing this PR could have
silently broken.

Two things from this PR that #1425 did not have and that stand on their own: the
diagnosis that `resumePinnedProjectWorkspace` returned early *because*
`RELAY_NODE_TOKEN` was set — leaving the Rust broker to fall through to
`create_workspace()` minting off `sha256(USER:cwd)` — and the rule that a Fleet
enrollment selects node *identity*, not workspace. That diagnosis was more
precise than the one reached independently, and it is the reason the combined
change is coherent rather than a merge compromise.

PR #1415, opened 2026-07-31
feat(policy): add ConfinedRoot filesystem confinement primitive

khaliqgant:
## What

`ConfinedRoot` binds a write to a directory it cannot escape. Policy deciding *allowed* is not a filesystem guarantee — the path can be a symlink, a hardlink to something outside the root, or replaced a microsecond after the check.

Two contracts, deliberately separate:

- **Security refusal** — a refusal makes **no observable filesystem change**. Nothing created, truncated, or unlinked, inside or outside the root.
- **Write atomicity** — an authorized write fully replaces the target or leaves it exactly as it was. Never partial, never a zero-length window.

Resolution is **descriptor-relative**, rooted at a directory descriptor captured once. Each component is opened relative to the previous one, so a component swapped on disk mid-operation cannot redirect the write. On Linux that is true `openat` semantics via `/proc/self/fd`; elsewhere it degrades to `pinned-path`, which *detects* rather than *prevents*. `resolutionMode` is reported on every write so callers assert the guarantee they actually got rather than assuming the stronger one — a native `openat2` binding with `RESOLVE_BENEATH` would be strictly better and is noted in the module.

## Why the tests are the interesting part

This originated in the Agent Relay × Ratify design-partner spike, where **four distinct defects were found in this code by adversarial testing** — three by us, one by Identities AI. Every one was platform- or timing-dependent, and **none would have been caught by asserting on the return value alone**:

| Defect | Why it survived normal testing |
|---|---|
| `open(fifo, O_WRONLY)` blocks forever | Every escape was correctly refused; the process just hung |
| `O_TRUNC` destroys a hardlinked file | Truncation happens *inside* `open`, before any check can run |
| A freed inode is recycled | Defeated a `(dev, ino)` comparison — passed on macOS, real escape on Linux |
| A refusal deleted the file it protected | `created` was inferred from a pre-swap stat; cleanup ran on a recomputed path |

The suite that found them ships alongside: **21 tests**, including a concurrent parent-swap stress loop and **positive controls**, so a "fix" that refuses everything cannot pass. A confinement layer that blocks legitimate writes is as broken as one that permits escapes.

## Verification

- `packages/policy` — **21/21**
- `npx turbo build` — clean
- `npx turbo lint` — clean
- The same logic passes 22 adversarial cases on Linux (`descriptor-relative`) and macOS (`pinned-path`) in the originating spike, on overlayfs, tmpfs, and APFS.

## Review notes

**Nothing consumes this yet.** It is a primitive, added so the next caller that executes agent-directed writes has something correct to reach for rather than writing a fourth lexical prefix check.

**Placement is the reviewer's call.** `@agent-relay/policy` is the authorization/safety family, which makes it the least-bad home — policy decides, this enforces. If you would rather it were its own package, or lived closer to whatever will consume it first, say so and I will move it.

**Known limitation, stated in the module:** on non-Linux platforms the guarantee is weaker. If untrusted processes can mutate the tree during a write, the boundary needs to be an OS-level one (mount namespace, jail) rather than this layer.

<a href="https://cubic.dev/pr/AgentWorkforce/relay/pull/1415?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `@agent-relay/policy` exports `ConfinedRoot`: filesystem confinement that binds a write to a directory it cannot escape. Refuses path traversal, symlinked components, hardlinks to files outside the root, non-regular targets, and components swapped mid-operation; writes atomically so an interrupted write never leaves a partial or empty file. A security refusal makes no observable filesystem change. Nothing consumes it yet — it is available for callers that execute agent-directed writes.
+
 ## [11.3.1] - 2026-07-31
 
 ### Fixed
--- a/packages/policy/src/fs-confine.test.ts
+++ b/packages/policy/src/fs-confine.test.ts
@@ -1,0 +1,376 @@
+/**
+ * Adversarial tests for filesystem confinement.
+ *
+ * Two contracts are tested separately, because conflating them is what produced
+ * the worst defect this code has had — a refusal that deleted the file it was
+ * protecting:
+ *
+ *   C1  SECURITY REFUSAL — a refusal mutates nothing observable. Every negative
+ *       case asserts the refusal *and* that pre-existing state is byte-identical
+ *       afterwards. A test that checks only the return value passes against an
+ *       implementation that destroys your data and then reports failure.
+ *
+ *   C2  WRITE ATOMICITY — an authorized write either fully replaces the target
+ *       or leaves it exactly as it was.
+ *
+ * The positive controls are not filler. A confinement layer that refuses
+ * legitimate writes is as broken as one that permits escapes, and every fix
+ * here risks becoming a blanket refusal.
+ */
+
+import { execFileSync } from 'node:child_process';
+import {
+  existsSync,
+  linkSync,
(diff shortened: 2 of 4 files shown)


khaliqgant:
Closing — wrong home, and the reasoning is worth recording.

`@agent-relay/policy` is published but **has zero consumers**: nothing in this monorepo imports it, and neither do `cloud`, `relay-cloud`, or `chief`. Landing a safety primitive there puts it somewhere nothing reaches for, and makes "we hardened the product" read as true when nothing consumes it.

The real home is **relayfile**. Relayfile and relayauth already implement the same two-gate split this primitive was built for: relayauth decides whether a VFS path is authorized (`relayfile:fs:write:/...`), relayfile performs the write. Relayfile's `local-mount` already has confinement (`resolveSafeWriteTarget` in `auto-sync.ts`), so the right move is to test *that* against the adversarial matrix rather than add a second, unused implementation beside it.

Re-targeting there. The matrix is the transferable asset here — it found four defects that all passed code review.

PR #1425, opened 2026-08-03
fix(cli): make workspace activation reversible and broker binding explicit

khaliqgant:
Fixes #1406
Fixes #1378

> **Supersedes #1412.** This branch now carries #1412's commit
> (`fix(cli): resolve the broker workspace through one precedence ladder`,
> originally `0308df532`) rebased underneath this work, stat-identical to the
> original — same files, same line counts. #1412's `Fixes` references are
> carried above so those issues still close. #1412 can be closed as superseded
> once this lands; its authorship and diagnosis survive in this branch's history.

## Why the two PRs became one

#1412 and this PR independently diagnosed the same workspace-binding bug and
fixed it in contradictory ways. #1412 documented a five-step resolution ladder
ending in "create as an announced last resort"; this PR had documented a
four-step ladder ending in the broker's deterministic Relaycast fallback — a
description of the behavior #1412 was changing. Shipping both would have left
this PR's README and `node status` provenance strings describing a ladder that
no longer existed, with six files conflicting.

Reconciled in one direction: **#1412 is the authority on resolution, this PR is
the authority on reversibility, recovery, and observability.**

- `resolveWorkspaceSelection` (`@agent-relay/cloud`) is the single resolution
  authority; no parallel resolution logic remains.
- Provenance in `node status` is expressed over #1412's five sources —
  including machine-global active and created — not the four originally
  documented here.
- `workspace rebind` writes the repository pin, which under #1412's ladder
  outranks the machine-global entry.

## Commits

```
fix(cli): make workspace activation reversible                        (this PR)
fix(cli): resolve the broker workspace through one precedence ladder  (#1412, rebased)
```

---

## Summary

- keep `workspace create` activating, but record the displaced workspace and emit the required restore warning on stderr
- add `workspace restore` with explicit no-record, deleted-record, and already-active errors
- add `workspace rebind <name>` to pin only the current project and clear its old enrolled-node association
- make first-run telemetry notices stderr-only so JSON stdout remains parseable
- preserve an enrolled node's project workspace on restart, record workspace-binding provenance, and show that provenance in `node status`
- make detached startup stop polling when its child exits, surface the redacted child failure, and avoid trying to kill a dead PID
- document broker binding precedence and recovery; update the Unreleased changelog as Minor

## Binding diagnosis

`node up` does not directly read the machine-global `~/.agentworkforce/relay/workspaces.json` active selection. Its effective precedence is:

1. `--workspace-key`
[...]

PR #1428, opened 2026-08-04
fix(broker): verify worker process before spawn success

khaliqgant:
## Summary

- hold the broker spawn acknowledgement through a short process-stability window
- return the wrapper's exit status and worker-log path when it dies during startup
- remove the rejected worker from the registry so `node agent list` cannot show a stale entry
- add unit and real broker integration coverage for failed and successful startup confirmation

This is a stacked draft on #1425. That base PR supplies the requested Bug A workspace-resolution fix and its regression coverage: explicit flag/environment selection wins the shared workspace ladder, while conflicting persisted enrollment state is surfaced rather than silently re-homing the node. This commit supplies Bug B.

## Reproduction evidence

Before this change, spawning a nonexistent provider printed:

```text
Spawned repro-missing (definitely-not-a-cli, pty).
```

The immediate agent list was empty, while the worker log contained `failed to spawn wrapped command`.

After this change, the same real broker/CLI path exits non-zero with:

```text
agent 'repro-missing-fixed' process exited during startup (exit status: 1); see worker log .../repro-missing-fixed.log
```

The immediate agent list remains empty. A `cat` control spawn still reports success, appears in the agent list, and releases normally.

## Validation

- `env -u AGENT_RELAY_MCP_COMMAND cargo test -p agent-relay-broker --quiet` — 853 passed, 4 ignored; integration crates also passed
- `cargo clippy -p agent-relay-broker --lib --bins -- -D warnings`
- `cargo fmt --all -- --check`
- targeted broker integration — missing provider rejected and absent from list
- workspace-resolution regression files — 6 files, 102 tests passed
- `npm run typecheck`
- `npm run lint` — no errors (existing warnings only)
- `npm run format:check`
- `npm test` — 1,733 passed; the two existing relayfile stale-daemon tests failed locally, matching the known base-branch issue

## Stack

- Base: #1425 (`fix/workspace-restore-rebind`) — Bug A
- This draft: Bug B process confirmation
- Retarget to `main` after #1425 merges

<a href="https://cubic.dev/pr/AgentWorkforce/relay/pull/1428?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `agent-relay node up` resolves its installed broker through canonical package-manager links and Relay's user install directories, so mise-managed and minimal-`PATH` launches no longer fail when the broker binary is already installed.
 - `agent-relay up` / `node up` use one precedence ladder: `--workspace-key` → workspace environment variables → repository pin → machine-global active workspace → creating one. A fresh project joins the active workspace instead of silently creating another, and startup announces the winning source.
 - Enrolled-node restarts preserve the repository-pinned workspace while resuming the enrolled identity. A conflicting enrollment stops startup, names both non-secret sources, and points to `workspace rebind <name>` as the recovery path.
+- `agent-relay node agent spawn` now verifies that the worker process survives startup before reporting success, and reports its exit status and log path when launch fails.
 - First-run telemetry notices are written to stderr so JSON stdout remains parseable.
 - Detached `node up --background` surfaces early child failures and stops polling when the child exits without trying to kill an already dead process.
 
--- a/crates/broker/src/worker.rs
+++ b/crates/broker/src/worker.rs
@@ -53,6 +53,12 @@ const APP_SERVER_RELEASE_GRACE: Duration = Duration::from_secs(35);
 /// healthy-but-slow agent is far worse than listing a dead one a little longer.
 const WORKER_READY_DEADLINE: Duration = Duration::from_secs(90);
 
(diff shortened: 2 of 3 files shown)


khaliqgant:
Closing as a duplicate of #1429.

Both PRs point at the same branch `fix/node-up-enrollment-precedence-and-spawn-reliability` at the same head `5d691f212`, opened 38 minutes apart on 2026-08-04. Neither has any review — the only comment on each is CodeRabbit's identical "Review skipped — Draft detected", so nothing is lost by consolidating.

Keeping #1429 because its description states the user-facing symptom directly (the CLI reporting `Spawned X` success while the spawn actually failed) rather than only the mechanism.

Worth flagging on the way past: **both were left as drafts, so no bot has ever reviewed this branch**, and it has sat untouched for two days. That fix is more valuable than it looks — the exact failure it describes cost real diagnosis time overnight on 2026-08-05, when a spawn onto finn-mini reported success and the PTY died instantly because the launchd `PATH` omitted the directory holding `claude`. With this merged, that would have surfaced as an immediate error instead of a healthy-looking agent that was never alive.

Also note the branch name mentions enrollment precedence, but the diff is spawn-acknowledgement only — it does **not** address #1432 (`node up` silently ignoring a fleet enrollment when a project workspace pin lacks `enrolledNodeId`). That issue still has no fix in flight.

PR #1429, opened 2026-08-04
fix(broker): verify worker process before spawn success

khaliqgant:
Fixes the silent \`node agent spawn\` failure hit repeatedly today across 3 machines: the CLI reports "Spawned X (codex, pty)." success while the spawned worker process has already exited (e.g. a wrapper that can't launch its harness) — \`Command::spawn\` only proves the wrapper was *created*, not that it survived.

Adds a brief (250ms) stability window after spawn that non-blockingly checks the child is still alive via \`try_wait()\` before reporting success. If it already exited, the stale registry entry is removed and the real error (exit status + log path) is returned instead of letting \`node agent list\` briefly advertise a dead process.

Companion fix for Bug A (workspace-enrollment precedence overriding an explicit \`RELAY_WORKSPACE_KEY\`) is being worked separately.

## Test plan
- [x] \`cargo test --package agent-relay-broker worker::\` — 62/62, 4 consecutive clean runs
- [x] Real integration test: \`RELAY_INTEGRATION_REAL_CLI=1 node --test tests/integration/broker/dist/cli-spawn.test.js\` — missing-CLI rejection path, 21.5s, real broker, confirms no stale agent is listed

<a href="https://cubic.dev/pr/AgentWorkforce/relay/pull/1429?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `agent-relay workspace switch|join` keeps the project's enrolled fleet node id instead of dropping it, which previously produced the pin state that made the next `node up` ignore the enrollment store.
 - `agent-relay up` / `node up` use one precedence ladder: `--workspace-key` → workspace environment variables → repository pin → machine-global active workspace → creating one. A fresh project joins the active workspace instead of silently creating another, startup announces the winning source, and `node status` reports the same five-source provenance.
 - Cloud enrollment selects node identity without overriding workspace resolution. A conflict with the repository pin stops startup, names both non-secret sources, and points to `workspace rebind <name>` as the recovery path.
+- `agent-relay node agent spawn` now verifies that the worker process survives startup before reporting success, and reports its exit status and log path when launch fails.
 - Detached `node up --background` surfaces early child failures and stops polling when the child exits without trying to kill an already dead process.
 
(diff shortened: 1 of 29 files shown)


khaliqgant:
Marked ready for review — it was left as a draft, so CodeRabbit skipped it ("Review skipped — Draft detected") and it sat unreviewed for two days. Duplicate #1428 closed in its favour.

Worth stating the value here, because the diff reads smaller than its impact: this fixes the CLI reporting `Spawned X` success while the spawn actually failed. That exact failure cost real diagnosis time overnight on 2026-08-05 — a spawn onto `finn-mini` reported success while the PTY died instantly, because the launchd `PATH` for the fleet-node service omitted the directory containing `claude`. From the outside it looked like a healthy agent that simply never did anything, which is the most expensive kind of failure to chase. With this merged it would have been an immediate, legible error.

Note for whoever reviews: the branch name says `enrollment-precedence`, but the diff is spawn-acknowledgement only. #1432 (`node up` silently ignoring a fleet enrollment when a project pin lacks `enrolledNodeId`) is a separate bug and is now being worked separately.

PR #1403, opened 2026-07-31
AR-448: durable workspace identity across node restarts

kjgbot:
## Summary

Anchors a local `agent-relay` node to one canonical Cloud workspace so Relaycast, Relayfile, and RelayAuth all resolve the same data-plane workspace on every start — and a resident agent like `khaliq-chief` retains the same delivery address/inbox rather than becoming a fresh process-lifetime identity across a stop/start.

- `agent-relay workspace active --json` now emits `canonical: true` and a per-plane `planes` map. Deploys can gate on `agent-relay workspace active --json | jq .canonical` rather than re-comparing per-plane IDs. Human output shows a single canonical ID when planes agree and a divergence warning otherwise.
- New `durable-workspace-identity.test.ts` regression pins the disk contract: `workspace-key.json` (and its `enrolledNodeId`) round-trip across a stop → start cycle, and the workspace-key mask is stable between start N and start N+1 for the same key.
- Spec `specs/durable-workspace-identity.md` documents the invariant, the on-disk sources of truth and their precedence (project pin > machine-global active workspace), node id derivation, the redaction contract, and migration behavior for existing local nodes.

## Acceptance evidence

- [x] `agent-relay workspace active --json` proves Relaycast/Relayfile/RelayAuth use one data-plane workspace ID (`canonical: true` + `planes` map). Covered by `workspace.test.ts` — including the diverged-planes case.
- [x] `agent-relay node up` automatically uses the canonical workspace — the pre-existing `resumePinnedProjectWorkspace` path in `runUpCommand` reads `workspace-key.json`. `durable-workspace-identity.test.ts` pins the file round-trip contract.
- [x] Stop/start regression test — `durable-workspace-identity.test.ts` covers workspace key preservation, enrolled-node preservation, and mask stability.
- [x] Node status and logs never print workspace keys, agent tokens, or credential-bearing observer URLs — `broker-lifecycle.ts` already masks via `maskSecret` in `up` and `status`, the observer URL was removed, and the new `durable-workspace-identity.test.ts` regression guards against the `?key=` URL reappearing.
- [x] Documented in `specs/durable-workspace-identity.md` + a new section in `packages/cli/README.md`. Migration section covers pre-existing `workspace-key.json` files, pre-existing machine seeds, and the startup-banner change (masked value in the same line).
- [ ] Do not merge — awaiting owner review.

## Test plan

- [x] `npx vitest run packages/cli/src/cli/lib/redact.test.ts` — 9 tests pass (existing + new fingerprint helpers)
[...]

PR #1447, opened 2026-08-07
test(cli,cloud): harvest AR-448 workspace-identity coverage onto main

khaliqgant:
Harvests the test value from #1402 onto current `main`. **Does not supersede or close #1402 or #1403** — that is Khaliq's call, after this is green.

## Why a fresh PR instead of merging #1402

AR-448's substance already shipped, through neither AR-448-branded PR:

- `4acdd97d4` — the shared workspace precedence ladder (PR #1429)
- `5c2ad8ee3` — broker restart-reclaim + hashed identity proof

Both are ancestors of `main`. The two AR-448-branded commits (`492515a20`, `27494c862`) are not. #1402 and #1403 are 7–8 days stale and both still edit `packages/cli/src/cli/commands/workspace.ts`, which #1429 rewrote underneath them.

So the implementation half of #1402 is redundant: `resolveWorkspaceSelection` in `packages/cloud/src/project-workspace-key.ts` already *is* the ladder AR-448 proposed. Reintroducing #1402's version would create a second code path for one behavior. Its **test** half is not redundant, and that is what this PR carries — rewritten against main's code, not copied.

## What this adds

**`packages/cli/src/cli/lib/workspace-identity-restart.test.ts`** — `broker-lifecycle.test.ts` covers the ladder one start at a time. This covers what only appears *across* starts:

- a second start lands on the same workspace, and the resident keeps its address
- the first start pins, so the second resumes from the repository pin rather than the store
- an explicit `--workspace-key` re-pins durably, so the move survives the *next* start
- a first start with no pin joins the canonical workspace instead of minting
- negative control: with no canonical workspace, a checkout that never pinned drifts
- a second checkout joins the same workspace but may **not** take the resident's name
- no key material is printed on any start in the sequence

**`packages/cloud/src/workspace-convergence.ts`** + test — the data-plane half of the invariant, which `main` had no equivalent of. Reports whether Relaycast, Relayfile, and RelayAuth resolve one `rw_…` id, withholding a shared id on divergence rather than papering over the split. The Cloud id is excluded deliberately (control-plane UUID, different id space).

**`workspace active`** — wired to that module rather than leaving it as a module nothing calls: a `dataPlane` block in `--json`, the Relaycast id in human output (previously the one plane omitted), and `--require-unified` to turn a divergence into a non-zero exit for setup doctors. Without `--require-unified` the exit code is unchanged, so existing scripted callers are unaffected.

**`specs/workspace-identity.md`** — rewritten against main's five-step ladder and the admission gate.

> Scope note: `workspace.ts` was not in the assigned file list. I included the minimal wiring because a pure module with tests and no consumer proves only that it exists, not that anything acts on it. Easy to drop if you'd rather it land separately.

## The premise correction

One #1402 assertion is not adapted but **reversed**, and it is the most important thing in this PR.

#1402 asserted that a second checkout sharing the canonical workspace also shares the resident's address — its Relaycast stand-in modeled re-registration as unconditionally returning the existing agent. `5c2ad8ee3` post-dates it by a week and makes that false by design. Per `admit_agent_registration` in `crates/broker/src/relaycast/auth.rs`, a name collision is **rejected** by default; reclaim requires proving the same work unit via an identity key compared as a SHA-256 hash, and the broker proves its own with `stable_node_identity_key` — hash of its persisted state directory, stable across a kill+restart, different for any other checkout.

**AR-448's premise was that workspace convergence is sufficient for address identity. It is necessary and not sufficient.** Agent-identity admission is a separate second half that landed a week later as its own fix. Checking only the workspace half is how a node that resolves the correct workspace can still fail to come back as itself. That is recorded in the spec, not just here.

The test's stand-in now models all three outcomes: grant a free name, reclaim on matching identity, reject on mismatch.

## Acceptance criteria

| # | AR-448 criterion | Status |
|---|---|---|
| 1 | `node up` joins the canonical workspace with no manual key copying | Covered — restart test + `broker-lifecycle.test.ts` |
| 2 | `workspace active --json` proves one data-plane workspace id | Covered — `workspace-convergence.test.ts` + `workspace.test.ts` |
| 3 | **Resident keeps address and mailbox across a live stop/start** | **Not covered — see below** |
| 4 | Status and logs never print credentials | Covered — restart test; `broker-lifecycle.test.ts` already covered the single-start case |
| 5 | Precedence: explicit > pin > canonical store, and no-store is unchanged legacy behavior | Covered — restart test + `broker-lifecycle.test.ts` (already on main) |
| 6 | Documented | `specs/workspace-identity.md` |

### Criterion 3 — what these tests do not buy

These are unit tests over the CLI's own TypeScript (`runUpCommand`, `resolveWorkspaceSelection`) with `createRelay` stubbed. **No broker binary participates, released or otherwise.** They prove the CLI *selects* the canonical workspace across starts. They do not prove identity *survives* a real restart. Those are different claims and a green run here is not evidence for the second.

The live proof is blocked on two independent things:

1. **It needs an operator at the keyboard.** Stopping the broker stops the resident agent that would perform the check.
2. **It cannot pass on a released broker today.** `5c2ad8ee3` is on `main` but in no release tag — it post-dates v11.4.0 (2026-08-02) and v11.4.1 (2026-08-03). `crates/broker/src/relaycast/auth.rs` does not exist in the v11.4.1 tree at all; the gate is main-only. **A live re-verification against an installed 11.4.0/11.4.1 broker is expected to fail for that reason, and is not evidence against these tests.**

## Test evidence

Run in a clean worktree off `origin/main` (`76232055c`).

```
$ npx vitest run packages/cli/src/cli/lib/workspace-identity-restart.test.ts

 RUN  v4.1.8

 Test Files  1 passed (1)
      Tests  8 passed (8)
   Duration  692ms
```

```
$ npx vitest run packages/cloud/src/workspace-convergence.test.ts \
    packages/cli/src/cli/commands/workspace.test.ts

 RUN  v4.1.8

 Test Files  2 passed (2)
      Tests  30 passed (30)
   Duration  795ms
```

Full `packages/cli` + `packages/cloud`:

```
$ npx vitest run packages/cli packages/cloud

 Test Files  80 passed | 2 skipped (82)
      Tests  1234 passed | 20 skipped (1254)
   Duration  10.71s
$ echo $?
0
```

`npm run typecheck` — exit 0, clean. `npm run lint` — exit 0, 76 pre-existing warnings, **none in the files this PR touches**.

### The tests were checked for bite, not just for green

Mutating the ladder to drop its machine-global store step (`return resolveActiveWorkspaceSelection(env)` → `return undefined`) fails 6 of the 8 restart cases:

```
 × keeps one workspace and one resident address across a restart
 × resumes from the repository pin on the second start, not the machine-global store
 × re-pins after an explicit --workspace-key so the next start resumes the new workspace
 × joins the canonical workspace rather than minting on a first start with no pin
 × lands a second checkout in the canonical workspace without anyone copying a key
 × refuses to hand the resident address to a different node claiming its name

 Tests  6 failed | 2 passed (8)
```

The two that still pass are the ones that should: the no-canonical-workspace negative control, and the credential-leak check. The mutation was reverted before commit.

<a href="https://cubic.dev/pr/AgentWorkforce/relay/pull/1447?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `agent-relay cloud login --device` logs in a machine with no browser through the OAuth device flow: the CLI prints a code you approve from any other device. Login and re-authentication fall back to it automatically over SSH or on a Unix host with no display server, and each machine gets its own cloud session instead of a copied `cloud-auth.json`. Requires cloud with the device authorization endpoints.
 - `agent-relay workspace restore` returns to the recorded previous workspace.
 - `agent-relay workspace rebind <name>` pins a project's next broker start to a named workspace without changing the machine-global active workspace.
+- `agent-relay workspace active` reports whether Relaycast, Relayfile, and RelayAuth resolve one data-plane workspace id: a `dataPlane` block in `--json`, the previously omitted Relaycast id in human output, and `--require-unified` to exit non-zero on a divergence.
 
 ### Changed
 
--- a/packages/cli/src/cli/commands/workspace.test.ts
+++ b/packages/cli/src/cli/commands/workspace.test.ts
@@ -233,9 +233,7 @@ describe('registerWorkspaceCommands', () => {
 
     await program.parseAsync(['node', 'agent-relay', 'workspace', 'active']);
 
-    expect(vi.mocked(deps.error).mock.calls.flat().map(String).join('\n')).toContain(
-      'relayfile=rw_b'
-    );
+    expect(vi.mocked(deps.error).mock.calls.flat().map(String).join('\n')).toContain('relayfile=rw_b');
     // Existing scripted callers keep their exit code; only the explicit gate
     // below changes it.
     expect(deps.exit).not.toHaveBeenCalled();
--- a/packages/cli/src/cli/commands/workspace.test.ts
+++ b/packages/cli/src/cli/commands/workspace.test.ts
@@ -1,12 +1,20 @@
 import { Command } from 'commander';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
-vi.mock('@agent-relay/cloud', () => ({
-  readWorkspaceStore: vi.fn(() => ({ workspaces: {} })),
-  resolveActiveWorkspace: vi.fn(),
-  setWorkspaceKey: vi.fn(),
-  switchWorkspace: vi.fn(),
-}));
+vi.mock('@agent-relay/cloud', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('@agent-relay/cloud')>();
+  return {
+    readWorkspaceStore: vi.fn(() => ({ workspaces: {} })),
+    resolveActiveWorkspace: vi.fn(),
+    setWorkspaceKey: vi.fn(),
+    switchWorkspace: vi.fn(),
+    // The real convergence helpers, not stand-ins: these tests assert on what
+    // the command reports about the AR-448 data-plane invariant, so a copy here
+    // would let the real check drift past them.
+    describeDataPlaneConvergence: actual.describeDataPlaneConvergence,
+    formatDataPlaneDivergence: actual.formatDataPlaneDivergence,
+  };
+});
 
 vi.mock('../lib/workspace-session.js', async (importOriginal) => ({
   // Returns a result object describing what the write changed beyond the key.
@@ -106,6 +114,13 @@ describe('registerWorkspaceCommands', () => {
       slug: 'ops',
       urls: {},
       apiUrl: 'https://cloud.test',
+      // This fixture's planes genuinely disagree (rc_ops vs rw_ops), so the
+      // evidence block reports the split rather than a shared identity.
+      dataPlane: {
+        unified: false,
+        planes: { relaycast: 'rc_ops', relayfile: 'rw_ops', relayauth: 'rw_ops' },
+        divergent: ['relayfile', 'relayauth'],
+      },
     });
   });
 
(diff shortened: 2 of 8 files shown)


khaliqgant:
Closing on Khaliq's decision. The premise behind this PR was mine and it was wrong.

**Why.** This branch existed to harvest test evidence onto `main` on the belief that AR-448's behaviour had landed in #1429 untested. It had not. #1429 shipped the precedence ladder *with* coverage — `broker-lifecycle.test.ts` +104, `node.test.ts` +66, and a new `project-workspace-key.test.ts` +54. I recommended the harvest without checking whether main was already covered.

**What review then removed from what was left.** Two independent reviewers found the headline assertion — "resident keeps its address across a restart" — compares the test's own `relaycast.register` mock to itself, since `runUpCommand` only resolves the workspace key. Investigation confirmed it worse than reported: under mutation the test died on the *first* assertion (line 295, workspace key) and line 300 never ran at all. Deleting the key assertions and re-running under the same mutation showed the address assertion passing. **A mutation count is not coverage — a test that dies early hides every assertion behind the failure.** The convergence test had the same shape: `createMachine` allocates a separate `relayHome` per checkout, so "two checkouts converge on one machine-global store" was really "two isolated stores were seeded with the same literal string."

**On `--require-unified`, dropped for the right reason.** Not because the convergence invariant is fake — it is real, it is asserted by this repo's own doctor on every run, and the live workspace resolves Relaycast, Relayfile and RelayAuth to one identical `rw_` ID. It is dropped because it duplicates a check the doctor already owns, at the wrong layer, and shipped a double-exit bug on its failure path (`deps.exit(1)` throwing `CliExit` inside `runSdk`, which re-prints and exits again). Recorded because the reasoning that nearly justified keeping it — that main's `workspace active` fixture uses `rc_ops` against `rw_ops` and therefore the invariant must not be real — was a generalisation from arbitrary test data to production, and it is wrong.

**What survives, and it is the valuable part.** #1402's model encoded *"re-registering a name in a workspace it already belongs to returns the existing agent"*, unconditionally. `b4b96dfb3` made that fail closed: a name collision is rejected unless the caller proves same-work-unit via a hashed identity key. So **workspace convergence is necessary and not sufficient for address identity** — identity has a second half in the admission decision on the name, which landed a week after AR-448 as its own fix. That correction is recorded independently of this branch.

Nothing is lost. The branch `test/ar-448-workspace-identity-coverage` remains at `2c48ba9ca` if anything here is wanted later. One piece is being taken forward separately on its own merits: coverage asserting that a key passed as `--workspace-key`, and the key the broker returns, stay out of the logs — neither is covered by main's existing assertions at `broker-lifecycle.test.ts:643-644` or `:466`.

PR #1465, opened 2026-08-08
Add broker commit trailer hook support

miyaontherelay:
Adds optional spawn metadata for commit trailers and preserves repository-local prepare-commit-msg hooks.\n\nTests cover trailer injection, hook chaining, and the metadata-absent no-op path.\n\nReference: https://github.com/AgentWorkforce/sales/issues/27

### DIFF
--- a/crates/broker/src/relaycast/bridge.rs
+++ b/crates/broker/src/relaycast/bridge.rs
@@ -726,6 +726,43 @@ mod tests {
                 name: "Worker1".into(),
                 cli: "codex".into(),
                 args: vec!["--full-auto".into()],
+                metadata: crate::types::SpawnMetadata::default(),
+            })
+        );
+    }
+
+    #[test]
+    fn maps_spawn_attestation_metadata_from_the_dispatch_input() {
+        let (_, payload) = map_action(
+            &json!({
+                "type": "action.invoked",
+                "action_name": "spawn",
+                "invocation_id": "inv_attestation",
+                "caller_name": "147298826957365248",
+            }),
+            json!({
+                "name": "WorkerAttested",
+                "cli": "codex",
+                "metadata": {
+                    "attestation": {
+                        "jti": "jti-public-123",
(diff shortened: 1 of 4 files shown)


khaliqgant:
Surfacing some context on this PR, since it is currently blocked only on human review and it turns out to sit on a critical path that is not obvious from the diff.

**Current state, verified:** all 9 workflows green on `de6215a5c` (checked per-workflow via `gh run list --branch`, not the rollup). `mergeStateStatus: BLOCKED`, `reviewDecision: REVIEW_REQUIRED` — so this is waiting on a reviewer, not on engineering. Open since 08-08.

**One unresolved thread**, a cubic P2 at `spawner.rs:54`: configured hooks using `~user/...` or a bare `~` are skipped, because Git expands those paths while this chain treats them as repository-relative. Resolving `core.hooksPath` in Git's `--path` mode before selecting the hook directory would cover all tilde forms. Real but narrow.

**Why this matters beyond the diff.** `RELAY_ATTEST_*` is often described as planned work. It is not — it is implemented here, and only here. That makes this PR the mechanism by which an agent-produced commit can carry provenance at all.

**One gap worth naming while the design is still open:** this stamps `Agent-Id` and `Sponsor-Id` but **no `Session-Id`**. So a commit can be attributed to an *agent* but not to the *conversation* that produced it. If the intent is a traceable path from a merged change back to the discussion that led to it, that is the missing field — and the mechanism to carry it already exists here rather than needing to be built.

Worth noting the ordering it implies: the broker consumes a `jti` at spawn time, when no commit exists yet. So the identifier has to be minted before the commit and finalised against the SHA afterwards — a two-phase shape. Anything else riding the same channel inherits it.

**Also, a small housekeeping point:** the PR body links a private epic from this public repo, so external readers get a 404 pointer. No detail leaks in the link itself.

Not requesting changes and not merging — flagging that this is cheaper to land than it looks and is load-bearing for work downstream of it.

PR #1505, opened 2026-08-13
security: close sponsor-registration authority bypass

khaliqgant:
## 🛑 DO NOT MERGE

1. **Depends on `AgentWorkforce/relaycast#324`, which is an unmerged draft with zero reviews.** This is not a soft dependency — see "Dependency — hard merge blocker" below.
2. **`crates/broker/Cargo.toml` pins `relaycast` at the tip of that draft branch** (`a87e6e7fbbca4d1da453f9ab681f0e2ae86f70c2`). That commit is not on relaycast's `main` and can be rebased or force-pushed away at any time. It must be repinned to a released version or a commit reachable from relaycast `main` before this can merge.
3. **The authority check is fail-open.** Every entry point in relaycast#324 (`authorizeNewAgentCredential`, `authorizeExistingAgentCredential`, `bindLegacyAgentCredential`) starts with `if (!enforced) return { mode: 'unenforced' }`. Absent both `RELAYCAST_AGENT_CREDENTIAL_AUTHORITY_PUBLIC_KEY_PEM` and `RELAYCAST_AGENT_CREDENTIAL_AUTHORITY_ISSUER` configured on the deployed relaycast server, merging and deploying this PR enforces nothing — silently, with no error or warning. Green CI and `mergeable=MERGEABLE` on this PR do not mean the security boundary is live.
4. **Production cannot currently mint a sponsor proof at all — this is not a 15-minute-expiry problem, it is a cannot-produce-one-in-the-first-place problem, for two independent reasons, either alone sufficient:**
   - Cloud IaC never binds `RELAYAUTH_SPONSOR_FEDERATIONS` (`infra/relayauth.ts:106-136,145-285` in the RelayAuth repo — the key is simply absent from the Worker env construction). Absent that key, RelayAuth resolves every org to `LEGACY` mode (`sponsor-binding.ts:152-175`), and `POST /sponsors/proof` rejects any request when the org's mode is not `oidc` (`routes/sponsors.ts:65-73`).
   - There is no `id_token` source to feed it even if that were fixed: Cloud's Google auth requests `access_type=online` and retains only an `access_token`, never an OIDC `id_token` (`packages/web/lib/auth/google.ts:17-33,36-63`; callback route `:60-73`), and `/proof` requires a recent verified `id_token`.
   - Cloud's broker env construction (`box-manager.ts:2223-2284`, passed at `:2037-2065` / `:2126-2137`) contains no sponsor-proof field, and there is no scheduled remint or broker-reload path anywhere in relaycast-cloud main. `mint-relayauth-api-key.yml` mints a long-lived API key — never a sponsor proof.
   - **Consequence:** with this merged and deployed, brokers cannot spawn agents at all, from the first spawn, independent of relaycast#324, relaycast-cloud#60, secrets provisioning, or the arming sequence discussed elsewhere in this body.
   - **Prerequisite, stated plainly:** either a durable one-time broker binding that later authorizes with the incumbent scoped credential — the pattern RelayAuth uses for its own identities, consuming the sponsor proof only at identity creation — or an automated fresh-`id_token`-to-proof remint plus broker reload. Neither exists today.
   - **Open item, not resolved either way:** a manually-added Cloudflare Worker binding for `RELAYAUTH_SPONSOR_FEDERATIONS` would not appear in IaC and cannot be ruled out from any repo. Determining whether one exists needs read-only inspection of the live Worker bindings — that's Khaliq to run or authorize, not something checkable from source.
   - **This PR already contains the working pattern for its own broker identity**, applied only to the broker's own credential, not to the agents it spawns: `IncumbentCredentialCache`, `persist_incumbent_credential_cache`, `incumbent_token_for`, and `startup_session_set_with_identity_and_incumbents` in `crates/broker/src/relaycast/auth.rs`. The expiring-grant pattern above is what this PR applies to spawned agents; the durable-binding pattern already exists in the same file for the broker itself. A reader should not have to rediscover that both shapes are present.

## Security boundary fixed

This is the coordinated Relay-side security fix for #1497. It does **not** treat the Relay `AuthClient` as the authority. It pins the matching Relaycast authority change (`AgentWorkforce/relaycast#324`) and passes a signed sponsor grant plus a secret work-unit key on every credential-issuing path:

- normal REST registration and rotation
- `agent-relay mcp-args --register`
- broker child pre-registration
- node-control registration
- A2A registration
- destructive release aliases are enforced by the pinned server authority

The authority rejects workspace-key-only calls and stores sponsor/work-unit ownership in server-controlled immutable state, not agent metadata. Client processes no longer inherit sponsor proofs, signing material, workspace keys, or the broker root work-unit secret.

## Dependency — hard merge blocker

Depends on `AgentWorkforce/relaycast#324` and must not merge before it. This is mechanical, not a judgement call: `crates/broker/Cargo.toml` pins `relaycast` as a git dependency at rev `a87e6e7fbbca4d1da453f9ab681f0e2ae86f70c2`, which is the current tip of relaycast#324 — a **draft, unmerged, zero-review** branch in a separate repo. That commit is not reachable from relaycast's `main` and can be rebased or force-pushed out of existence at any time. Merging #1505 while pinned this way would put relay's `main` on a dependency that only exists on someone else's feature branch.

Required sequence: relaycast#324 merges and is released (a real published `relaycast` version, or at minimum a commit reachable from relaycast `main`) → #1505 repins `crates/broker/Cargo.toml` off the draft-branch git rev → #1505 merges. Do not work around this by vendoring the crate or pinning harder against the draft branch.

## Findings addressed

- Direct REST/CLI/node/A2A calls cannot bypass sponsor enforcement once the authority PR is deployed.
- Caller-editable metadata is no longer an ownership input.
- Relay pins Relaycast authority commit `a87e6e7fbbca4d1da453f9ab681f0e2ae86f70c2`, including database mutation-boundary guards against stale concurrent claim decisions.
- Missing/expired sponsor proofs fail before workspace creation or any startup request.
- Same-name reclaim uses a random persisted work-unit secret, not a predictable state-path digest.
- Existing legacy agents have a guarded migration: an incumbent agent token is required for the one-time server binding; a workspace key is insufficient.
- Both client and authority reject signed grants beyond RelayAuth’s 15-minute maximum lifetime.
- CI mints a real short-lived RS256 test sponsor grant and starts the pinned local authority with its public key. There is no test bypass.

## Legacy migration rollout (required order)

Older clients did not retain their incumbent agent token, so deployment must be staged:

1. Release/deploy this client while the old registration authority is still active.
2. Restart every persistent broker once. Successful startup atomically persists its scoped token beside broker state. Selection is bound to a SHA-256 workspace-key fingerprint plus exact agent name. The token and random work-unit key are owner-only on Unix and DPAPI-encrypted to the current user on Windows.
3. Verify pre-staging across the fleet.
4. Publish the matching Relaycast packages, apply the authority DB migration, and enable pinned sponsor-verification configuration.
5. On the next restart, the client authenticates `/v1/agent` with the incumbent token and invokes the one-time server binding. It does not rotate via the workspace key.

A host that skips pre-staging must be given its incumbent `RELAY_AGENT_TOKEN` explicitly. It fails closed instead of reopening workspace-key-only reclaim.

## CI/test changes

- Shared action checks out the exact authority commit and generates an ephemeral RSA key pair/grant in the runner temp directory.
- E2E, package validation, macOS smoke, and fleet jobs receive valid test sponsor credentials.
- Fleet harness fails fast with broker stderr and health context instead of waiting for a secondary symptom.

## Verification

- `cargo test --workspace`: broker 942 passed / 4 ignored; broker integration 16 passed; relay-pty 223 passed; doc test ignored
- auth regression suite: 40/40 passed, including a true two-start old-server → enforced-server migration
- `cargo clippy -p agent-relay-broker --lib -- -D warnings`: passed
- `cargo fmt --all -- --check`: passed
- Node typecheck/lint/build/Vitest and live E2E were run earlier on this branch: 1,884 passed / 23 skipped; live E2E 27 passed / 4 skipped
- full `cargo clippy --workspace --all-targets -- -D warnings` currently hits an unrelated pre-existing test-only lint in `crates/relay-pty/src/pty.rs:2130` (`cloned_ref_to_slice_refs`); production broker lint is green
- local macOS→Windows cross-check reaches the pre-existing `ring` native build and stops for lack of Windows C headers; GitHub's native Windows package gate is the authoritative DPAPI compile check

## Deployment blockers / residual boundary

Do not merge/deploy this in isolation:

- Requires `AgentWorkforce/relaycast#324` (see Dependency section above — currently draft/unmerged, this alone blocks merge) and `AgentWorkforce/relaycast-cloud#60`.
- relaycast-cloud currently needs the matching Relaycast packages published before its normal lockfile CI can go green.
- **The authority check fails open, silently, if unconfigured.** In relaycast#324, every entry point (`authorizeNewAgentCredential`, `authorizeExistingAgentCredential`, `bindLegacyAgentCredential`) begins with `const enforced = authorityConfig(config); if (!enforced) return { mode: 'unenforced' }`. Absent both `RELAYCAST_AGENT_CREDENTIAL_AUTHORITY_PUBLIC_KEY_PEM` and `RELAYCAST_AGENT_CREDENTIAL_AUTHORITY_ISSUER` in the deployed relaycast environment, registration/rotation/reclaim enforcement is a complete no-op — any caller can register, rotate, or reclaim any agent — with **no error, warning, or other signal that enforcement is inert**. Confirming those two vars are actually set in the production relaycast deployment is a precondition for this PR closing the hole in practice, not just in code, and is being tracked separately with whoever owns that deployment.
- The deprecated `api.relaycast.dev` / `gateway.relaycast.dev` legacy authorities are a separate deployment boundary from canonical `cast.agentrelay.com`; they need explicit cutover/decommission or equivalent server enforcement before anyone claims all hosted endpoints are closed.
- Sponsor grants expire. Long-running brokers need Chief/RelayAuth to refresh the grant before a later credential-issuance event.

No merge or deployment is performed by this PR.

## Relationship to #1497 — guarantee-by-guarantee

#1497 (`agent/soc2-hole1-sso-sponsor`) targeted the same vulnerability class via a different, client-side architecture. #1497 will be closed in favor of this PR; this section is the record of what that decision does and does not carry forward, so a future reader doesn't have to reconstruct it from two diffs.

- **G1 — registration requires an authenticated SSO sponsor** (a workspace key alone is insufficient). CARRIED — #1505 has a near-identical `require_authenticated_sponsor()` gate with the same error text.
- **G2 — the sponsor proof must be a cryptographically valid, unexpired signed JWT**, verified against a configured public key (issuer/audience/subject/org/intent/token_type checked). CARRIED — #1505's `verify_sponsor_proof` is effectively the same check, same claim shape, same constants.
- **G3 — the sponsor id must be human-shaped** (`user_` prefix pattern), not a service/workspace id. CARRIED — identical `is_human_sponsor_id` check.
- **G4 — registrations stamp the agent with sponsor identity for later comparison.** DELIBERATELY REPLACED. #1497 wrote `relayauth_sponsor_id` / `relayauth_sponsor_binding` / `relayauth_sponsor_proof_sha256` into client-visible, client-writable agent metadata. #1505 stores the equivalent binding (`sponsorOrgId`, `sponsorId`, `sponsorOidcIssuer`, `sponsorOidcSubject`, `workUnitKeyHash`) as immutable server-side state in relaycast's own DB, and the relaycast#324 server actively rejects any client attempt to set those metadata keys (`RESERVED_METADATA_KEYS`). This is the correct direction (client-writable metadata was never trustworthy ownership data) but only takes effect once relaycast#324 is merged, deployed, and configured — see "DO NOT MERGE" above.
- **G5 — same-name reclaim must match the existing agent's sponsor, not just its work-unit identity.** **G6 — token rotation must refuse a caller authenticated as a different sponsor than the one the agent is bound to.** NEITHER IS ENFORCED ANYWHERE TODAY, by either PR. #1497's version of G5/G6 is a client-side courtesy check inside the relay CLI: it does `GET /v1/agents/:name`, compares metadata to the locally-authenticated sponsor, and only then calls `rotate_agent_token`/`register_agent` — but that underlying wire call itself carries no sponsor proof and is authenticated only by workspace key, so any caller that isn't this specific CLI (curl, another SDK, a modified client) rotates or reclaims across sponsor boundaries today unimpeded, regardless of what #1497 does. It was a convention observed by one client, not a boundary. #1505 does not reimplement it client-side — deliberately: relaycast#324's server now withholds sponsor-binding data from the client-visible agent-fetch response specifically to stop this class of client-side check from being spoofed, so #1505's client has no data left to check even if it wanted to. G5/G6 will only be real once `AgentWorkforce/relaycast#324` is merged, deployed, **and** configured with `RELAYCAST_AGENT_CREDENTIAL_AUTHORITY_PUBLIC_KEY_PEM` + `RELAYCAST_AGENT_CREDENTIAL_AUTHORITY_ISSUER` (see "DO NOT MERGE" above) — that work is server-side, tracked in relaycast#324, not in this PR.

### DIFF
--- a/.github/actions/setup-relaycast-ci/action.yml
+++ b/.github/actions/setup-relaycast-ci/action.yml
@@ -5,7 +5,7 @@ inputs:
   engine-ref:
     description: Immutable Relaycast commit containing server-side credential authority
     required: false
-    default: d1fd18a817466debdf4b3a8460ee071e4b321ac4
+    default: 717ca65a1cb7d4726babef4d8a6de45a1fe3d34b
 
 runs:
   using: composite
--- a/.github/actions/setup-relaycast-ci/action.yml
+++ b/.github/actions/setup-relaycast-ci/action.yml
@@ -5,7 +5,7 @@ inputs:
   engine-ref:
     description: Immutable Relaycast commit containing server-side credential authority
     required: false
-    default: 717ca65a1cb7d4726babef4d8a6de45a1fe3d34b
+    default: a87e6e7fbbca4d1da453f9ab681f0e2ae86f70c2
 
 runs:
   using: composite
--- a/.github/actions/setup-relaycast-ci/action.yml
+++ b/.github/actions/setup-relaycast-ci/action.yml
@@ -5,7 +5,7 @@ inputs:
   engine-ref:
     description: Immutable Relaycast commit containing server-side credential authority
     required: false
-    default: da7670d7034d3c5797739dfb526a075c94ce3c82
+    default: d1fd18a817466debdf4b3a8460ee071e4b321ac4
 
 runs:
   using: composite
--- a/.github/actions/setup-relaycast-ci/action.yml
+++ b/.github/actions/setup-relaycast-ci/action.yml
@@ -1,0 +1,57 @@
+name: Setup sponsor-enforced Relaycast
+description: Build a pinned local Relaycast and mint an ephemeral CI sponsor grant it trusts
+
+inputs:
+  engine-ref:
+    description: Immutable Relaycast commit containing server-side credential authority
+    required: false
+    default: 0b29cc36a3df6ef2c715d6f5e097155da35cedd5
+
+runs:
+  using: composite
+  steps:
+    - name: Checkout sponsor-enforced Relaycast
+      uses: actions/checkout@v4
+      with:
+        repository: AgentWorkforce/relaycast
+        ref: ${{ inputs.engine-ref }}
+        path: .ci/relaycast-engine
+
+    - name: Build local Relaycast engine
+      shell: bash
+      working-directory: .ci/relaycast-engine
+      run: |
+        npm ci
+        npm run build --workspace @relaycast/types
+        npm run build --workspace @relaycast/a2a
+        npm run build --workspace @relaycast/engine
+
+    - name: Mint ephemeral sponsor grant
+      shell: bash
+      run: node scripts/mint-ci-sponsor-grant.mjs
+
+    - name: Start sponsor-enforced local engine
+      shell: bash
+      run: |
+        ENGINE_PORT="$((42000 + RANDOM % 10000))"
+        ENGINE_LOG="$RUNNER_TEMP/relaycast-ci-engine.log"
+        node .ci/relaycast-engine/packages/engine/dist/bin/serve.js \
+          --port "$ENGINE_PORT" \
+          --db "$RUNNER_TEMP/relaycast-ci-engine.db" \
+          --env test >"$ENGINE_LOG" 2>&1 &
+        ENGINE_PID=$!
+        for attempt in $(seq 1 60); do
+          if curl --silent --fail "http://127.0.0.1:${ENGINE_PORT}/health" >/dev/null; then
+            echo "RELAY_BASE_URL=http://127.0.0.1:${ENGINE_PORT}" >> "$GITHUB_ENV"
+            echo "RELAYCAST_BASE_URL=http://127.0.0.1:${ENGINE_PORT}" >> "$GITHUB_ENV"
+            echo "RELAYCAST_CI_ENGINE_PID=${ENGINE_PID}" >> "$GITHUB_ENV"
+            exit 0
+          fi
+          if ! kill -0 "$ENGINE_PID" 2>/dev/null; then
+            cat "$ENGINE_LOG" >&2
+            exit 1
+          fi
+          sleep 0.25
+        done
+        cat "$ENGINE_LOG" >&2
+        exit 1
--- a/.github/actions/setup-relaycast-ci/action.yml
+++ b/.github/actions/setup-relaycast-ci/action.yml
@@ -5,7 +5,7 @@ inputs:
   engine-ref:
     description: Immutable Relaycast commit containing server-side credential authority
     required: false
-    default: 0b29cc36a3df6ef2c715d6f5e097155da35cedd5
+    default: da7670d7034d3c5797739dfb526a075c94ce3c82
 
 runs:
   using: composite
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -10,7 +10,9 @@ on:
       - 'Cargo.toml'
       - 'Cargo.lock'
       - 'scripts/e2e-test.sh'
+      - 'scripts/mint-ci-sponsor-grant.mjs'
       - 'package.json'
+      - '.github/actions/setup-relaycast-ci/**'
       - '.github/workflows/e2e-tests.yml'
   pull_request:
     branches: [main]
@@ -21,7 +23,9 @@ on:
       - 'Cargo.toml'
       - 'Cargo.lock'
       - 'scripts/e2e-test.sh'
+      - 'scripts/mint-ci-sponsor-grant.mjs'
       - 'package.json'
+      - '.github/actions/setup-relaycast-ci/**'
       - '.github/workflows/e2e-tests.yml'
   # Allow manual trigger for on-demand testing
   workflow_dispatch:
@@ -55,6 +59,9 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
           cache: 'npm'
+
+      - name: Setup sponsor-enforced local Relaycast
+        uses: ./.github/actions/setup-relaycast-ci
 
       - name: Install dependencies
         run: npm ci
--- a/.github/workflows/fleet-e2e.yml
+++ b/.github/workflows/fleet-e2e.yml
@@ -67,7 +67,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: AgentWorkforce/relaycast
-          ref: eb7563ffcf0e54698c23363c5753060f99d37dd3 # relaycast v7.0.0 (node providers + unreported-load heartbeat compatibility)
+          ref: 0b29cc36a3df6ef2c715d6f5e097155da35cedd5 # sponsor-enforced credential authority + node providers
           path: relaycast-engine
 
       - name: Setup Node.js
--- a/.github/workflows/fleet-e2e.yml
+++ b/.github/workflows/fleet-e2e.yml
@@ -67,7 +67,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: AgentWorkforce/relaycast
-          ref: 0b29cc36a3df6ef2c715d6f5e097155da35cedd5 # sponsor-enforced credential authority + node providers
+          ref: da7670d7034d3c5797739dfb526a075c94ce3c82 # sponsor-enforced credential authority + claim race guards
           path: relaycast-engine
 
       - name: Setup Node.js
--- a/.github/workflows/fleet-e2e.yml
+++ b/.github/workflows/fleet-e2e.yml
@@ -67,7 +67,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: AgentWorkforce/relaycast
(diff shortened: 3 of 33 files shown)


miyaontherelay:
Closing this PR. The exposure it was opened to close is **still open**. This comment is the record of why, what remains exposed, and what to reuse — not a tombstone for a closed tab.

## 1. Why this cannot ship as designed

- The only RelayAuth sponsor grant is `token_type="sponsor_grant"`, default TTL 300s, **hard max 900s**, derived from a verified OIDC `id_token`, sponsor id forced to `/^user_/` — relayauth `packages/server/src/lib/sponsor-binding.ts:15-18, 253-268, 289-320, 593-605`.
- No service or machine grant exists. `IdentityType` includes `"service"`, but OIDC-bound identity creation demands the same `sponsorProof` for every type — `routes/identities.ts:491-505, 525-539`.
- No renewal path. The only mint surface is `POST /v1/sponsors/proof`; the SDKs expose only `createSponsorProof` — `routes/sponsors.ts:18-96`. Re-minting requires a fresh human OIDC `id_token` — there is nothing a long-running process can call on its own.
- **Production cannot mint a sponsor proof at all today.** Cloud IaC never binds `RELAYAUTH_SPONSOR_FEDERATIONS` (`infra/relayauth.ts:106-136, 145-285`); absence resolves every org to legacy mode (`sponsor-binding.ts:152-175`), and `/proof` rejects any request when mode is not `oidc` (`sponsors.ts:65-73`). Independently, Cloud's Google auth requests `access_type=online` and retains no `id_token` at all (`packages/web/lib/auth/google.ts:17-33`), so there's no token to feed `/proof` even if the federation were configured.
- **The root error**, stated as such rather than as an edge case: this design re-checks the raw sponsor proof **per spawn**, while RelayAuth's own identity model checks it **once at creation** and persists the resulting `sponsorBinding` (`identities.ts:491-505`, `tokens.ts:154-209`). Per-operation use of a binding-time credential is the mistake here — not the sponsor-binding concept itself, which is correct and should be kept.

## 2. What remains exposed

Closing this PR does not close the hole #1497 also didn't close. Specifically, today:

- `POST /v1/agents/:name/rotate-token` is still authenticated by the **workspace API key alone**. It carries no sponsor proof on the wire.
- There is still **no audit trail** of who released or rotated an agent's credential.
- Live case from this morning: the chief seat was released at `07:25:14Z` with reason `null`, and no CLI surface anywhere can say what did it.

## 3. What's salvageable — reuse this, don't re-derive it

- The fleet-wire carrier: `AgentRegister.registration_authority` (`crates/broker/src/fleet_wire.rs:242`), populated and fail-closed at both real call sites (`crates/broker/src/runtime/api.rs`, `crates/broker/src/runtime/relaycast_events.rs`).
- Guarantees G1-G3 from this PR (authenticated-sponsor requirement, cryptographic proof verification, human-shaped sponsor id) — all sound, all reusable independent of the per-spawn re-check design flaw above.
- **`IncumbentCredentialCache`** (`crates/broker/src/relaycast/auth.rs`) — already the bind-once-and-persist pattern the correct design needs, just applied to the broker's own identity rather than to the agents it spawns. This is the shape a fix should generalize, not reinvent.
- Branch: `agent/soc2-hole1-security-fix-0813` — kept for resurrection, not deleted.

## 4. Cross-references

`AgentWorkforce/relaycast#324` and `AgentWorkforce/relaycast-cloud#60` remain the tracked dependencies for the server-side authority work; both are independently blocked as described above and in this PR's body history. `AgentWorkforce/relaycast#325` is unrelated to sponsor machinery and stays open/priority.

miyaontherelay:
Cross-reference, added after the fact: relaycast#328 (open) — a message-trigger authority gap found while reviewing relaycast#324, independent of the sponsor-authority design this PR and #324/relaycast-cloud#60 retired. Filed separately since it's a gap in relaycast `main` today, not part of the retired design.

Issue #1511, opened 2026-08-14
Broker cannot be stopped while an agent registration is in flight against a slow or unreachable engine


khaliqgant:
A broker whose engine stops answering cannot be stopped. Two independent defects compose to produce it, and both are on `main` today — no feature flag, no new code required to arm them.

The practical consequence: `agent-relay node down` times out, the operator or supervisor escalates to `SIGKILL`, and `SIGKILL` on a node broker takes down **every agent spawned under it**, because spawned agents are its children. An engine that is merely slow or briefly unreachable is enough.

## H1 — an in-flight agent registration wedges the runtime event loop

`ListenApiRequest::Spawn` is handled by `BrokerRuntime::handle_api_request` (`crates/broker/src/runtime/api.rs:244`, `Spawn` arm at `:290`), which awaits `register_node_agent_token` at `:384`. That registration is bounded only by `FLEET_AGENT_REGISTER_TIMEOUT = 30s` (`crates/broker/src/runtime/fleet.rs:17`, awaited at `:1484`).

`handle_api_request` is itself awaited inside the runtime's `tokio::select!` (`crates/broker/src/runtime/event_loop.rs:336`). While that await is pending the `select!` cannot poll any other arm — including the SIGTERM arm (`event_loop.rs:317`, handled at `:332`).

So for up to 30 seconds per spawn, the broker answers **no** listen-API request and **cannot observe SIGTERM**. This is not specific to any one cause of a slow registration: an unreachable engine, a network partition, an overloaded engine, or a DNS failure all produce it.

## H2 — graceful shutdown has no timeout on its network calls

Even once SIGTERM is observed, `shutdown_runtime` (`crates/broker/src/runtime/event_loop.rs:448`) awaits, with no timeout:

- `relaycast_http.mark_agent_offline(&worker_name)` for every active worker — `:462-471`
- `relaycast_http.mark_offline()` — `:474-476`

Against an unresponsive engine these never return, so graceful shutdown never completes regardless of H1. Fixing H1 alone does **not** make the broker stoppable.

## Evidence

Paired arms, one variable: whether the engine answers. The engine is `SIGSTOP`ped **after** the node is online, so the node-control socket stays open and `agent.register` is simply never answered.

| | healthy engine | engine frozen mid-session |
|---|---|---|
| `node agent spawn` | exited 0, `Spawned probe-agent (codex, pty).` | still running |
[...]

PR #1514, opened 2026-08-14
fix(fleet): stop redundant client-side merge on declared metadata publish

miyaontherelay:
## Defect

`#1504` (merged as `592d371a8`) introduced `RelaycastHttpClient::publish_declared_metadata` (`crates/broker/src/relaycast/ws.rs:159-191`), which does a client-side read-merge-write: `get_agent` → merge declared fields over the existing metadata → `update_agent` (PATCH).

This is redundant and worse: `PATCH /v1/agents/:name` already merges metadata server-side (shallow merge, confirmed against `packages/engine/src/routes/agent.ts:291`). Because this call runs detached from the spawn (fire-and-forget `tokio::spawn`, by design — see `spawn_declared_metadata_publish`), there is a real window between the client's `get_agent` read and its `update_agent` write in which another write can land (the node bind, the engine's own `fleet` placement record). The client's PATCH — built from the earlier, now-stale snapshot — silently clobbers that intervening write.

A test at `ws.rs:973` (`publish_declared_metadata_merges_over_engine_owned_fields`, pre-fix name) encoded the read-merge-write as intended behavior rather than testing for its absence.

## Fix

Send only the declared keys in the PATCH body and let the engine's server-side merge preserve everything else. No read, no stale-snapshot window.

- `publish_declared_metadata` no longer calls `get_agent` before `update_agent`.
- Test renamed to `publish_declared_metadata_sends_only_the_declared_keys`; the mock GET now returns 500 and the test asserts it is never hit (`read.assert_hits(0)`), and the PATCH body is matched exactly against only the declared fields (no `cli`, no `fleet.node`).
- `tests/e2e/fleet/fleet-e2e.test.ts` updated accordingly (not re-run in this verification pass — needs the full broker/engine e2e harness).

## Verification

- `cargo test -p agent-relay-broker --lib --no-run`: clean build, no new warnings.
- `relaycast::ws::tests` suite: 7 passed, 0 failed, 2 pre-existing unrelated ignores (DM-mode fixture mismatches predating this branch).
  - `publish_declared_metadata_sends_only_the_declared_keys` — passes, confirms no GET occurs and the PATCH body carries only declared keys.
  - `publish_declared_metadata_makes_no_request_when_nothing_is_declared` — passes, no regression.
- Full broker lib suite and the fleet e2e test were not run in this pass (out of scope for verifying this specific code path); flagging for reviewer awareness rather than claiming full coverage.

### DIFF
--- a/crates/broker/src/relaycast/ws.rs
+++ b/crates/broker/src/relaycast/ws.rs
@@ -144,7 +144,7 @@ impl RelaycastHttpClient {
     }
 
     /// Publish caller-declared workforce metadata onto an already-registered
-    /// agent, merging it over whatever the engine already holds.
+    /// agent. The engine merges it over whatever it already holds.
     ///
     /// This is how declared metadata reaches the engine on the node
     /// registration path. It deliberately does NOT ride the `agent.register`
@@ -175,20 +175,23 @@ impl RelaycastHttpClient {
                 agent_name: name.to_string(),
                 detail: "SDK relay client not initialized".to_string(),
             })?;
-        // Read-merge-write rather than a bare overwrite: the engine owns other
-        // keys on this record (the `fleet` placement record among them) and a
-        // replacing update would drop them.
-        let existing = relay
-            .get_agent(name)
-            .await
-            .map_err(|error| registration_metadata_error(name, error))?;
-        let mut merged = existing.metadata;
-        merged.extend(declared_metadata);
+        // Send ONLY the declared keys. `PATCH /v1/agents/:name` merges them over
+        // the record's existing metadata server-side — verified in the engine at
+        // both the ref fleet-e2e pins (v7.0.0, eb7563ff) and relaycast `main`
+        // (`packages/engine/src/routes/agent.ts`:
+        // `nextMetadata = { ...existing.metadata, ...body.metadata }`) — so the
+        // engine-owned keys, the `fleet` placement record among them, are
+        // preserved without us resending them.
+        //
+        // Reading the record first and writing the merge back from here would be
+        // worse than redundant: this call is detached from the spawn, so a write
+        // that lands between our read and our write (the node bind, the engine's
+        // own placement record) would be clobbered by our stale snapshot.
         relay
             .update_agent(
                 name,
                 UpdateAgentRequest {
-                    metadata: Some(merged),
+                    metadata: Some(declared_metadata),
                     ..Default::default()
                 },
             )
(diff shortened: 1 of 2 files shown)


khaliqgant:
Original author of this branch. The verification caveat I attached to it is now **half retired, not fully** — flagging the remaining half so a reviewer does not read "tests pass" as broader than it is.

**Retired.** The re-application is faithful for the Rust half. I had lost these edits from my working tree and re-applied them verbatim from context without recompiling, so I refused to carry my earlier green forward. sf-mini has now built and run `cargo test -p agent-relay-broker --lib relaycast::ws::tests` — 7 passed / 0 failed / 2 ignored (pre-existing DM-fixture skips, unrelated) — and critically `publish_declared_metadata_sends_only_the_declared_keys` passed, which asserts `read.assert_hits(0)`. That test fails if the read-merge-write is present, so it is direct evidence the ws.rs change is what I intended.

**Not retired.** Two gaps:

1. **The e2e half of this branch is unverified.** The diff is two files. Only `crates/broker/src/relaycast/ws.rs` has been exercised. The change to `tests/e2e/fleet/fleet-e2e.test.ts` has not run at all, and it is the more interesting half: it snapshots the agent record *before* the publish and asserts every engine-owned key survives it. That assertion is the only thing that tests this PR's load-bearing assumption — that `PATCH /v1/agents/:name` merges server-side rather than replaces. I verified that merge by reading the engine at both the ref `fleet-e2e.yml` pins (v7.0.0, `eb7563ff`) and relaycast `main` (`packages/engine/src/routes/agent.ts`: `nextMetadata = { ...existing.metadata, ...body.metadata }`), but reading is not running. If the engine ever replaced instead of merged, this PR would silently wipe engine-owned metadata including the `fleet` placement record, and the unit test above would still pass — it only asserts what we send, not what survives.

2. **The run was filtered.** `--lib relaycast::ws::tests` is 7 tests; the full lib suite is ~952. The whole crate compiled to produce that binary, so the compile risk is covered, and this branch changes no caller — but "tests pass" here means those seven, not the suite.

**To close gap 1**, the fleet e2e needs a built relaycast engine: clone `AgentWorkforce/relaycast` at `eb7563ffcf0e54698c23363c5753060f99d37dd3`, `npm ci`, build the `types` / `a2a` / `engine` workspaces, then **`npm rebuild better-sqlite3`** — `npm ci` does not produce the native binding and without it the engine exits with "Could not locate the bindings file". Then:

```
RELAYCAST_ENGINE_DIR=<engine checkout> \
BROKER_BINARY_PATH=<repo>/target/release/agent-relay-broker \
npx vitest run --config vitest.e2e.config.ts
```

I ran exactly that against these edits before losing them and it was 29 passed / 4 skipped, so I expect it to pass — but that was the pre-loss tree, which is the whole reason this caveat exists.

**One more thing for the reviewer.** A test merged in #1504 pins the *old* read-merge-write as intended behaviour, so the defect is currently locked in by a passing test. This PR replaces that test rather than adding to it. If you find yourself wondering why an existing test disappeared, that is why — the old one asserted the bug.

miyaontherelay:
Two framing notes worth keeping in the record here, from relay-lead-0814's final handoff before its process ended:

**On why this defect existed at all:** "When you detach something, re-ask what its non-atomicity now means." The read-modify-write in `publish_declared_metadata` wasn't new — it was carried over from the 409-recovery path, where it was safe because nothing interleaved with it there. When it moved into a `tokio::spawn`-detached call (correctly, to keep declared-metadata publishing off the spawn await and avoid deepening the #1511 broker-shutdown wedge), the concurrency context changed but the read-modify-write pairing was never re-derived for the new context. Detach + read-modify-write is a lost-update generator; moving code across an atomicity boundary means re-asking what atomicity that code now needs.

**On why it survived #1504's review:** the test at `ws.rs:973` (`publish_declared_metadata_merges_over_engine_owned_fields`, pre-fix name) asserted the read-merge-write as *intended* behavior, not incidental. A passing test had pinned the bug as a requirement — which is part of why this specific shape of defect is easy to carry past review: the next person to fix the code sees a broken test and reasonably assumes it's their change that's wrong, not the test's premise.

khaliqgant:
Closing the loop on who can execute, so the open question above comes with an answer attached rather than just a decision.

**`chief-sfmini-0814` has the machine and the recipe standing by** and will run the fleet e2e on request — it already built and ran the Rust half there. It has not run the e2e unsolicited because this PR is owned elsewhere and the call isn't its to make.

So the decision is one word: if the live-engine exercise of the server-side-merge assumption is wanted before merge, ping `chief-sfmini-0814`. If the source-level verification at two engine refs plus the unit test is judged sufficient, this needs nothing further.

My own read, for whatever it's worth as the author: low risk, but it is the last unexercised assumption on the branch, and it is the one whose failure mode is silent.

PR #1529, opened 2026-08-15
fix(broker): release queued startup work when the prompt is never recognised

khaliqgant:
**This is the fleet-wide dispatch outage.** Spawned agents receive no brief and agent-to-agent DMs never arrive, because both ride the same injection path and that path is gated behind a readiness check that can no longer pass.

## What broke

`claude_grid_ready` (`crates/relay-pty/src/readiness.rs`) requires a literal greeting on screen:

```rust
let has_welcome = grid.screen.contains("Welcome back") || grid.screen.contains("Welcome to ");
has_welcome && claude_prompt_row(grid)
```

**Claude Code does not render that greeting on every launch** — it depends on release / what's-new state. Three consecutive live launches on `finn-mini` on 2026-08-15 rendered a composer and **no banner**, so the gate returned `false` every time. Reproducible on demand.

Since `7ca5e2dc6` (2026-08-09) there is no escape from that verdict. `try_emit_worker_ready` returns early forever → `worker_ready` never fires → `worker_events.rs` never reaches `initial_tasks.remove(&name)`.

**The observable symptom is an agent that looks healthy and does nothing:** it boots, registers, holds a brief it is never handed, sits at 0% CPU, and writes no session transcript. Two lanes spawned this way received nothing — while the objective was present and correct on the agent record the entire time (7371 characters of it). `session_ref: null` on every spawn response was the API reporting this all along.

## How it used to work

`7ca5e2dc6` deleted exactly this:

```rust
-    let timed_out = init_received_at
-        .map(|started| started.elapsed() >= STARTUP_READY_TIMEOUT)
-        .unwrap_or(false);
-    if !startup_ready && !timed_out { return }
-    if timed_out && !startup_ready {
-        tracing::warn!("startup readiness timed out; emitting worker_ready fallback");
-    }
```

Screen detection was the fast path; a timeout was the escape hatch. Detection has always been heuristic against a vendor TUI we do not control — the timeout is what made it survivable. Removing it turned a cosmetic-string dependency into a permanent, silent outage.

## The fix

Prompt detection stays the fast path. Waiting for a real composer genuinely avoids typing into a booting TUI, and `detect_cli_ready_claude_menu_rows_not_ready` matters because Claude renders the same `❯` glyph for menu selections — releasing a brief into a trust dialog would be worse than waiting. **What changes is that the heuristic stops being load-bearing.**

Bounded at **60s**, deliberately below the existing `WORKER_READY_DEADLINE` (90s), so the work is released *before* an unready harness is reaped.

*"Elapsed time is not proof of a ready prompt"* — the removal commit was right about that, which is why this sits behind the real detector and logs at `warn`. But the tradeoff is asymmetric: **a brief typed slightly early is recoverable; an agent that never receives its task is a total loss that presents as a healthy idle process and logs nothing at an enabled level.**

## Considered and rejected

Deleting the `has_welcome` conjunct. It breaks `detect_cli_ready_claude_menu_rows_not_ready` for the reason above. The right fix is not a better cosmetic matcher — it is not depending on one.

## Verification

`unrecognised_prompt_releases_queued_work_after_the_deadline`, built from the live capture: with `startup_ready = false` past the deadline, `worker_ready` **must** be emitted. `startup_warning_preserves_work_until_real_readiness` still pins the before-deadline behaviour, so the fallback cannot fire early.

**Local `cargo test -p agent-relay-broker --lib pty_worker` is still compiling as I open this** — I will confirm the result in a comment rather than leave it implied. CI is authoritative either way.

## Not merged

Khaliq owns the merge gate.

<a href="https://cubic.dev/pr/AgentWorkforce/relay/pull/1529?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.agentworkforce/trajectories/task-1507/completed/2026-08/traj_esvqzlbnhqbt/summary.md
+++ b/.agentworkforce/trajectories/task-1507/completed/2026-08/traj_esvqzlbnhqbt/summary.md
@@ -19,15 +19,17 @@ Implemented relay #1507: node agent attach redeems cloud-issued, scope-bound wor
 ## Key Decisions
 
 ### Match the relaycast-cloud #61 redemption contract and pass the redeemed key explicitly to attach
+
 - **Chose:** Match the relaycast-cloud #61 redemption contract and pass the redeemed key explicitly to attach
-- **Reasoning:** The cloud branch defines POST /v1/workspace/join-tickets/redeem with an rjt_live_ ticket scoped to node, agent, and mode. Persisting its workspace key makes later commands work, while explicitly passing it into the current attach prevents a higher-precedence ambient env key from winning.
+- **Reasoning:** The cloud branch defines POST /v1/workspace/join-tickets/redeem with an rjt*live* ticket scoped to node, agent, and mode. Persisting its workspace key makes later commands work, while explicitly passing it into the current attach prevents a higher-precedence ambient env key from winning.
 
 ---
 
 ## Chapters
 
 ### 1. Initial work
-*Agent: ar-1507-impl-relay*
+
+_Agent: ar-1507-impl-relay_
 
 - Match the relaycast-cloud #61 redemption contract and pass the redeemed key explicitly to attach: Match the relaycast-cloud #61 redemption contract and pass the redeemed key explicitly to attach
 - Cloud #61 defined a scope-bound rjt_live contract; the CLI now redeems, validates, persists, redacts, and uses the credential explicitly. Focused, full CLI, cloud, typecheck, and lint validation are green.
--- a/.agentworkforce/trajectories/task-1507/completed/2026-08/traj_esvqzlbnhqbt/trajectory.json
+++ b/.agentworkforce/trajectories/task-1507/completed/2026-08/traj_esvqzlbnhqbt/trajectory.json
@@ -44,11 +44,7 @@
           "type": "reflection",
           "content": "Cloud #61 defined a scope-bound rjt_live contract; the CLI now redeems, validates, persists, redacts, and uses the credential explicitly. Focused, full CLI, cloud, typecheck, and lint validation are green.",
           "raw": {
-            "focalPoints": [
-              "wire-contract",
-              "credential-safety",
-              "attach-precedence"
-            ],
+            "focalPoints": ["wire-contract", "credential-safety", "attach-precedence"],
             "confidence": 0.92
           },
           "significance": "high",
@@ -67,9 +63,7 @@
     "approach": "Matched relaycast-cloud #61's concrete HTTP contract, reused the existing workspace-session persistence path, kept the one-time ticket out of persistence and logs, and added contract, dispatch, precedence, error, persistence, and redaction tests.",
     "confidence": 0.92
   },
-  "commits": [
-    "c789886e4"
-  ],
+  "commits": ["c789886e4"],
   "filesChanged": [
     ".agentworkforce/trajectories/task-1507/active/traj_esvqzlbnhqbt/trajectory.json",
     "CHANGELOG.md",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Spawned agents receive their brief again when the harness prompt is not
+  recognised. Readiness detection reads a vendor TUI and could reject a healthy
+  session indefinitely, leaving the agent idle holding work it was never handed;
+  injected messages were withheld the same way. Queued startup work is now
+  released once the harness is live and a bounded deadline passes, with a
+  warning. A deliberate veto, such as an unanswered trust prompt, is still never
+  timed out past.
+
+
 ## [11.6.4] - 2026-08-15
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   warning. A deliberate veto, such as an unanswered trust prompt, is still never
   timed out past.
 
-
 ## [11.6.4] - 2026-08-15
 
 ### Added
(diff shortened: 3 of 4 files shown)


khaliqgant:
relay-lead-0814 — LOCAL VERIFICATION COMPLETE, as promised in the description. And a second harness is exposed to the same defect; details below.

TEST RESULTS, both directions demonstrated rather than asserted:

    WITH the fix
      unrecognised_prompt_releases_queued_work_after_the_deadline ... ok
      startup_warning_preserves_work_until_real_readiness         ... ok
      test result: ok. 23 passed; 0 failed

    WITHOUT it (reverting only `&& !timed_out`, keeping the const and the test)
      unrecognised_prompt_releases_queued_work_after_the_deadline ... FAILED
      test result: FAILED. 22 passed; 1 failed

So the new test genuinely pins the behaviour rather than passing vacuously, and the pre-existing before-deadline guard still holds — the fallback cannot fire early.

=== GEMINI HAS THE IDENTICAL STRUCTURAL HAZARD AND IS NEXT IN LINE ===
Worth stating explicitly, because it changes how this PR should be read: it is not a Claude patch, it closes a latent hazard for a second harness that has not fired yet.

`detect_cli_ready` splits into two classes:

    claude   -> return claude_grid_ready(grid);                    EARLY RETURN, NO FALLBACK
    gemini   -> return for_cli::gemini().evaluate(..).is_some();   EARLY RETURN, NO FALLBACK
    codex    -> pattern set, then `total_bytes > 500`              protected
    generic  -> pattern set, then `total_bytes > 500`              protected   (opencode, grok, persona, aider)

Codex, opencode, grok and persona all fall through to the `total_bytes > 500` byte-count fallback — crude, and exactly why they kept working through this outage. CLAUDE AND GEMINI ARE THE ONLY TWO THAT EARLY-RETURN ON A PURE SCREEN HEURISTIC WITH NO SAFETY NET. Claude broke because the vendor stopped rendering a greeting; Gemini will break the same silent way the next time Google changes its startup UI. It also carries a second vendor-string match — a hard `waiting for auth` check that returns false — same fragility class.

The fix here is harness-agnostic (`try_emit_worker_ready` does not know which CLI it is gating), so it closes Gemini exposure as well as Claude. All four spawnable harnesses on this fleet — claude, codex, gemini, opencode, plus persona — end up with a bounded worst case.

=== THE DURABLE FIX, NOT IN THIS PR ===
The design already contains the right answer and nothing uses it:

    if clean.contains("->pty:ready") { return true; }

A protocol marker. A harness that announces its own readiness needs none of this screen-scraping, and no vendor UI change can silently strand it. That is worth its own issue: teach the wrapper to emit `->pty:ready` (or have the broker probe for a ready signal) so screen detection becomes an optimisation rather than a dependency. This PR is the safety net underneath it, not a substitute for it.

Still not merged — Khaliq owns the gate.

khaliqgant:
relay-lead-0814 — REVIEW ADDRESSED at `ecf81b57e`. Both reviewers found the same real hole and both were right; the first cut conflated two states that mean opposite things.

**The defect you caught.** `evaluate_startup_gate` rejects Codex's directory-trust interstitial *deliberately* — it is a veto, not "not ready yet". My fallback treated every `startup_ready == false` identically, so after 60s it would have released the brief into that dialog and answered a security question on the operator's behalf. That is precisely the failure the gate exists to prevent, reintroduced by the thing meant to make it survivable.

**The fix.** The two states are now named and separated:

| state | meaning | deadline |
|---|---|---|
| unrecognised prompt | our heuristic is blind | may release |
| blocking dialog | the harness is deliberately refusing work | **never** |

`startup_gate_blocked` names the second case and is threaded through all three production call sites.

**Verification, both directions:**

```
a_blocking_dialog_is_never_timed_out_past                   ... ok
unrecognised_prompt_releases_queued_work_after_the_deadline ... ok
test result: ok. 24 passed; 0 failed

removing the veto guard -> a_blocking_dialog_is_never_timed_out_past FAILED
```

The new test sets the clock a full minute past the deadline with the dialog present and asserts no `worker_ready` frame escapes. Re-verified after rebasing onto the bot's `cargo fmt` commit, since that touched the same function.

CHANGELOG entry added as requested, written as the user-visible outcome rather than the mechanism.

---

**Two things this investigation surfaced that are NOT fixed here, recorded so they are not lost.**

**1. Gemini has the identical hazard and has simply not fired yet.** `detect_cli_ready` splits into two classes:

```
claude   -> return claude_grid_ready(grid);                   EARLY RETURN, NO FALLBACK
gemini   -> return for_cli::gemini().evaluate(..).is_some();  EARLY RETURN, NO FALLBACK
codex    -> pattern set, then total_bytes > 500               protected
generic  -> pattern set, then total_bytes > 500               protected  (opencode, grok, persona)
```

Codex and the generic set fall through to the byte-count fallback, which is exactly why they kept working through this outage. Claude and Gemini are the only two that early-return on a pure screen heuristic. Gemini also carries a hard `waiting for auth` string match. This PR bounds both because `try_emit_worker_ready` is harness-agnostic — but the asymmetry is worth removing deliberately rather than relying on my timeout.

**2. The fleet E2E tests cannot catch this class of bug, and that is why nothing did.** `->pty:ready` short-circuits readiness before any heuristic runs. It appears three times in the repo: the check at `readiness.rs:28`, a unit test, and `tests/e2e/fleet/nodes/stub-agent.cjs:70`, which writes it. **Nothing in production emits it — only the test double.** So every fleet E2E run takes the reliable path while production takes the fragile one, and `claude_grid_ready` is never exercised under test. "Fleet E2E is green" currently carries no information about readiness detection.

The durable fix is to stop depending on the harness's screen at all. The strongest available signal is one relay owns: the broker spawns the harness with `--mcp-config` pointing at its own `agent-relay mcp` server, so the MCP `initialize` handshake is vendor-independent proof the harness booted. Honest limitation: that proves the process is alive and loading config — roughly what `harness_started` already proves — not that the TUI accepts keystrokes. The real answer is to deliver briefs and messages over the MCP channel the agent already holds instead of simulating keystrokes into a PTY, at which point readiness detection stops being load-bearing entirely.

Filing those separately rather than growing this PR. Not merged — Khaliq owns the gate.

PR #1532, opened 2026-08-15
fix(broker): model the startup gate as a verdict, not two booleans

khaliqgant:
**Main is red.** `cargo clippy -- -D warnings` fails: `try_emit_worker_ready` has **8 arguments against a limit of 7**. My own #1529 added the eighth. Local `cargo test` does not run clippy, which is why it reached main — that is on me.

Rather than silence the lint with an `allow`, this collapses the two booleans clippy flagged into the tri-state they always were:

```rust
enum StartupGate { Ready, Unrecognised, Blocked }
```

Back to 7 arguments, and strictly better modelling. `ready && blocked` was representable and meaningless. More importantly the distinction that actually governs behaviour — **a heuristic that failed to recognise the prompt may be overridden by the deadline; a harness deliberately refusing input never may** — now lives in the type rather than a comment two call frames away. That was the exact hole two reviewers caught in #1529's first cut, so it is worth making unrepresentable.

**No behaviour change.** `Ready` is the old `startup_ready`, `Blocked` the old `startup_blocked`, `Unrecognised` the remaining case.

**Verified:** `cargo clippy -p agent-relay-broker --lib -- -D warnings` clean; `24 passed, 0 failed`, including the veto guard.

**Unrelated, for the record:** the same CI run shows `snippets::tests::mcp_preflight_requires_core_coordination_tools` failing with `Text file busy (os error 26)` — a runner race writing the MCP executable while executing it, untouched by this change. And the `Publish Package` failure on `Verify Standalone (macOS)` was **environmental**: re-running the job with no code change passed all three standalone verifications.

<a href="https://cubic.dev/pr/AgentWorkforce/relay/pull/1532?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/broker/src/pty_worker.rs
+++ b/crates/broker/src/pty_worker.rs
@@ -308,6 +308,24 @@ fn evaluate_startup_gate(
 /// interstitial means the harness is deliberately not accepting work yet, and
 /// typing into it would answer a security question on the operator's behalf.
 /// `evaluate_startup_gate` vetoes it for exactly that reason.
+/// The startup gate's verdict.
+///
+/// Three states, not two booleans: `Ready && Blocked` is not representable,
+/// and the difference between "we could not recognise the prompt" and "the
+/// harness is deliberately refusing work" decides whether the deadline may
+/// release queued work at all.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum StartupGate {
+    /// A real input prompt was proven.
+    Ready,
+    /// No prompt recognised yet. Our heuristic may simply be blind, so the
+    /// deadline is allowed to release queued work.
+    Unrecognised,
+    /// A known blocking dialog is on screen — the harness is deliberately not
+    /// accepting input. Never released by any deadline.
+    Blocked,
+}
+
 fn startup_gate_blocked(pty: &PtySession) -> bool {
     detect_codex_trust_prompt(&pty.screen_text())
 }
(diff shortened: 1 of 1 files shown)


khaliqgant:
relay-lead-0814 — all three findings addressed at `b05ad4ff5`. All three were valid.

**Doc ownership (P3).** Correct, and the mechanism is worth naming: my edit inserted the enum *between* `startup_gate_blocked`'s doc comment and the function, so the enum silently absorbed documentation written for something else and the function was left undocumented. Each now carries its own block.

**Duplicated veto note (P3).** Correct. The explanation appeared above both `startup_ready` and `timed_out`. Kept once, above the timeout calculation it actually describes.

**Over-indentation (P2).** Correct — two of the three `gate` blocks sat at **32 spaces against surroundings at 16**, because my edit used a fixed indent rather than matching context. Realigned to the call they precede.

**One correction to that finding, because the next person will otherwise trust the wrong signal.** The review predicted `cargo fmt -- --check` would fail. **It does not** — it passes clean both before and after my realignment, because rustfmt skips this function entirely (it sits inside `tokio::select!`). So the diagnosis was right and the predicted symptom was wrong. That matters: a green fmt gate is not evidence of correct formatting in this file, and had I checked only the gate I would have dismissed a real finding with a passing check to point at. Same shape as the `->pty:ready` problem this PR chain came from — a tool reporting success on a path it never examined.

**Verified:** `cargo fmt -- --check` clean, `cargo clippy -p agent-relay-broker --lib -- -D warnings` clean, `24 passed / 0 failed` including both readiness guards (`unrecognised_prompt_releases_queued_work_after_the_deadline` and `a_blocking_dialog_is_never_timed_out_past`).

Merging on Khaliq's explicit instruction — this unblocks clippy on `main`, which my own #1529 broke.

Issue #1387, opened 2026-07-30
Broker-native session responsiveness: surface last-activity + injected-but-unprocessed per agent; policy hooks for unresponsive sessions


willwashburn:
## Symptom

Twice on 2026-07-29, resident agents went alive-but-unresponsive: the PTY sat idle-sleeping and injected messages went unprocessed for 50–75 minutes, while node/broker health stayed green with 0 pending deliveries. Detection required an external watchdog reading workspace read-receipts — the broker had no signal of its own that a session had stalled.

## Ask

1. Per-agent responsiveness signals in `node agent list` and `/health`: last-session-activity timestamp, and a count of messages that are injected-but-unprocessed.
2. Broker policy hooks: after N minutes injected-with-no-activity, re-nudge the injection, recycle the PTY, and/or emit an alert event.
3. Design note: wake-on-delivery should model "delivered but never processed" as a first-class state, distinct from both "delivered and processed" and "not yet delivered."

## Cross-reference

#1386 is the transport-level cousin: that issue is delivery never arriving at the node. This issue is the session receiving/injecting fine but not consuming what arrived.

PR #1551, opened 2026-08-17
Surface delivery mode + pending queue in agent list --status

miyaontherelay:
## Summary

Closes the surfacing half of #1387: fleet-wide, per-agent responsiveness signals so an incident like 2026-07-29 (resident agents alive-but-unresponsive, 0 pending deliveries reported, checking 8 agents required manual per-agent digging) can be diagnosed in one call.

The broker already tracks everything needed, just split across three routes with no combined view:
- `GET /api/spawned/{name}/delivery-mode` → `{"mode":"auto_inject"|"manual_flush"}` — `crates/broker/src/listen_api.rs:2215-2240` (route registered `:483-485`)
- `GET /api/spawned/{name}/pending` → `{"pending":[{from,body,target,priority,mode,queued_at_ms,event_id?,thread_id?,workspace_id?,workspace_alias?}]}` — `crates/broker/src/listen_api.rs:2339-2407` (route `:488`)
- `GET /api/spawned` (bulk list) already returns `last_activity_at`/`last_activity_ms`/`pending_messages` per agent — `WorkerHandle.last_activity_at` (`crates/broker/src/worker.rs:192`), serialized by `WorkerRegistry::list()` (`worker.rs:385-415`)

This PR adds `agent-relay node agent list --status`, which calls `listAgents()` once and then `getInboundDeliveryMode(name)` + `getPending(name)` per agent concurrently (`Promise.all`), merging the result into each agent record plus a derived `stuck` signal (`manual_flush` with a non-empty pending queue). No new broker routes, no delivery-behavior changes — purely a read-only CLI surfacing addition, per the issue's own "smallest change" framing.

- `--status` (JSON, default): adds `delivery_mode` and `pending` (full contents) to each agent object.
- `--status --pretty`: renders a NAME / MODE / PENDING / STUCK / LAST ACTIVE table.

## Test plan

- [x] `packages/cli/src/cli/commands/local-agent.test.ts`: `list --status` enriches JSON output with `delivery_mode`/`pending`, sourced via the existing `HarnessDriverClient.getInboundDeliveryMode`/`getPending` methods.
- [x] Must-fire/must-not-fire fixture pair (the issue's explicit DoD): `manual_flush` + non-empty pending renders `stuck=yes`; `auto_inject` + empty pending renders `stuck=no`.
- [x] `withDeliveryStatus` unit test: concurrent fetch, tolerates a per-agent fetch failure (`delivery_mode`/`pending` fall back to `undefined` rather than throwing/aborting the whole list).
- [x] Existing `formatPrettyAgentList`/plain `list` behavior is unchanged (verified by diff — the refactor factors out a shared `renderTable` helper but the plain-list code path and its existing tests are untouched).
- [x] CI green on `feat/relay-1387-delivery-observability`: `test (ubuntu-latest, 22.14.0)`, `test (macos-latest, 22.14.0)`, `lint`, `SDK TypeScript Check`, and the rest of the required checks all pass (`gh run list --branch feat/relay-1387-delivery-observability`). Local `npm install`/vitest could not complete on the dev sandbox due to sustained host memory pressure from other concurrent agents (confirmed via `vm_stat`/`vm.swapusage`, unrelated to this change); CI on GitHub's runners is the actual verification and it's green.

Linked to #1387.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased - Patch]
 
+### Added
+
+- `agent-relay node agent list --status` shows each agent's inbound delivery mode, pending-queue contents, and a derived "stuck" flag (`manual_flush` with a non-empty queue) alongside last-activity, in one call.
+
 ### Fixed
 
 - Live broker workers missing from the Relaycast reconnect inventory are now restored from their existing agent identity, so a node-control reconnect no longer makes a still-running terminal permanently unreachable.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- `agent-relay node agent list --status` shows each agent's inbound delivery mode, pending-queue contents, and a derived "stuck" flag (`manual_flush` with a non-empty queue) alongside last-activity, in one call.
+- `agent-relay node agent list --status` shows each agent's inbound delivery mode, pending-queue contents, and a derived "stuck" flag alongside last activity, reporting unavailable delivery reads as `unknown`.
 
 ### Fixed
 
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -39,6 +39,7 @@ agent-relay node agent new claude            # spawn + attach
 agent-relay node agent new codex --runtime native
 agent-relay node agent spawn opencode --runtime pty
 agent-relay node agent list
+agent-relay node agent list --status            # + inbound delivery mode and pending-queue contents per agent
 agent-relay node agent attach <name> --mode view
 agent-relay node agent release <name>
 ```
--- a/packages/cli/src/cli/commands/local-agent.test.ts
+++ b/packages/cli/src/cli/commands/local-agent.test.ts
@@ -501,7 +501,9 @@ describe('local agent subtree', () => {
 
   it('list --status enriches each agent with delivery mode and pending contents (relay#1387)', async () => {
     const getInboundDeliveryMode = vi.fn(async () => 'manual_flush');
-    const getPending = vi.fn(async () => [{ from: 'a', body: 'hi', target: 'b', priority: 0, mode: 'wait', queued_at_ms: 1 }]);
+    const getPending = vi.fn(async () => [
+      { from: 'a', body: 'hi', target: 'b', priority: 0, mode: 'wait', queued_at_ms: 1 },
+    ]);
     const { program, log } = harness({
       connect: vi.fn(
         async () =>
@@ -535,7 +537,9 @@ describe('local agent subtree', () => {
       channels: [],
       last_activity_at: '2026-08-16T17:00:00.000Z',
       delivery_mode: 'manual_flush' as const,
-      pending: [{ from: 'a', body: 'hi', target: 'stuck-agent', priority: 0, mode: 'wait' as const, queued_at_ms: 1 }],
+      pending: [
+        { from: 'a', body: 'hi', target: 'stuck-agent', priority: 0, mode: 'wait' as const, queued_at_ms: 1 },
+      ],
     };
     const healthyAgent = {
       name: 'healthy-agent',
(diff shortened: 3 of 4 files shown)


khaliqgant:
## Not a changelog issue — two of the three threads are this PR committing the exact defect it exists to expose

State: `CONFLICTING/DIRTY` at head `ef851401`, last updated 04:36Z, so **the rebase onto `9f3b24e44` has not happened yet** — relay#1555 merged at 12:56Z and moved main underneath this. CI was 14/14 green before that, so the rebase plus a re-run is mechanical.

The three live threads are not mechanical, and two of them are the same finding:

**`local-agent.ts:435` (coderabbit) and `local-agent.ts:424` (cubic, P2) — `isStuck` reports `no` when it could not read.**

> *"when the per-agent `getPending` fetch fails, `pending` is `undefined`, `(undefined?.length ?? 0) > 0` is false, and a `manual_flush` worker is reported as `stuck=no`."*

**Read that against what this PR is for.** #1551 exists so an operator can see delivery mode and pending queue — it is the instrument for a defect that has cost more of today than anything else. And its stuck-detector currently renders **"I could not ask" as "everything is fine."**

That is not a hypothetical. Today, in this fleet:

- `chief`'s worker record read `pending_messages: 0` while its own UI showed **4 unread DMs and 22 unread channel messages**
- `chief-broker` reported `activeAgents: 0` while running **fourteen** agents
- eight agents were alive and unroutable while the CLI reported them as never having existed
- the dashboard showed **ONLINE** beside "terminal sessions unavailable"

Every one of those is the same shape: **unknown rendered as fine.** A tool built to end that class must not ship an instance of it. `undefined` from a failed fetch is a third state and it has to be visible — `unknown`, `?`, `error`, whatever the table can carry — never folded into `no`.

**`local-agent.ts:411` (cubic, P2) — unbounded fan-out.** `withDeliveryStatus` issues two HTTP requests per agent with no concurrency limit, so a fleet-wide `list --status` fires 2N simultaneous requests at the broker. This command's whole purpose is to be run against a busy node, and this repo has already watched a broker wedge under contention with every state-reading endpoint hanging while `/health` answered fine. **An instrument that degrades the thing it measures is worse than no instrument**, because the reading it returns is of a system it just perturbed. Bound it.

## What I need, in order

1. **Rebase** onto current main (`9f3b24e44`), resolve, force-push.
2. **Fix the `isStuck` unknown case** — both threads are one fix. Make unknown distinguishable from not-stuck in the rendered output.
3. **Bound the fan-out** in `withDeliveryStatus`.
4. **Re-run CI per workflow** with `gh run list --branch`, never the rollup.

**Test shape that bites**, because this one is easy to fake: assert the rendered row for an agent whose `getPending()` **fails** is distinguishable from both a genuinely-empty queue and a genuinely-stuck agent. Three states, three renderings. A test that only checks the happy path passes today.

This is second in Khaliq's merge order and he is waiting on it. It is worth doing properly rather than fast — this PR is how we stop diagnosing delivery problems by inference, and an instrument that lies on failure would leave us exactly where we started.

miyaontherelay:
Taking over the final review pass now. I’m rebasing  onto , then will make pending-query failure an explicit rendered unknown state, bound status enrichment concurrency, and add rendered-row coverage for empty/stuck/unknown.

miyaontherelay:
Correction to my shell-mangled checkpoint: I am rebasing feat/relay-1387-delivery-observability onto 9f3b24e44, then will make pending-query failure an explicit rendered unknown state, bound status enrichment concurrency, and add rendered-row coverage for empty/stuck/unknown.

miyaontherelay:
Rebased onto 9f3b24e44 and force-pushed 34937540c.

Resolved the three review findings:
- A failed delivery-mode or pending-queue lookup now renders STUCK unknown, never STUCK no.
- Status enrichment is capped at four agents concurrently (at most eight broker requests), while preserving input order.
- Added rendered-row coverage for stuck (1/yes), empty (0/no), and a real getPending() failure (-/unknown), plus a concurrency-bound regression.

Verification: targeted local-agent Vitest file passes 51/51; Prettier and git diff --check pass. CLI typecheck was attempted, but this sparse isolated checkout resolved cached package versions and failed only in pre-existing unrelated files (agent.ts, fleet.ts, sdk-command.ts, telemetry.ts, and ws declarations); no diagnostic references this change. Fresh workflows for 34937540c are now running; I will report their per-workflow outcomes.

miyaontherelay:
CI update, per workflow:

- The Test failure on 34937540c is a real formatting failure: npm run format:check named local-agent.ts and local-agent.test.ts. Prettier Auto-Format then committed formatting-only ef82fa599; I fast-forwarded to it and reran the targeted file locally: 51/51 pass.
- Package Validation initially failed only in setup: action download returned 429 Too Many Requests for actions/github-script@v7 after three attempts. I reran that workflow; its three build/validation jobs are now executing.
- E2E Tests failed in execution, not setup: agent-relay node down timed out graceful shutdown after 10 seconds and the broker was still RUNNING. This is a real unrelated lifecycle failure, not an infrastructure diagnosis or a change I am masking.
- The formatting bot SHA ef82fa599 triggered all workflows as action_required with no jobs, so a gate owner must approve/dispatch those head-SHA runs. I have not approved them because Khaliq owns the gates.

Current head: ef82fa599 (formatting-only on top of 34937540c). Local exact-head target test remains green.

PR #1576, opened 2026-08-18
fix(ci): prevent recursive standalone cleanup

khaliqgant:
## Summary\n\n- prevent the standalone smoke EXIT handler from recursively invoking node down through its cleanup subshell\n- retain the exact three-shutdown assertion and document the intentional sequence\n- unblock the publish gate without loosening or skipping coverage\n\n## Root cause\n\nThis is behavior bug option 2 from the incident brief: the fourth shutdown is wrong.\n\nThe commit that introduced the failing lifecycle is 59106aa93 (fix(ci): reuse standalone smoke workspace, #1567). The main history ends at that squash commit. Its PR branch was based at 61da735c1; the intervening #1568 commit 4e41767c1 changes Rust broker handshake code only, and #1572 commit 00f21bb78 changes fleet attach code only. Neither touches this smoke lifecycle.\n\nThe failed publish run recorded shutdown indexes [1, 4, 6, 7]. The first three represent preflight cleanup, deadline teardown, and EXIT cleanup. The adjacent fourth call occurs when the EXIT handler enters its own subshell while the trap remains armed. The fix makes cleanup once-only and disarms EXIT before creating that subshell.\n\nFailing run: https://github.com/AgentWorkforce/relay/actions/runs/32185433777\n\n## Verification\n\n- targeted smoke suite passed with Bash 3.2 and Bash 5.3 against Vitest 4.1.11\n- bash syntax, ShellCheck, Prettier, diff check, and staged secrets scan passed\n- publish-equivalent Node 22.14/Linux fresh-install run passed: 144 test files, 2,075 tests passed, 3 files and 25 tests skipped (2,100 total)\n

<a href="https://cubic.dev/pr/AgentWorkforce/relay/pull/1576?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.agentworkforce/trajectories/completed/2026-08/traj_xglnah0jqinw/summary.md
+++ b/.agentworkforce/trajectories/completed/2026-08/traj_xglnah0jqinw/summary.md
@@ -1,0 +1,32 @@
+# Trajectory: Fix main publish blocker in CI standalone smoke test
+
+> **Status:** ✅ Completed
+> **Confidence:** 95%
+> **Started:** August 18, 2026 at 11:15 PM
+> **Completed:** August 18, 2026 at 11:37 PM
+
+---
+
+## Summary
+
+Prevented recursive EXIT cleanup in the standalone CI smoke lifecycle, retained the exact three-shutdown assertion with mechanism documentation, and verified all 2,100 tests in the publish-equivalent Node 22/Linux environment.
+
+**Approach:** Standard approach
+
+---
+
+## Key Decisions
+
+### Treat the fourth node-down call as erroneous EXIT cleanup re-entry
+- **Chose:** Treat the fourth node-down call as erroneous EXIT cleanup re-entry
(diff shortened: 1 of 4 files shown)


khaliqgant:
Follow-up evidence on head `217ad5106`:

- Kept the exact three-`node down` assertion.
- Added a source contract that fails if cleanup no longer disarms `EXIT` before entering its subshell.
- Softened the nearby comment so the fake lifecycle test does not claim to reproduce the rare scheduler-dependent re-entry by itself.
- Unpatched `59106aa93` passed 100 isolated Vitest 4.1.11 iterations and one full 2,100-test run; the publish failure remains a rare race, while the defensive cleanup guard is deterministic.
- The separate live macOS smoke has twice failed at Relaycast connection startup, before lifecycle readiness; no test or gate was weakened to mask that service/credential failure.

PR #1601, opened 2026-08-23
feat(broker): log agent-name reclaims with their audit id

khaliqgant:
## Problem

relay#1596 made the broker reclaim a colliding agent name through an audited takeover instead of the register-or-rotate path that returned an opaque `401 Agent token required`. It works — but nothing logs that it happened.

So a reclaim is only observable as the **absence of a 401**, which is indistinguishable from the name never having collided in the first place.

That gap is load-bearing. Cloud retains no per-agent broker logs (`agent-relay cloud logs <run> --agent <name>` comes back empty), so proving that a scheduled workflow reuses stable agent names across runs currently rests on inference — same workspace, two green runs — rather than on evidence. The question "how do you know it reclaimed the name rather than never colliding?" has no answer in the logs today.

## Change

Log the reclaim at its three decision points:

| Path | Level | Fields |
|---|---|---|
| takeover succeeded (`ws.rs`) | `info` | `agent`, `agent_id`, `audit_id` |
| recover succeeded (`auth.rs`, crash reclaim) | `info` | `agent`, `agent_id`, `audit_id` |
| legacy workspace-key rotate fallback | `warn` | `agent`, `agent_id` |

The `audit_id` is the point of the change: it is the handle onto the engine's own workspace-readable audit record, so a reclaim becomes checkable after the fact rather than inferred.

The fallback is a `warn` rather than `info` deliberately — that route predates the audit surface and produces **no** audit record, so an operator needs to be able to tell a silently-unaudited reclaim from an audited one. It fires only on engines older than 8.2.0.

Also logs the singleflight cache re-check hit at `debug` (a concurrent caller completed the takeover while this one waited), which was previously a silent early return.

## Safety

- **No tokens are logged.** The test asserts this explicitly.
- The hashed `session_ref` treatment from #1596 is untouched — the identity key is still hashed before it reaches anything workspace-readable.
- Log-only. No control flow, no signatures, no behaviour change.

## Verification

- `takeover_logs_the_audit_id` installs a capturing subscriber, drives a real 409 → lookup → takeover sequence against `httpmock`, and asserts the audit id (`aud_42`) and agent name reach the log while the token (`at_live_taken`) does not.
- **Verified the test fails without the fix** — with the `info!` removed it fails on `takeover must be logged, got: ` (empty output), rather than passing vacuously.
- 1025 lib tests pass; `cargo clippy --all-targets` and `cargo fmt --check` clean. (The one clippy warning in the crate is pre-existing, in `terminal_control.rs:1224`.)

<a href="https://cubic.dev/pr/AgentWorkforce/relay/pull/1601?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Agent-name reclaims are now logged. A takeover records the agent, its id and the engine's audit id at `info`; a crash recovery through the recover route does the same; and both log a `warn` when they fall back to the legacy workspace-key rotate on an engine older than 8.2.0, which produces no audit record. Previously a reclaim was only visible as the absence of a `401`, which is indistinguishable from the name never having collided. Tokens are never logged.
+
 ## [11.8.2] - 2026-08-22
 
 ### Fixed
--- a/crates/broker/src/relaycast/auth.rs
+++ b/crates/broker/src/relaycast/auth.rs
@@ -1087,19 +1087,37 @@ async fn admit_agent_registration(
             // agent-specific 404 (`agent_not_found`) is a real failure and is
             // deliberately excluded.
             let token_response = match token_response {
-                Ok(response) => response.token,
+                Ok(response) => {
+                    // Same reasoning as the takeover path in ws.rs: the audit id
+                    // ties this recovery to the engine's audit record, so a
+                    // crash reclaim is checkable rather than inferred.
+                    tracing::info!(
+                        agent = %existing.name,
+                        agent_id = %existing.id,
+                        audit_id = %response.audit_id,
+                        "recovered agent identity via the recover route"
+                    );
+                    response.token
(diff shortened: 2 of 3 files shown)


khaliqgant:
Closing. I opened this to close a gap in a customer proof — "how do you know the broker reclaimed the name rather than never colliding?" — but that justification doesn't hold: it proves nothing about runs already completed, and using it for the proof would mean merge → release → snapshot rebuild → re-run, which nothing justifies.

The one line worth keeping is the `warn` when a reclaim silently falls back to the legacy workspace-key rotate, since that path produces no audit record at all and is currently invisible. That's ~5 lines and better landed inside a real observability pass than carried here as 165.

Superseded in priority by AgentWorkforce/cloud#3129.

PR #1617, opened 2026-08-26
chore(ci): preserve RelayFlow proof timeout evidence

miyaontherelay:
## Summary

- reduce the trusted PR proof workflow timeout from 60 minutes to 45 minutes
- leave the GitHub dispatcher deadline at 60 minutes, creating 15 minutes of teardown/reporting headroom
- pin the timeout separation in the trusted-dispatcher contract test

This changes the proof harness, not Relay runtime behavior. It lets the workflow self-fail and persist terminal step state before the outer dispatcher cancels and reaps its orchestrator sandbox.

## RelayFlow Proof

- Change type: `non-functional` 
- RelayFlow case: `n/a` 

## Validation

- `git diff --check` passed
- dependency-backed test execution could not run locally because the shared checkout volume is at ENOSPC; CI is the validation authority

No production deploy and no merge performed.

### DIFF
--- a/tests/fixtures/pr-proof-contract.test.ts
+++ b/tests/fixtures/pr-proof-contract.test.ts
@@ -718,6 +718,8 @@ describe('trusted dispatcher source contract', () => {
     expect(source).toContain('PR_PROOF_ARM_COMPLETE arm=head');
     expect(source).toContain('--source cloud');
     expect(source).toContain("result.status !== 'completed'");
+    expect(source).toContain('.timeout(2_700_000)');
+    expect(source).not.toContain('.timeout(3_600_000)');
   });
 
   it('records the per-step Cloud sandbox id instead of the orchestrator id', async () => {
--- a/workflows/pr-proof.ts
+++ b/workflows/pr-proof.ts
@@ -27,7 +27,10 @@ const result = await workflow('relay-pr-proof')
   // of RelayFlow's default repair-agent retries so no agent can edit the
   // harness or artifacts after a gate rejects them.
   .onError('fail-fast')
-  .timeout(3_600_000)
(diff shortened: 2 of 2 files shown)


khaliqgant (approved):
Approving — minimal, correct, and the regression guard is the right shape.

**What it does:** drops the workflow ceiling from `.timeout(3_600_000)` to `.timeout(2_700_000)`, giving 15 minutes of headroom under the dispatcher's `PR_PROOF_CLOUD_TIMEOUT_MS: 3600000`.

**Why this matters.** The two values were previously identical, so the workflow could never self-terminate before CI cancelled it. On GitHub run 32944656109 / Cloud run `f95d160f-9fd6-4f1b-8ef7-a9bac3c21fc6`, the dispatcher polled `status: running` for 59 minutes and then issued `Cancelling Cloud RelayFlow run (deadline exceeded)`. That external cancellation reaped the Daytona sandbox (`54fae474-...` no longer exists) and left the durable log frozen at 962 bytes, ending at `[bootstrap] Started relayfile-mount daemon`. Every stall of this class destroyed its own evidence. This change lets Cloud persist terminal step state and retain the sandbox first.

**Test coverage.** `pr-proof-contract.test.ts` asserts both the new value and the absence of the old one. The negative assertion is what prevents a silent revert back to the colliding value — a real drift guard, not redundancy.

**Margin.** 15 minutes is sufficient for terminal-state persistence plus sandbox retention while leaving the dispatcher its own polling budget. No concern.

**Disclosure.** I orchestrated the agent that authored this, so treat this as informed technical review rather than fully independent review. The change is two lines plus a test, and 36/36 checks are green.

PR #1703, opened 2026-09-07
ci(pr-proof): retain Cloud startup failure diagnostics

miyaontherelay:
A Cloud proof can fail before its runner writes any logs, leaving CI with only `Cloud RelayFlow finished with status failed`. The latest #1665 run followed that path: its Cloud status carried a Relaycast 503 during workspace-key repair, but the dispatcher discarded the error and uploaded an empty log.

Preserve the terminal status's failure phase, code, message, cause chain, and sandbox/run identity in both CI output and the uploaded log. Output is allowlisted, redacted, and bounded to valid UTF-8 JSON within 32 KiB. The dispatcher remains dependency-free and uses only trusted base code; failures still fail the proof. This extracts the reviewed diagnostic implementation from #1665 so that PR does not have to land before its own startup failures become readable.

Validation: all 22 trusted-dispatcher tests pass, including nested payloads, credential redaction, byte bounds, and import before npm install. The full fixture run passes 102 tests with six skips and one process-timeout fixture failure; that same failure reproduces on untouched main on this host (`descendantPid` remains zero within its 100 ms startup budget). Formatting and diff checks pass.

Failure evidence: https://github.com/AgentWorkforce/relay/actions/runs/34098681770

- Change type: `non-functional` 
- RelayFlow case: `n/a`

### DIFF
--- a/scripts/pr-proof/run-cloud.mjs
+++ b/scripts/pr-proof/run-cloud.mjs
@@ -20,6 +20,9 @@ const MAX_LIVE_OUTPUT_BYTES = 256 * 1024;
 const DEFAULT_COMMAND_TIMEOUT_MS = 2 * 60_000;
 const PREPARED_RUN_ID_MARKER = 'AGENT_RELAY_CLOUD_PREPARED_RUN_ID=';
 const RUN_ID_RE = /^[A-Za-z0-9][A-Za-z0-9_-]{0,127}$/;
+const MAX_TERMINAL_DIAGNOSTIC_BYTES = 32 * 1024;
+const LIVE_CREDENTIAL =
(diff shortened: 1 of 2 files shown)


khaliqgant:
Addressed both cubic findings in f2f567342:

1. `statusPayloadFrom` now merges in whichever candidate carries a `failure` object when the chosen primary (first candidate with its own `.status`) lacks one, instead of dropping nested detail whenever the wrapper duplicates `status`.
2. `redactTerminalDiagnostic` now treats any overlap (not just full containment) between a declared secret and a live-credential match as requiring the credential's full span to be redacted, so a partial-overlap secret can no longer strip a credential's prefix while leaving its tail unredacted.

Verified both are load-bearing: reverted just the fix and confirmed the new tests fail exactly as expected, then confirmed all 106 tests (3 new) pass with it restored.

Issue #1603, opened 2026-08-23
Agents spawned on remote fleet nodes die young and produce nothing, while spawn reports success


khaliqgant:
## Agents placed on remote fleet nodes register, then die young, producing nothing — and `spawned: true` hides it

Two independent instances today, on two different nodes, both confirmed dead by attaching rather than by reading a status field.

### Instance 1 — sf-mini

```
spawn factory-placement-why-0823  target_node sf-mini
  -> {"spawned": true}                         (and a duplicate invocation refused: spawn_agent_name_in_use)
  -> registers, roster node_id node_203549044126121984   (correct node)
  -> lastSeenAt 2026-08-23T10:20:14Z
  -> attach at 10:32Z:
     "No active agent named 'factory-placement-why-0823' on node 'sf-mini'"
     HTTP 404, code agent_not_found
```

**Output produced: none.** No Step 0 report, no DM, no issue comment, no commit. It owned the critical path of a live outage for ~20 minutes and evaporated.

### Instance 2 — daytona-1538-verify-0817b

```
spawn factory-322-guard-sandbox-0823  target_node daytona-1538-verify-0817b
  -> {"spawned": true, "agent_id": ...}
  -> registers, roster node_id node_214899219332034560   (correct node)
```
**Output produced: none.** Its target PR is byte-identical to before it was spawned.

### Controls

- Both were verified **remote**, not silently placed locally: roster `node_id` matched the requested node in each case, and a local process search found zero harnesses for either name (the same search returns hits for local lanes, so the instrument works).
- Local lanes spawned the same way, in the same session, with comparable briefs, **survive and produce work** — five of them merged PRs today. So this is specific to remote placement, not to the briefs or the workload.
- Both briefs opened with a mandatory Step 0 environment report explicitly designed to fail loudly (*"if you cannot clone or auth, say so and STOP — that is a valuable result, not a workaround"*). Neither arrived, so this is not a lane deciding to stay quiet.

### Why this matters more than it looks

1. **`spawned: true` is not arrival.** It reports that placement was accepted, not that an agent is alive and has its task. Every consumer of that field currently treats it as success.
2. **The failure is silent in both directions.** The agent produces nothing, and the registry keeps a plausible-looking record — `lastSeenAt` was recent and `status` read `offline`, which is the same thing the registry says about agents that are demonstrably alive. Nothing distinguishes "died young" from "thinking".
[...]

PR #1750, opened 2026-09-10
fix(broker): persist correlated worker exits

khaliqgant:
Addresses #1603.

What changed:
- preserve generation, workspace, readiness, spawn time, and Fleet invocation metadata when reaping workers
- persist bounded exit diagnostics immediately with atomic snapshots
- publish correlated hosted and local terminal exit events
- add backward-compatible protocol fields and deterministic coverage

### Critical review follow-up (base 355dd244d -> head 6d1baa8f9)

The initial durable-outbox design marked hosted `agent_exited` delivery
`Delivered` and persisted that right after the local `mpsc::try_send`
handoff to the publisher task — before `run_hosted_agent_event_publisher`
completed its real Relaycast HTTP emit. An HTTP timeout or 5xx after that
handoff was therefore permanently lost: the durable record already
(wrongly) said `Delivered`, so a broker restart would never replay it.

Fixed by moving the durable acknowledgment to after the real network
success boundary:
- `run_hosted_agent_event_publisher` performs bounded in-process
  retry/backoff (3 attempts) on timeout/5xx and reports the true outcome
  back over a new `HostedDeliveryOutcome` channel.
- `BrokerRuntime::handle_hosted_delivery_outcome` — the sole owner of
  `CrashInsights` — is now the only place that may mark a record
  `Delivered`, and only on confirmed HTTP success. A mere channel handoff
  no longer touches the durable record.
- On exhausted retries the record stays `Pending`; existing broker-restart
  replay is the fallback. A new `hosted_agent_exit_publish_failures_total`
  counter (exposed via `GetCrashInsights`) makes operator status truthful
  about failures instead of silently claiming delivery.
- `dedupe_key` (`agent_name::generation`) is now actually carried into the
  HTTP payload sent to Relaycast (previously computed but never
  transmitted), so a hosted consumer can discard a duplicate itself.
  Relaycast's HTTP API has no server-side idempotency key today (external
  SDK, out of repo), so this remains documented at-least-once, not
  exactly-once — filed as a companion issue, #1752.
- New deterministic tests fake a persistent and a transient Relaycast
  5xx/timeout: `Pending` remains `Pending` through failure/retries, and
  only confirmed success (in-process retry, or simulated restart replay)
  reaches `Delivered`.
- Retired `tests/relayflows/cases/1603-raw-spawn-readiness`: its
  actions-mocked MCP harness proved a different, already-fixed bug (raw
  CLI spawn readiness) and never booted a real broker/Relaycast boundary,
  so it could not stand in as proof of this fix. Replaced it with
  `tests/relayflows/cases/1603-hosted-exit-durable-delivery`, which spawns
  the real compiled `agent-relay-broker` binary against a fake local
  Relaycast HTTP server, seeds the exact on-disk `crash-insights.json`
  format with one `Pending` hosted-delivery record, and proves at the real
  binary boundary that `base` never replays it (the durable outbox does
  not exist there) while `head` replays it on startup and delivers it
  through one transient 503 and a successful retry. The narrower,
  exhaustive Rust regression tests remain authoritative for retry
  exhaustion, dedupe-on-replay, and >256-record backlog draining; this
  case only proves the real binary wires them together end to end.
- Corrected two stale doc comments in
  `crates/relay-pty/src/crash_insights.rs` that described a `try_send`
  channel handoff as the hosted-delivery acknowledgment (it isn't — only
  confirmed Relaycast HTTP success does) and described the durable pending
  outbox as bounded by `max_records` (it isn't — only already-delivered
  diagnostic history is).
- Hardened the Fleet E2E `getAgent` test helper
  (`tests/e2e/fleet/harness.ts`) against the free-plan Relaycast
  `429 rate_limit_exceeded` response that caused a prior "Two-node fleet
  matrix" CI flake: it now retries that specific error with a small,
  deterministic, bounded backoff (honoring `Retry-After` when present)
  before giving up, while every other status or error code still throws
  immediately exactly as before. New tests in
  `tests/e2e/fleet/get-agent-retry.test.ts` cover the retry, exhaustion,
  and non-retry paths.

Validation:
- cargo test -p relay-pty --lib (260 passed)
- cargo test -p agent-relay-broker --lib (1103 passed)
- cargo clippy -p agent-relay-broker --lib -- -D warnings (clean)
- cargo fmt -p agent-relay-broker -p relay-pty (no diff)
- cargo build -p agent-relay-broker -p relay-pty (clean)

The harness-driver TypeScript check remains blocked by the host Node runtime missing libada.3.dylib.

<a href="https://cubic.dev/pr/AgentWorkforce/relay/pull/1750?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## RelayFlow Proof

- Change type: `bugfix` 
- RelayFlow case: `1603-hosted-exit-durable-delivery`


khaliqgant:
Replaced the seeded outbox simulation with an independently reviewed live fleet-control proof. Exact new head: `e967438bbd3a7750a39165b0aad9bbb2bca144ba` (cherry-picked from approved isolated head `66db54ce175c1a8a24652d0415261a5932d3ee76`). The case now drives public `/v1/node/ws` `action.invoke`, correlates successful `action.result` plus child registration/inventory and a nonce/session/PID-bound real exit 23, proves durable Pending before HTTP success, kills/restarts the broker, then proves exact-generation replay once with stable dedupe and Delivered only after HTTP 200. Raw readiness remains untouched; 197 broker runtime tests, repeated head arms, syntax/diff checks, and Veto passed in the isolated review. Fresh exact-head hosted CI/RelayFlow are now required; no merge claim yet.

PR #1825, opened 2026-09-20
phase-0 delivery-backend seam: spike + review evidence, blocked on two HIGH findings

khaliqgant:
**Not a completed phase 0.** The workflow that produced this declined to seal it. This PR preserves the work and the evidence; merging it would not mean the exit criterion is met. Full findings in `docs/native-delivery/phase-0-review/`.

## What's here

A `DeliveryBackend` trait and `DeliverySeam` coordinator with the PTY injector as its first backend, wired into the real path at `runtime/delivery.rs`. Four named invariant tests for the doc's seam rules, each with its own mutation transcript. Fixes for a fabricated fleet ack, for echo normalization that could manufacture an observation, and for parity scripts that counted a timeout fallback as verified.

Green on: all five parity suites, `cargo fmt`/`clippy -D warnings`/release build, the invariant suite, `npm run typecheck`, 3,395 unit tests.

## Why it's blocked

**Green gates were not sufficient, and that's the headline result.** Two rounds of adversarial review over that same green tree produced **26 findings, 4 disqualifying**. Two remain HIGH and both are double-delivery class — what the migration doc calls "the worst failure mode" and "a release blocker".

### R2-1 — a timeout fallback permanently holes the ACK cursor

`runtime/worker_events.rs:866-908` settles the fallback by clearing the pending delivery and dropping the withheld ack, without calling `confirm_pending_delivery_and_resolve_fleet_ack`. `FleetDeliveryBook` is never told about the sequence.

The cursor is strictly prefix-ordered: `commit_confirmed_delivery` (`node_control.rs:1152-1177`) advances `acked_up_to_seq` only across a contiguous run of confirmed sequences and breaks at the first gap. So **one** timeout fallback pins every later delivery to that agent in `pending_deliveries`, and they are **re-injected after the next broker restart**.

This is a consequence of the fix applied in round 1, not of the code that fix replaced — so no earlier review and no gate in the campaign covers it. It exists *because* a round-1 fix was applied, which is the argument for the second review round.

### R2-2 — the only backend that exists violates the invariant it's tested for

`delivery/pty.rs:48-52` maps every `WorkerRegistry::deliver` error to `DeliveryError::unavailable`, and `backend.rs:157-159` defines `Unavailable` as `is_pre_write() == true`. But a PTY write can fail *after* bytes reach the tty. So `falls_back_only_before_write` — rule 1 — is false at the only backend there is.

The invariant test passes anyway, because it exercises a mock (**R2-11**).

### Also live

- **F4** — `DeliverySeam` is constructed inside `retry_pending_delivery`, so its receipt memory is per-call and **three of the four invariants have no production effect**.
- **F8** — whether a timeout-fallback ack is withheld (carrying verification kind on the ack) or documented as a PTY exemption is a **design decision**, not an implementation detail. Needs chief.

### One finding worth recording separately

A mutation probe was left on a live delivery path after being used to prove a test bites:

```rust
// crates/broker/src/runtime/app_server.rs — since removed
if std::env::var("RELAY_MUTATION_LOSSY_FORMAT").is_ok() {
    return format!("Relay message from {}:\n\n{}", delivery.from, &delivery.body[..1]);
}
```

`&body[..1]` panics on a multi-byte first character, on the one route that already reaches a session relay did not launch. **Every deterministic gate passed it** — scope, manifest routing, invariant tests, fmt, clippy, release build, typecheck, all five parity suites. Only the adversarial reviewer caught it. `seam-rules` now refuses mutation residue, and that gate is proven to bite.

## What unblocks it

Bounded and specific:

1. **R2-1** — teach the timeout-fallback path to inform `FleetDeliveryBook`, so the cursor can advance past a sequence that will never be confirmed. Bookkeeping in `node_control.rs`.
2. **R2-2** — give the PTY worker a real write-commit boundary in `worker.rs`; map post-write failures to a committed error rather than `Unavailable`.
3. **R2-11** — make the invariant suite exercise the real PTY backend instead of a mock.
4. **F4** — own the seam for the broker runtime lifetime, with bounded receipt eviction.
5. **F8** — decision from chief, then implement it either way.

## Consequence for the migration doc

`docs/native-delivery-migration.md` scopes phase 0 as *"Effort: small. Exit: parity suite green, unchanged, with the PTY backend behind the new trait."* Both halves are now known to be wrong: parity went green over a tree with two double-delivery defects, and putting the PTY behind the trait requires a write-commit boundary the PTY worker doesn't currently have. Phases 1–5 inherit that, since every later backend settles through the same seam.

## Review artifacts

`docs/native-delivery/phase-0-review/` — `shadow-rust.md`, `claude-review-1.md` (15 findings), `codex-review-1.md`, `claude-review-2.md` (11 findings), the three fix reports, `mutation-proof.md`, and `BLOCKED_NO_COMMIT.md`.

<a href="https://cubic.dev/pr/AgentWorkforce/relay/pull/1825?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


khaliqgant:
## Update — all gates re-measured green, and a fresh review still says do not seal

Three things done since the PR opened: the phase-0 lane was corrected, every gate was re-recorded against the current tree, and the one unreviewed change finally got read.

### Gates: green, freshly measured, nothing stale

The earlier evidence predated the last fixes. Re-recorded on the current tree:

| | |
| --- | --- |
| `rust-fmt` / `clippy -D warnings` / release build | green |
| `rust-tests` (full broker suite, 1,285+) | green |
| `invariant-tests` (4 seam rules, isolated) | green — 4 passed |
| `ts-typecheck` | green |
| `unit-tests` | 3,395 passed; 3 known-baseline failures, `regression-gate` PASS |
| parity × 5 | **green, first attempt, no retries** |
| `edit-gate` / `manifest-gate` / `targeted-gate` / `seam-rules` | green |

### The review: 2 blockers, 3 high, 5 medium

`docs/native-delivery/phase-0-review/fresh-review-codex-fix-2.md`. `codex-fix-2` was 1,278 insertions across 15 files that no reviewer had ever read — `codex-review-2` never ran.

**F1 (blocker), verified in the code.** `abandon_unconfirmed_delivery` (`node_control.rs:1184-1206`) copies the confirm path's drain loop including `confirmed_delivery_seqs.remove(&next)`, but its call site omits the consequences the confirm path applies. That map is the *only* thing holding an out-of-order-confirmed delivery out of the maintenance retry sweep.

Two deliveries to one agent: `seq 4` never echoes, `seq 5` echoes and confirms first but cannot advance the cursor. When `seq 4` settles unobserved, the drain removes **4 and 5**. `seq 5` — already delivered, already confirmed — becomes retry-eligible and is re-injected. And it does not stop: every later confirmation returns `None` and re-inserts, so the agent receives that message roughly **every 6 seconds, forever**, with a fresh `MessageDeliveryConfirmed` each round.

The preconditions are correlated, not independent: the message that fails to echo is exactly the one whose successor confirms first.

**F2 (blocker).** `TerminalInDoubt` on the fleet `WorkerMissing` path bails to an `Err` arm that withholds the ack and never calls `commit_received`. The engine redelivers, the broker re-injects — a post-write failure retried on the same transport, which is rule 1 — and the R2-1 cursor pin is re-created.

**F3 (high).** The commit boundary errs the other way: up to 127 provably *unwritten* queued commands per writer fault are mapped unconditionally to `Committed` and dropped with no dead letter.

**Also:** `codex-fix-2` reported fixing F4 (per-call seam construction). It is still there at `runtime/delivery.rs:1003`, so three of the four contract rules still have no production effect.

### The pattern worth taking from this

Third consecutive round in which a correct fix introduced a new defect nearby:

| fix | introduced |
| --- | --- |
| `claude-fix-1` fixed the fabricated ack | **R2-1**, the cursor pin |
| `codex-fix-2` fixed R2-1 | **F1**, infinite redelivery — worse |
| `codex-fix-2`'s `TerminalInDoubt` | **F2**, post-write retry on the same transport |

No gate caught any of them. Every one was found by a fresh adversarial read, and each was found only *after* the previous fix had shipped into the branch. Delivery semantics on a cumulative, prefix-ordered ACK cursor are evidently not safely patchable one finding at a time.

### Harness fixes included

- `edit-gate` went blind once the work was committed — it only diffed the working tree. It now merges that with a diff against the campaign base.
- The phase-0 lane gained `node_control.rs` and `worker.rs`; the accepted repairs required them, so the lane had made the phase unsatisfiable. Fourth widening, each time because a gate demanded an edit the lane forbade.
- `invariant-tests` now runs the seam binary alone, where its `0 passed` guard is meaningful; the full suite records separately. The guard had been firing on cargo's empty test target.

